PR #4, opened 2016-02-13
More derived SI units

JeffreySarnoff:
Newton, Pascal,
Ohm, Siemens, Frad, Henry, Tesla, Weber

### DIFF
--- a/src/Defaults.jl
+++ b/src/Defaults.jl
@@ -36,11 +36,19 @@ export ha
 @u    rad     _Radian
 
 # Derived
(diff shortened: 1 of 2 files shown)


JeffreySarnoff:
The -1 shown in the alteration count is the accidental deletion of a line,
which was immediately restored (moving the alteration count from +40 to +41).

If there is a way that you prefer to sequence the S! derived dimensions, or  If there is a way you prefer to sequence the dictionary entries (MLTA), tell me.

PR #6, opened 2016-02-16
make Kilogram the SI unit of mass (it is) and keep Gram

JeffreySarnoff:


ajkeller34:
Thanks for the contribution, but I don't see why this is necessary. This PR adds code with no additional functionality or improvement in performance. The reason the existing code uses gram instead of kilogram is because gram has no power-of-ten prefix. My code does not decree that the gram is the base SI unit.

JeffreySarnoff:
Ahh, I follow that. Do you want this package to have a way of collecting units as a system (SI, mks, cgs, atomic)?

PR #7, opened 2016-02-22
additional CODATA constants

JeffreySarnoff:
CODATA uses symbols that we have available.  
Other Julia apps define both the \sym and its similar letter in this situation.

This pull request gives you a way to see what you approach prefer.  Use what you like.


ajkeller34:
I think this PR has largely been superseded now that so many changes have happened, but many of the constants you wanted are in the defaults file now.

PR #10, opened 2016-07-28
Fix StepRange and enable many tests

timholy:
This causes `colon` to dispatch to `StepRange` rather than `FloatRange` for floating-point unitful numbers. There's some danger here, because it lacks the I-can-read-your-mind construction that `FloatRange` exhibits, so if you don't want this I'll understand.

I should also emphasize that this *deletes* the tests for the unitful `UnitRange`s. I don't think that concept makes sense, but if you disagree then it's another reason to be cautious about merging this.

### DIFF
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -12,7 +12,7 @@ import Base: mod, rem, div, fld, cld, trunc, round, sign, signbit
 import Base: isless, isapprox, isinteger, isreal, isinf, isfinite
 import Base: prevfloat, nextfloat, maxintfloat, rat, step #, linspace
 import Base: promote_op, promote_array_type, promote_rule, unsafe_getindex
-import Base: length, float, start, done, next, last, one, zero#, colon, range
+import Base: length, float, start, done, next, last, one, zero, colon#, range
 import Base: getindex, eltype, step, last, first, frexp
 import Base: Rational, Complex, typemin, typemax
 # import Base: steprange_last, unitrange_last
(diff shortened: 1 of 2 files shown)


timholy:
You might consider going to travis-ci.org, clicking on your icon, select Accounts, and then flip the switch on testing Unitful.

ajkeller34:
Thank you for the PR! I will take a look and likely approve within a day. This sounds like a good solution for now. I've flipped the switch on Unitful but I haven't used CI before, so I'll probably need to play around to get it working and have the nifty badge on the README. 

I'd rather have you on board and interested with this package than argue over a minor point, so I'm fine with removing `UnitRange` tests. I still don't see the problem if you think of `UnitRange{T}` as having a step equal to the additive group generator for the type `T`. If you think of `UnitRange` as a range with step 1 or `one(T)`, then I agree adding 1 to something with units doesn't make much sense. Also, for less experienced Julia users I'd need to explain why `1m:5m` doesn't work, or maybe just override `colon` to make a `StepRange`.

timholy:
Great!

The main reason I don't like the UnitRanges is this: let's say `x` and `y` are two positions, and we want to generate a unit range `x:y` that spans between them. The number of elements in that range is going to vary considerably depending on whether we measure position in `km` or `mm`, and you couldn't e.g. `convert` one range into the other in any meaningful way. (You'd have to sub- or super-sample the range.) In contrast, if you specify the `step`, then no matter what units you choose to write things down in, you'll get the same range out in the end.

timholy:
It's probably clear to you, but permit me to explain it another way. When I say two positions `x` and `y`, I'm thinking in concrete terms: for example, the left and right sides of my desk. Those are actual places, independently of how I write down their coordinates. Construct a "UnitRange" in millimeters, and you'll get more than a thousand steps between them; construct a "UnitRange" in decimeters, and you'll get only 10-20 steps.

But if I say "with a step equal to the length of this pen," then that specifies exactly how many steps there should be, no matter what units I measure all those quantities in. So the StepRange has a real-world meaning, whereas the resulting UnitRange is entirely an artifact of which measurement system you adopt.

ajkeller34:
That's a compelling reason and one I don't remember reading before. Thank you for clarifying.

ajkeller34:
So, if there is no opposition I will override `colon` to return a `StepRange` given the syntax `1m:5m`. I still think supporting this syntax could be convenient.

timholy:
I generally think that when in doubt, it's better not to define something. Typically there's a more verbose way to force it, and the harm of verbosity is not very large. If you define it anyway, and later realize you've made a mistake, it's often more painful for both you and the users.

In the case of single-character units like `m`, we're talking about 3 characters here, so I don't think that's too high a price to pay for clarity. One option, if you like it, is to return an informative error:
```jl
julia> 1m:5m
ERROR: UnitRange is not defined for physical units, since this concept is not invariant. Use `1m:1m:5m` instead.
```

PR #28, opened 2016-09-30
Support (1:3)*mm

timholy:
Given previous discussions and my opposition to unitful UnitRanges: note this creates a StepRange and is effectively shorthand for `1mm:1mm:3mm`, so I'm not being inconsistent here. I'd still complain if this returned a `UnitRange` :wink:.


ajkeller34:
Thank you, this is great!

PR #29, opened 2016-09-30
Simplify creation of typed arrays: `[1, 2]u"km"`

mdavezac:
It's already possible for quantities `[1, 1] * 1u"m"`, and `[1, 1] .* u"m"` is also explicitly  declared. So adds it for `*`.


ajkeller34:
Thanks for the PR!

Issue #24, opened 2016-09-26
StackOverflowError from @fastmath 


ChrisRackauckas:
Unitful.jl is incompatible with `@fastmath`. Here's a small example showing how to get the StackOverflowError (requires that you have optimized Julia to your system, like re-built the system image, and I can only get this error on Linux but not Windows)

```julia
using Unitful
u = 1.5u"N"
halfΔt = 0.03125u"s"
tmp2 = .25u"N/s"
u+halfΔt*tmp2
@fastmath u+halfΔt*tmp2
```

SIUnits has the same error at Keno/SIUnits.jl#94

ajkeller34:
Thanks for filing this issue. I can reproduce this on the release binary of 0.5.0 for macOS Sierra. The problem seems to be that promotion doesn't always return the same type if there are quantities involved (because it can't, there's no safe type to promote to from quantities with units `N/s` and `s`, which have different dimensions). So you get a `StackOverflowError` at line 241 of fastmath.jl in base. It's reasonable that fastmath.jl expects promotion to always return the same types, which is true for numeric types in Base, but promotion is not required to return the same types, e.g. `promote("a", 1) == ("a",1)`.

I don't think it would be that hard to work around by just specializing some methods for `Quantities`, hopefully I'll get some time to do this soon. I'm more concerned about the following behavior:

```
julia> promote(1u"s", 1.0u"s")
(1.0 s,1.0 s)

julia> promote(1u"s", 1.0u"N/s")
(1 s,1.0 N s^-1)
```

I consider this a bug in Unitful---promotion should at least make the "numeric part" of the types the same, in my view.

ajkeller34:
Having implemented some important changes to promotion, I now have fastmath working nicely on my local repository; once I finish repeating most of the fastmath tests from base (but with units) I will push to master. I guess that will be in another day or two. Making sure I haven't introduced performance regressions is another matter and I would be curious to see if you notice any. Also, `fma` is implemented on master already if you make use of that.

ChrisRackauckas:
This sounds great. Is there any way these promotions could be also added to SIUnits so it's compatible? I'd hate to have to drop SIUnits, but it is looking more and more like a "legacy" library.

ajkeller34:
Fixed on master.

[...]

PR #31, opened 2016-10-04
Support fma (fused multiply-add)

timholy:
In trying to use this for various real world applications, I finally got tired of not having decent ranges. The result was https://github.com/JuliaLang/julia/pull/18777, for which `fma` support seems likely to be important.

I also have a package brewing so we can use this on 0.5.

### DIFF
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -3,7 +3,7 @@ module Unitful
 
 import Base: ==, <, <=, +, -, *, /, .+, .-, .*, ./, .\, //, ^, .^
 import Base: show, convert
-import Base: abs, abs2, float, inv, sqrt
+import Base: abs, abs2, float, fma, inv, sqrt
 import Base: min, max, floor, ceil, log, log10, real, imag, conj
 
 import Base: mod, rem, div, fld, cld, trunc, round, sign, signbit
@@ -588,6 +588,12 @@ end
 ^{T,D,U}(x::Quantity{T,D,U}, y::Real) = Quantity((x.val)^y, U()^y)
 
 # Other mathematical functions
+@inline function fma{T,D,U}(x::Number, y::Quantity{T,D,U}, z::Quantity{T,D,U})
+    c = fma(x, y.val, z.val)
+    Quantity(c, U())
+end
+fma(x::Number, y::Quantity, z::Quantity) = fma(x, promote(y, z)...)
+
 """
 ```
 sqrt(x::Quantity)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -170,6 +170,8 @@ end
         @test V*(3+4im) == (3V+4V*im)
         @test (3.0+4.0im)*V == (3+4im)*V
         @test im*V == Complex(0,1)*V
+        @test fma(2.0, 3.0m, 1.0m) == 7.0m
+        @test fma(2.0, 3.0m, 35mm) == 6.035m
     end
 
     @testset "> Addition and subtraction" begin


ajkeller34:
This looks great, and I really appreciate your efforts to improve ranges. Even better that some of this functionality will be available on 0.5!

Before merging, I wanted to bring your attention to a problem I've encountered, which seems like it will be relevant here. For this PR, the problem is not limiting, so I'm willing to push through this PR and leave an open issue regarding promotion, if you'd like. This is a larger problem though, appearing also in #24 regarding `@fastmath`.

The problem is that because quantities have different types for different units and dimensions, promoting fails to return a common type if you promote quantities of differing dimensions. If you try and do `promote(1m, 1N)`, for example, you will get `(1 m,1 N)`, although `promote(1m, 1mm)` gives `(1//1 m,1//1000 m)`. The result is that if a user tries to call your more generic method improperly, by adding quantities of differing dimension, I think a stack overflow will occur rather than a graceful failure: the method will keep calling itself, because promotion couldn't return a common type.

One idea I had would be to have a values-based (rather than type-based) quantity as a fallback, which could represent any units or dimensions in one type, at the expense of run-time penalties. This way, I could guarantee that promotion could always yield a common type (type-based quantities could be promoted to value-based quantities if there were no other option). Obviously I don't like the idea of sacrificing performance to get `@fastmath` to work, however, and it seems like overkill to fix the problem of graceful failure in this PR.

ajkeller34:
On second thought it's not really as if the problem I'm describing is worth holding this up, so I'll just merge it. At least you're aware of the issue now.

timholy:
I've noticed this as a problem, too.

Perhaps we need to use the following design:
```jl
foo{T}(a::T, b::T) = # the real method, i.e., the one that implements the computation
foo(a, b) = _foo(promote(a, b)...)
_foo{T}(a::T, b::T) = foo(a, b)   # promotion worked
_foo(a, b) = error("promotion of $a and $b failed to produce a common type")
```

ajkeller34:
I can certainly implement that design pattern in Unitful for your `fma` methods, and elsewhere as appropriate. I feel like line 241 of [fastmath.jl](https://github.com/JuliaLang/julia/blob/master/base/fastmath.jl) may need a PR along these lines for Unitful to work with it though... I've tried, but it's pretty awkward to work around that method, especially when unitless numbers and quantities are mixed in legitimate ways for some operations. I'll think about it a little more.

timholy:
Can you submit a julia PR? If not, I'll do it (eventually).

ajkeller34:
I'll add that to my list, but do let me know if I'm taking too long and you want to take care of it instead... having some build problems with Julia on macOS Sierra that I need to figure out.

Please see [here](https://github.com/ajkeller34/Unitful.jl/blob/fma/src/Unitful.jl), lines 590–625, for a proposed change to your `fma` contributions. It passes more [test cases](https://github.com/ajkeller34/Unitful.jl/blob/fma/test/runtests.jl) than before and errors gracefully, but I have not benchmarked the performance and the code is not exactly elegant. There are no detected ambiguities.

ajkeller34:
Having learned my lesson about submitting julia PRs prematurely some months ago, I thought some more and I think I may be able to get fastmath working without one. In trying to get around an existing method in base like this (e.g. line 241 of fastmath.jl):
```
f(x::Number, y::Number, zs::Number...) = f(promote(x,y,zs...)...) # method (1)
```

It seems you can write these methods, with no dispatch ambiguity:
```
f(x::Number, y::Quantity) = _f(promote(x,y)...) # (2)
f(x::Quantity, y::Number) = _f(promote(x,y)...) # (3)
f(x::Quantity, y::Quantity) = _f(promote(x,y)...) # (4)
f(x::Number, y::Number, z::Number, ts::Number...) = _f(promote(x,y,z,ts...)...) # (5)
```

Method (1) would only be called when passing two arguments, neither of which are `Quantity`s. I feel silly for not trying this before, I assumed method (5) would be ambiguous with method (1). Until I see some new problem with this route I don't think we will need a julia PR after all.

Anyway, thanks for humoring me, I'll stop bothering you now 😅

timholy:
Cool github trick: navigate to the page containing the code you want to link to, select the range of line numbers you want to highlight (you can use the shift key), hit the 'y' key on your keyboard (this inserts the current commit sha into the url, making it a "permalink" even as the code evolves), and then copy paste the url into a discussion page.

The one issue I see with the `_fma` methods is they seem to do a lot of runtime checking. Can any of that be done by the type system? `fma` (when hardware support is available) is pretty high performance, so having all those branches will be bad news.

timholy:
With regards to eliminating runtime checking: you can either wait until we get triangular dispatch (julia 0.6, see https://github.com/JuliaLang/julia/pull/18457) or do stuff like this:
```jl
_fma{T<:Quantity}(x, y::T, z::T) = _fma_dimension(x, dimension(x), y, z)
_fma_dimension(x, ::Dimension{()}, y, z) = fma(x, y, z)
_fma_dimension(x, ::Dimension, y, z) = throw(DimensionError())
```
The general trick is to add arguments that "peel out" the things you need to examine, then use dispatch to direct the execution to either error messages or computation.

All this gets inlined by the compiler, so (unlike the runtime checks) it doesn't have any negative impact on performance.

ajkeller34:
Thank you for the tips! Some follow-up questions:

- The `dimension` and `unit` functions in Unitful give a result based only on the type of the argument. Should I be annotating the `dimension` and `unit` functions to indicate that to the compiler (maybe with `@pure` or some such thing)? I guess even then a runtime check [like this](https://github.com/ajkeller34/Unitful.jl/blob/67c7f185f2978136fd3c2fa8e642b68f3a828927/src/Unitful.jl#L606) would still be a runtime check, and would not be magically moved to compile time.

- What's the best way to verify things are being inlined? Would you want explicit `@inline` annotations on any of that?

ajkeller34:
Alright, what I'm specifically confused about is the following. Using `master` branch, which has your `fma` implementation, the following calls [here](https://github.com/ajkeller34/Unitful.jl/blob/80b3d4c5d033c4b9881fd24dda8bc6a08b463ee1/src/Unitful.jl#L591):

```
julia> @code_llvm Unitful.fma(1.0, 1.0m, 2.0m)

define %Quantity @julia_fma_71023(double, %Quantity*, %Quantity*) #0 {
top:
  %3 = getelementptr inbounds %Quantity, %Quantity* %1, i64 0, i32 0
  %4 = load double, double* %3, align 8
  %5 = getelementptr inbounds %Quantity, %Quantity* %2, i64 0, i32 0
  %6 = load double, double* %5, align 8
  %7 = call double @llvm.fma.f64(double %0, double %4, double %6)
  %8 = insertvalue %Quantity undef, double %7, 0
  ret %Quantity %8
}
```

Whereas with my implementation in the `fma` branch, the following calls [here](https://github.com/ajkeller34/Unitful.jl/blob/67c7f185f2978136fd3c2fa8e642b68f3a828927/src/Unitful.jl#L605), which appears to have a run-time check, though code_llvm suggests otherwise:

```
julia> @code_llvm Unitful._fma(1.0, 1.0m, 2.0m)

define %Quantity @julia__fma_70971(double, %Quantity*, %Quantity*) #0 {
top:
  %3 = getelementptr inbounds %Quantity, %Quantity* %1, i64 0, i32 0
  %4 = load double, double* %3, align 8
  %5 = getelementptr inbounds %Quantity, %Quantity* %2, i64 0, i32 0
  %6 = load double, double* %5, align 8
  %7 = call double @llvm.fma.f64(double %0, double %4, double %6)
  %8 = insertvalue %Quantity undef, double %7, 0
  ret %Quantity %8
}
```

Even more surprisingly, [this](https://github.com/ajkeller34/Unitful.jl/blob/67c7f185f2978136fd3c2fa8e642b68f3a828927/src/Unitful.jl#L619) doesn't appear to have a run-time check either:

```
julia> @code_llvm Unitful._fma(1.0N/m, 1.0m, 2.0N)

define %Quantity.66 @julia__fma_71230(%Quantity.73*, %Quantity*, %Quantity.66*) #0 {
top:
  %3 = getelementptr inbounds %Quantity.66, %Quantity.66* %2, i64 0, i32 0
  %4 = getelementptr inbounds %Quantity.73, %Quantity.73* %0, i64 0, i32 0
  %5 = load double, double* %4, align 8
  %6 = getelementptr inbounds %Quantity, %Quantity* %1, i64 0, i32 0
  %7 = load double, double* %6, align 8
  %8 = load double, double* %3, align 8
  %9 = call double @llvm.fma.f64(double %5, double %7, double %8)
  %10 = insertvalue %Quantity.66 undef, double %9, 0
  ret %Quantity.66 %10
}
```

So I don't think they are doing run-time checks, despite appearances...?

timholy:
Hey, if the compiler is eliding them, great! Much easier this way.

ajkeller34:
Alright, so fma is maybe 90% working, but I can't devote any more time to this right now; seemingly small changes to the code (to me anyway) gave pretty different llvm outputs, and I kept hitting method ambiguities. You may want to checkout master and test for yourself that you're satisfied. I tried to avoid regressing on the method you implemented, since I presume it is the most important case for you. runtests.jl gives some indication of where I saw problems and where llvm output was "good" or "bad." Unless you're using dimensionless quantities a lot you shouldn't run into trouble, at least on julia 0.5.0.

With a view on future changes: This has been very interesting to see the limits of promotion as I've currently implemented it. I think one problem is that `::Number` gives you no insight about whether something has dimensions or not, since `Quantity <: Number`. My hope is that I could simplify all of this with tweaks to promotion on my end. For instance, `promote_type(typeof(1.0), typeof(3m))` will result in `Number` right now, but the situation may improve if it instead returned `Quantity{Float64}`. If `Number`s from base were wrapped in a dimensionless, unitless `Quantity` type, maybe dispatch could be simplified, since then I could just write `fma(x::Quantity, y::Quantity, z::Quantity)` and go from there, rather than starting with the 7 generic `fma` methods I need now to work around what is in base.

Issue #35, opened 2016-10-13
Converting between unit systems


mdavezac:
It would be nice to have the ability to transform from on unit system, say standard SI, to another, say atomic units, without having to specify the explicit units as with `uconvert`.

I'm not sure exactly how to go about it. One way would be have a function

~~~Julia
system = u"g * km * eV"
simplify(a, system)
~~~~

We can then project `units(simplify(a))` (simplified as in #34) onto the units in the second argument + an orthogonal subspace of units defined by: (i) just SI units, or (ii) the units of `simplify(a * system)` over each individual dimension. I think I would go for (ii).
One caveat is users entering a unit system that is not linearly independent.

Another possibility is to define a new type with an appropriate `*` operator. And then proceed in a fairly similar manner. The difference is that the type and associated arithmetic could be made to guarantee that the units are linearly independent. And `uconvert(a, system)` could be used. 

Comments and preferences?

ajkeller34:
Good questions, to which I have no immediate answer... I'll think about it.

Issue #34, opened 2016-10-13
Function to simplify units, e.g. m/ft -> dimensionless


mdavezac:
@ajkeller34, unless I'm mistaken, the following functionality (`simplify`) does not yet exist. Happy to try and implement it if it doesn't:

~~~Julia
notsosimple = 1u"m/ft*kg"
simple = simplify(notsosimple)
@assert units(simple) === units(1u"kg")
~~~

It should be possible to determine which unit to use of m or ft, the same way it is done for +, via` promote_type`

ajkeller34:
There's this:

```
julia> notsosimple = 1u"m/ft*kg"
1 ft^-1 kg m

julia> upreferred(ans)
1250//381 kg
```

This may not be what you want though, since the result would be in `kg` even if the mass unit in the first expression were a different unit. What this does is look at the dimensions of the quantity, and for each power of length, have that many powers of a meter, etc., where the dimension-to-unit mapping is given by the `@preferunit` macro in the defaults. So this basically is a "convert to base SI units" function given what is usually in `deps/Defaults.jl`.

If you have different functionality in mind, how would you simplify when you have units of mixed dimensions like joules where it's not just length or mass or time but a combination?

mdavezac:
I guess I'm thinking of a function that would simplify units that have the exact same dimensions and leave those that partially overlap. More complex conversions would happen in #35 (for which upreferred would be one instance). I might trial an implementation so we have something concrete to discuss.

Issue #33, opened 2016-10-13
Defaults handling needs improvement


ajkeller34:
The `deps/Defaults.jl` file is not regenerated with each invocation of `Pkg.build("Unitful")` if the file exists already. This is to avoid overwriting user defaults. However, in some cases the defaults need to be changed to be compliant with changes to code in the package. It would be nice if there were a mechanism to handle this so that the user doesn't have to clear out that file or merge their changes manually.

ChrisRackauckas:
Could you have a way to override so a user could do `Pkg.build("Unitful",override_defaults=true)`, and throw an informative error if the defaults are wrong: 

```julia
Your defaults are bad. If you have recently upgraded, you may need to use `Pkg.build("Unitful",override_defaults=true)`. Warning: This will override user defaults.
```

This is similar to how in Plots.jl or DifferentialEquations.jl if you try to use a conditional dependency it will give install instructions. Not every time the user needs it, but I think the vast majority of users will find this helpful (and those who are using their own defaults are probably power users who will understand what this error message means anyways)

ajkeller34:
For now I will catch errors emitted from `include("Defaults.jl")`, let the user know that the defaults are bad, and suggest a course of action. I'm hesitant to extend `Pkg.build` but maybe in the future I could provide a convenience function `rm_defaults()` or something. Anyways, thanks, will tag and release after CI finishes.

mweastwood:
In theory I like the idea that the user may be able to specify their own units, but I strongly believe that this should happen in their `.juliarc.jl` instead of in `deps/Defaults.jl`. I want to be able to reliably count on the existence of certain units, because I don't want to distribute code that depends on the "day" unit if is possible that somebody may have deleted that line or changed its definition. I would much prefer the definition of "day" to be fixed.

ChrisRackauckas:
Good point with the `.juliarc`, that's much more standard.

ajkeller34:
I would like to do this in principle but there are a few obstacles:

- I'm not sure how you guarantee the existence of certain units if you also allow the user the flexibility to choose whatever reference system of units they want. The `@unit` macro requires a quantity given in terms of previously defined units. If the user wanted non-SI reference units, all subsequent unit definitions would need to change. This may be a niche case, but to maintain current levels of customizability while guaranteeing the user hasn't removed some of the unit definitions is a tall order. Maybe this degree of customization is just totally excessive.
[...]

PR #39, opened 2016-10-18
First attempt at a simplification mechanism

mdavezac:
This  pull request is far from complete. If anything, it requires some more unit-testing. But I would appreciate some thoughts on whether it could eventually make it. The main problem, as I see it, is that I have to use a non-exact algorithm to project onto a non-orthogonal set of basis vector, each vector being a unit.

Usage is given in the docstring for `simplify`.

The simplification proceeds by:
1. figuring out a minimal set of linearly independent units from the input quantity/units. The set of units are chosen to have the smallest possible dimensions at this stage.
2. projecting the input units onto this minimal set units. The set can be non-orthogonal, so the projection takes place via a iterating projection algorithm. It is not an exact algorithm, so some rounding is done at the end. In most cases, I would expect this is sufficient to recover the exact projection.
3. conversion to the new units using `uconvert`

The main problem is figuring out how to project onto a non-orthogonal
basis. There are probably other (better?) ways of doing this.

Another issue is how to figure out some heuristics to choose the linearly
independent units. We could try and add conditions to favor SI units, and/or
units with the smallest prefix (as in smallest abs(log10(prefix))), and/or
even the units with the smallest screen foot-print when dumping to STDOUT.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -204,6 +204,9 @@ else
                 end
             end
         end
+
+        # Defines units preferred by the SI
+        const SI = (m, s, A, K, cd, kg, mol)
         """)
     end
 end
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -23,6 +23,7 @@ export @dimension, @derived_dimension, @refunit, @unit, @u_str
 export Quantity
 export DimensionlessQuantity
 export NoUnits, NoDims
+export simplify
 
 include("Types.jl")
 include("User.jl")
@@ -888,6 +889,7 @@ include("Display.jl")
 include("Promotion.jl")
 include("Conversion.jl")
 include("fastmath.jl")
+include("VectorialSpace.jl")
 
 defpath = joinpath(dirname(dirname(@__FILE__)),"deps","Defaults.jl")
 if isfile(defpath)
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -24,6 +24,7 @@ export Quantity
 export DimensionlessQuantity
 export NoUnits, NoDims
 export simplify
+export SI
 
 include("Types.jl")
 include("User.jl")
--- a/src/VectorialSpace.jl
+++ b/src/VectorialSpace.jl
@@ -5,6 +5,7 @@ dimensional_space(units::Units)
 ```
 
 Aggregates the dimensions present in the units (whether or not they cancel out).
+Returns the dimensions present in the unit with a power of 1.
 
 ```jldoctest
 julia> dimensional_space(u"m/km")
@@ -34,9 +35,9 @@ dimensional_space(u::Units) = dimensional_space(typeof(u))
 
 """
 ```
-dimensional_points{Us <: Units}(::Type{Us}, dimspace::Dimensions)
-dimensional_points{Us <: Units}(::Type{Us})
-dimensional_points(units::Units)
+dimensional_matrix{Us <: Units}(::Type{Us}, dimspace::Dimensions)
+dimensional_matrix{Us <: Units}(::Type{Us})
+dimensional_matrix(units::Units)
 ```
 
 Transforms input unit to vectorial format, where each component in the vector is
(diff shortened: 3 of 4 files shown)


mdavezac:
@ajkeller34, the pull-request now implements the following two simplification mechanisms (from #34 and #35):

```Julia
# simplify using the Unit types listed in the Units input instance
@test simplify(u"m*m*km*J/cm") == u"J*cm^2"

# simplify using the preferred Unit types in the list, in order, as well as those in the instance
@test simplify(u"J", Unitful.Units[u"Hz", u"nm", u"kg"]) === u"kg*Hz^2*nm^2"
@test simplify(u"J", Unitful.Units[u"Hz", u"nm", u"kg", u"J"]) === u"kg*Hz^2*nm^2"
@test simplify(u"J", Unitful.Units[u"C"]) === u"J"
```

Overloads are defined for actual quantities:

```Julia
@test simplify(1u"m*m*km*J/cm") == 1000000000//1 * u"J*cm^2"
```

It is possible to define constant tuples for each system, SI, atomic units, etc and use those as shorthand for simplification into the corresponding system:

```Julia
@test simplify(u"J*N", SI) === u"kg^2*m^3*s^-4"
@test simplify(1u"J*N", SI) == 1u"J*N"
```
With `SI` currently defined in the default `Defaults.jl`.

However, unless a new type is created for these systems, I don't think we can make simplification a generated function. So that could be the next step. What do you think?

ajkeller34:
I do hope to review this at some point soon but I am currently rather behind. I'm sorry for the delay. This sort of thing will be important for Unitful but I consider #33 and associated PR #47 higher priority, since the current system can be frustrating at times.

tlnagy:
Hey @ajkeller34, any chance you could look back over this?

My use case is that I have files that encode unit types, which I map to Unitful types. However, I want to make sure that they are simplified for easiest display.

```julia
data = 10000.0
str_type_from_file = "ms"
unittype = getfield(Unitful, Symbol(str_type_from_file))
unitfuldata = data*unittype
```
which gives me
```
10000.0 ms
```
Now it would be fantastic if I could simplify this
```julia
simplify(unitfuldata)
```
```
10.0 s
```

mdavezac:
@tlnagy, have you looked at `Unitful.upreferred`? It will transform units to SI in most cases.
This pull-request was meant to be a more versatile approach than `upreferred`.

@ajkeller34, should I try and resurrect this? Is Unitful in a place where you could now consider it?

tlnagy:
@mdavezac thanks for the pointer, but `upreferred` seems to generate pretty ugly output:

```julia
Unitful.upreferred.((10000:1000:13000)*u"ms")
```
```
4-element Array{Quantity{Rational{Int64}, Dimensions:{𝐓}, Units:{s}},1}:
 10//1 s
 11//1 s
 12//1 s
 13//1 s
```
I would prefer something like 10s, 11s, 12s, 13s without the rational

mdavezac:
Arguably, that's one of the nice features of Unitful. It's not always possible to keep to integers when doing a conversion. Try `1u"ms" |> u"s"` rather than your examples above. 

There are two constraints:

1. accuracy
1. type stability: the type of the result cannot depend on the values of the input (but only on the type of the input). So converting `1u"ms"` should have the same type as converting `10_000u"ms"`

So going from integers to rationals is the best way of fulfilling both constraints. It only works for conversion factors that can be inputed as rationals in the first place, though.

mdavezac:
For what it's worth, you could convert to integers, but it will fail in cases where the conversions would yield numbers < 1: 

```Julia
julia> convert.(typeof(1u"s"), Unitful.upreferred.((10000:1000:13000)*u"ms"))
4-element Array{Quantity{Int64, Dimensions:{ùêì}, Units:{s}},1}:
 10 s
 11 s
 12 s
 13 s

julia> convert.(typeof(1u"s"), Unitful.upreferred.((100:1000:13000)*u"ms"))
ERROR: InexactError()
````

The other option is to start off from floats:

```Julia
julia> Unitful.upreferred.((100.0:1000.0:13000.0)*u"ms")
13-element Array{Quantity{Float64, Dimensions:{ùêì}, Units:{s}},1}:
  0.1 s
  1.1 s
  2.1 s
  3.1 s
  4.1 s
  5.1 s
  6.1 s
  7.1 s
  8.1 s
  9.1 s
 10.1 s
 11.1 s
 12.1 s
```

tlnagy:
The float suggestion works wonderfully. I realized that I had tried that, but forgot to broadcast the `Unitful.upreferred` function call so I was getting an error, but it totally works when I do

```
Unitful.upreferred.((100.0:1000.0:13000.0)*u"ms")
```
Thanks!

PR #43, opened 2016-11-04
Allow inversion of unitful matrices

mdavezac:
At least in some cases, the pivot step in the matrix inversion requires
scaling one coefficient by another. However this fails when implemented
in terms of two operations: first an inversion which gives an object
with an inverse unit (rather than a dimensionless quantity), then a
`*=` multiplication. The second step tries (incorrectly) to change the
dimension of the matrix coefficient to dimensionless.

The relevant lines in `linalg/lu.jl` are:

```Julia
Akkinv = inv(A[k,k])
for i = k+1:m
    A[i,k] *= Akkinv
end
```

This commit hacks an `inv` function for the relevant matrices that first
strips the units from the matrix.

### DIFF
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -559,6 +559,15 @@ end
     :($y)
 end
 
+function inv{T <: Quantity}(x::Matrix{T})
+    length(x) > 0 &&
+        any(dimension(x[1, 1]) .!= collect(dimension(u) for u in x)) &&
+        error("Matrix must have consistent dimensions to perform inversion")
+    # make sure that matrix we will invert has consistent units
+    # only then can we invert the underlying matrix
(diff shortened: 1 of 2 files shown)


src/Unitful.jl
```diff
@@ -559,6 +559,15 @@ end
     :($y)
 end
 
+function inv{T <: Quantity}(x::Matrix{T})
+    length(x) > 0 &&
+        any(dimension(x[1, 1]) .!= collect(dimension(u) for u in x)) &&
```
ajkeller34: This check is not necessary because the method signature implies the matrix is concretely-typed. The dimension of each element will be the same.

ajkeller34:
Thanks for the PR and sorry for the delay. I rewrote your `inv` method for some efficiency gains and merged it.

PR #47, opened 2016-11-17
[RFC] proof of concept for associating a system of units with modules

mweastwood:
Don't merge this. This is meant to be a demonstration of what I think needs to happen for #33.

* Moved `deps/Defaults.jl` to `src/Defaults.jl`
* A sensible deprecation strategy for `deps/Defaults.jl` might be necessary
* You get the default units by doing `using Unitful.Defaults`
* You start building your own set of units by doing something like:
``` julia
module MyUnits
importall Unitful.Builder
export @u_str
...
end
```
(see `src/Defaults.jl` for an example of this)

I commented out one test because it creates a new unit without putting it in a module. Before this was fine because everything just gets added to `Unitful`. This test can be fixed but that would complicate this proof of concept somewhat.


mweastwood:
Closing because #33 has been resolved. :fireworks:

PR #51, opened 2017-02-05
Get most range tests working

timholy:
On Julia 0.5, this exploits Ranges.jl to give "decent" (but still imperfect) ranges. On Julia 0.6, this gets most of the tests working using the new range infrastructure. Requires https://github.com/JuliaLang/julia/pull/20466. There's still at least one problem, though, which stems from this:
```julia
julia> x, y = 10m, 1
(10 m, 1)

julia> px, py = promote(x, y)
(10 m, 1)

julia> ppx, ppy = promote(px, py)
(10 m, 1)

julia> (typeof(y), typeof(py), typeof(ppy))
(Int64, Quantity{Int64, Dimensions:{}, Units:{}}, Int64)
```
Hilariously, this defeats what I thought was an oh-so-clever mechanism behind `Base.promote_noncircular` https://github.com/JuliaLang/julia/pull/19918. Good exercise in humility :smile:. I didn't have the time to dig into why this "type-oscillation" happens, but at least this is progress. The tests don't seem to pass anyway on 0.6, so I didn't seem much harm in leaving some of the uncommented.

Prompted by https://github.com/JuliaImages/juliaimages.github.io/issues/8.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,3 @@
 julia 0.5
 Compat 0.9.5
+Ranges
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -3,6 +3,10 @@ module Unitful
 
 using Compat
 
+@static if VERSION < v"0.6.0-dev.2390"
+    using Ranges
+end
+
 @static if VERSION < v"0.6.0-"
     import Base: .+, .-, .*, ./, .\
 end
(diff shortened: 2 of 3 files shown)


ajkeller34:
Thanks for the PR. I'll take a look over this. I had a local branch with some fixes too, so probably I'll merge yours first and then see if I had any additional fixes. I just fixed the other failures on master which were related to fastmath.

Quick question if you don't mind: how do you get the build numbers for Julia version strings, as you've done in this PR? I'd like to clean up some of my version checks. I tried following what was happening in `base/version_git.sh` and it seems like you need to do something like `git rev-list --count HEAD "^$verchanged"` in a julia repo where `verchanged` is the git commit ID when VERSION was changed. However, when I do this, I don't understand the result. For instance, [this commit](https://github.com/JuliaLang/julia/commit/f0827da6cc53a55a7ad7d99a9518962b77170b0f) gives build 2207 and [this commit](https://github.com/JuliaLang/julia/commit/69b0f439c32987245a5446202f00a6ba610cf69f) gives build 2205, even though they were made more than a month apart. Perhaps I misunderstand what the build number means? I thought it was something like "number of commits since last time VERSION was changed."

timholy:
From inside the julia source directory, `contrib/commit-name.sh $sha`. See [this file](https://github.com/JuliaLang/julia/blob/master/contrib/commit-name.sh).

timholy:
Oh, and forgot to add: I think the version numbering reflects the merge into master, not when individual commits were made. Make sure you always run that script on the merge commit itself (assuming there is one---anything contributed by PR rather than a direct push should have one).

ajkeller34:
I worked on the range code a little more. All the tests pass on master now, for both julia 0.5 and 0.6 (though CI is still running). I've made changes to the code I merged from your branch for julia 0.6, which hopefully are not too heretical.

I follow the logic in base/range.jl and base/twiceprecision.jl pretty closely, so the end result is that you should get the TwicePrecision magic when you expect it and Unitful users can reap the benefits of your hard work.

More tests are always welcome, of course. I'll make a separate issue for the enchanting type oscillation you found. Thanks again!

PR #58, opened 2017-03-06
fix isapprox units

simonbyrne:
I realised that there was a potential error in my earlier PR if the tolerance units were different. This should fix it.


ajkeller34:
Thanks!

PR #60, opened 2017-03-09
add temperature conversion example

ggggggggg:
I had to look at your code to even figure out how to enter degrees C, so this serves as an example as that as well.


ajkeller34:
Thanks!

PR #84, opened 2017-05-25
Fix definition of Bohr magneton

klickverbot:
`e` was supposed to be the elementary charge, not the base of the
natural logarithm.

Thankfully, this was easy to spot using the unit system. ;)


ajkeller34:
Ha, whoops. Thanks!

PR #96, opened 2017-09-01
Adds overloads for `rand` and `ones`

mdavezac:
Overload for `ones` might not be necessary if JuliaLang/julia#23544 is accepted.


ajkeller34:
I am making some changes based off of your PR (to avoid `reinterpret` since that seems to be in flux, https://github.com/JuliaLang/julia/issues/22849) but will merge later today. Here and elsewhere there were enough differences in behavior between 0.5 and 0.6 that I'm just going to target 0.6 moving forward, so I'm waiting for CI to pass on master after having merged a 0.6-only branch.

Issue #94, opened 2017-07-27
Need better accounting of where units are defined


mweastwood:
Consider this example:

```julia
julia> using Unitful

julia> module UnitfulTest 
           import Unitful
           using Unitful: @unit, s
           @unit twoseconds "twoseconds" TwoSeconds 2s false
           Unitful.register(UnitfulTest)
       end
UnitfulTest

julia> uconvert(u"s", 1*u"twoseconds")
WARNING: Symbol s was found in multiple registered unit modules. We will use the one from UnitfulTest.
2//1 s
```

At the moment it is impossible to import units into a module that is later registered with Unitful. This generates a warning because Unitful thinks the unit is defined twice.

I think the `@unit` macro should push to a module-local list that keeps track of which units are defined in that module. Then you no longer need to use `isdefined(module, symbol)` to locate where units are defined.

ajkeller34:
Closed by #98.

PR #97, opened 2017-09-13
Additional units: erg, Jy

pboley:
I suggest to include erg (1e-7 J) as a separate energy unit, and (maybe more controversial) Jansky (1e-26 W/m^2/Hz), which is very commonly used in astronomy for measuring spectral irradiance.

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -100,6 +100,7 @@ end
 # Energy
 const q = 1.6021766208e-19*C        # CODATA 2014; `e` means 2.718...
 @unit eV     "eV"       eV          q*V                     true
+@unit erg    "erg"      erg         1e-7*J                  false
 
 # For convenience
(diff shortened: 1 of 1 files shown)


mweastwood:
FYI, both of these units will be part of [UnitfulAstro](https://github.com/mweastwood/UnitfulAstro.jl/blob/0166797e2b80a8115f7259bfdfa0bd32aefb3da5/src/UnitfulAstro.jl). Which is, at the moment, held up by #94.

ajkeller34:
Thanks for the PR. UnitfulAstro was recently announced on Julia's Discourse forum, so I believe it is ready for use (though not yet registered in METADATA). I suggest you use that package. I admit the choice of what units should go into Unitful or separate packages is pretty arbitrary. Certainly a case could be made for erg belonging in Unitful itself, but it sounds like you might find other astronomical units useful anyway.

PR #98, opened 2017-09-14
don't warn for identical units when resolving symbols from u"..."

mweastwood:
This is just a quick patch I whipped up to try and address #94. Basically, if two units have the same hash, we will no longer warn about finding two copies of the same unit. We also check that these symbols are actually units before warning (I think just having a symbol named "s" was enough to get the warning).

Unfortunately there is a drawback to the approach I've used here. Because Dicts are unordered, I  think there's no longer any guarantee about which unit is selected in the event that there are in fact multiple units with the same name.

*Before:*
```
julia> using Unitful

julia> module UnitfulTest
           import Unitful
           using Unitful: @unit, s
           @unit twoseconds "twoseconds" TwoSeconds 2s false
           Unitful.register(UnitfulTest)
       end
UnitfulTest

julia> uconvert(u"s", 1*u"twoseconds")
WARNING: Symbol s was found in multiple registered unit modules. We will use the one from UnitfulTest.
2//1 s
```

*After:*
```
julia> using Unitful

julia> module UnitfulTest
           import Unitful
           using Unitful: @unit, s
           @unit twoseconds "twoseconds" TwoSeconds 2s false
           Unitful.register(UnitfulTest)
       end
UnitfulTest

julia> uconvert(u"s", 1*u"twoseconds")
2//1 s

julia> module UnitfulTest2
           import Unitful
           using Unitful: @unit
           @unit s "weird" Weird 1Unitful.m false
           Unitful.register(UnitfulTest2)
       end
UnitfulTest2

julia> uconvert(u"s", 1*u"twoseconds")
WARNING: Symbol s was found in multiple registered unit modules. We will use the one from UnitfulTest.
2//1 s
```

### DIFF
--- a/src/user.jl
+++ b/src/user.jl
@@ -360,26 +360,20 @@ dottify(s) = s
 dottify() = Main        # needed because fullname(Main) == (). TODO: How to test?
 
 function replace_value(sym::Symbol)
-    # `modules` is a dictionary that maps the value of `module.sym` to the module it
-    # is defined in.  Because we are using a dictionary, identical units will be
-    # automatically combined into one entry.
-    modules = Dict(getfield(m, sym)=>m for m in unitmodules if isdefined(m, sym))
-    filter!((u, m)->ustrcheck_bool(u), modules) # make sure the symbol is a unit
-
-    # One drawback of this approach is that dictionaries are unordered, so if
-    # multiple units with the same name are defined, we can't guarantee which one is
-    # being selected by `first(...)`.
-    u = first(keys(modules))
-    m = modules[u] # the module the unit is defined in
-
-    if length(modules) == 0
+    f = m->(isdefined(m,sym) && ustrcheck_bool(getfield(m, sym)))
+    inds = find(f, unitmodules)
+    isempty(inds) &&
         error("Symbol $sym could not be found in registered unit modules.")
-    elseif length(modules) > 1
+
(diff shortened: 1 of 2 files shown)


ajkeller34:
I updated your PR to address the issue you pointed out. Do you want to take a look and see if you agree with the changes? (btw, your example didn't quite work as advertised, the last warning should say `UnitfulTest2` right?)

P.S. It occurred to me that currently `macroexpand(:(u"m/s"))` yields `:(Unitful.m / Unitful.s)`, but in most cases we could simply return e.g. `m/s`, such that any computations inside are done at parse time. I played around with this and couldn't come up with any non-pathological cases where the change was problematic, but I got cold feet and didn't include the change in this commit.

mweastwood:
I think in my implementation, if there were multiple units, the one that is selected was essentially random and I wouldn't have been surprised to see it change after restarting Julia. I had a comment to that effect.

I realize I broke some of the tests you had written by returning `m/s`. I *think* that change should have been ok in principle, but I agree with your changes as well. I think there are several places in Unitful where it could be doing more work at parse time.

Anyway I'm happy with the changes you've made as well. Looks good to me!

ajkeller34:
I'll just make a separate commit that changes the macro behavior to return `m/s` so that we can revert easily if it turns out to be a bad idea somehow. Thanks!

PR #99, opened 2017-09-15
added convertion to julia's Dates.Time (function, doc string, docs, a…

yakir12:
…nd tests)

This came about because of [this discussion](https://discourse.julialang.org/t/why-do-time-quantities-have-to-be-integers/5864?u=yakir12) on Discourse.

### DIFF
--- a/docs/src/conversion.md
+++ b/docs/src/conversion.md
@@ -71,6 +71,18 @@ julia> uconvert(u"K", 21.0u"°C")
 294.15 K
 ```
 
+### Time conversion
+
+To convert from some time quantity to the time representation native to Julia,
+use `uconvert(Dates.CompoundPeriod, x)`. This allows for the conversion of 
+floats to the integer-based representation in Julia. Note that in some cases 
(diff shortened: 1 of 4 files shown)


ajkeller34:
Do you have a use case for this? There's some uncertainty about the future of Base.Dates, and a lot of bike-shedding in the Discourse thread that's hard to follow. Code added here will cost me time to maintain in the future if anything changes in Base.Dates.

If you think this is a problem worth tackling, bidirectional conversion is a reasonable expectation. You can't convert back from a `Dates.CompoundPeriod` to Unitful quantities at present. I'd like to see that added before I consider merging.

I'm not sure if converting to `ns` first is a good idea. This code will fail with an overflow for integer-valued Unitful times above 293 years or so (e.g. `uconvert(Base.Dates.CompoundPeriod, 52*295u"wk")`). It might be better to convert to days first, subtract the integer number of days off, convert that to hours, etc., but I'm not sure. More strenuous tests would help find the corner cases.

Thanks for the PR.

yakir12:
I see, fair enough. I'll close this now then, and hopefully the problem of using a float time will be solved somewhere else. Thanks for the heads-up about the overflow.

PR #105, opened 2017-10-12
Change printing of default units to match their symbols

### LINKED ISSUE #104: Consistency between unit symbols and display
I just noticed that `u"minute"` prints as `min`, whereas for nearly every other unit we have `u"X"` printing as `X`. I think it will be easier for users to remember the correct symbol for each unit if it matches the printed form.

There are currently only three units in `pkgdefaults.jl` which do not match: 

* `u"in"` -> "inch"
* `u"minute"` -> "min"
* `u"Hz2π"` ->  "Hz*2π"

Would you be open to a PR making those three consistent (by changing the printing, not the symbol)?

rdeits:
Fixes #104 

This changes three units in `pkgdefaults.jl` so that for any unit X `u"X"` prints as `X`. This should make it a bit easier to remember whether the correct symbol for minutes is `u"minute"` and not `u"min"`, and likewise for inches and angular frequency.

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -78,7 +78,7 @@ end
 Unitful.offsettemp(::Unitful.Unit{:Celsius}) = 27315//100
 
 # Common units of time
-@unit minute "min"      Minute      60s                     false
+@unit minute "minute"   Minute      60s                     false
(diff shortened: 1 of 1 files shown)


ajkeller34:
Thanks!

PR #106, opened 2017-11-03
Add slugs

danielmatz:
I noticed that you have lb and lbf defined, but not slugs.

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -163,6 +163,9 @@ Unitful.offsettemp(::Unitful.Unit{:Fahrenheit}) = 45967//100
 # Force
 @unit lbf       "lbf"      PoundsForce  1lb*ge                  false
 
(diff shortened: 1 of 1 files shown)


danielmatz:
I used "sl", but I suppose "slug" could also work, if you'd prefer that.

danielmatz:
I just noticed the order of the units is inconsistent:

```julia
julia> a = 1 * unitful("kg*m/s2")
1 kg m s^-2
julia> uconvert(u"sl*ft/s^2", a)
0.2248089430997105 ft s^-2 sl
```

Is it possible to get sl up to the front of that units expression?

ajkeller34:
Thanks for the contribution. I closed this since it was coincidentally superseded by #115.

Currently, the order in which units are displayed is set by a sorting algorithm that doesn't take "neatness" into account. I agree it should probably sort mass units first.

PR #109, opened 2017-12-02
WIP: Nightly fixes

### LINKED ISSUE #101: Unitful broken on nightly
Just fyi:

```julia
julia> Pkg.add("Unitful")
INFO: Initializing package repository /Users/tamasnagy/.julia/v0.7
INFO: Cloning METADATA from https://github.com/JuliaLang/METADATA.jl
INFO: Installing Unitful v0.3.0
INFO: Package database updated

julia> using Unitful
INFO: Precompiling module Unitful.
WARNING: importing deprecated binding Base.Range into Unitful.
WARNING: Base.Range is deprecated, use Base.AbstractRange{T} where T instead.
  likely near /Users/tamasnagy/.julia/v0.7/Unitful/src/Unitful.jl:857
WARNING: Base.Range is deprecated, use Base.AbstractRange{T} where T instead.
  likely near /Users/tamasnagy/.julia/v0.7/Unitful/src/Unitful.jl:858
WARNING: Base.Range is deprecated, use Base.AbstractRange{T} where T instead.
  likely near /Users/tamasnagy/.julia/v0.7/Unitful/src/Unitful.jl:859
WARNING: could not import Intrinsics.sqrt_llvm_fast into Unitful
ERROR: LoadError: LoadError: MethodError: Cannot `convert` an object of type Expr to an object of type Symbol
This may have arisen from a call to the constructor Symbol(...),
since type constructors fall back to convert methods.
Stacktrace:
 [1] setindex!(::Dict{Symbol,Tuple{Float64,Rational{Int64}}}, ::Tuple{Float64,Int64}, ::Expr) at ./dict.jl:418
 [2] include_relative(::Module, ::String) at ./loading.jl:526
 [3] include at ./sysimg.jl:14 [inlined]
 [4] include(::String) at /Users/tamasnagy/.julia/v0.7/Unitful/src/Unitful.jl:2
 [5] include_relative(::Module, ::String) at ./loading.jl:526
 [6] include(::Module, ::String) at ./sysimg.jl:14
 [7] anonymous at ./<missing>:2
while loading /Users/tamasnagy/.julia/v0.7/Unitful/src/pkgdefaults.jl, in expression starting on line 188
while loading /Users/tamasnagy/.julia/v0.7/Unitful/src/Unitful.jl, in expression starting on line 1044
ERROR: Failed to precompile Unitful to /Users/tamasnagy/.julia/lib/v0.7/Unitful.ji.
Stacktrace:
 [1] compilecache(::String) at ./loading.jl:654
 [2] _require(::Symbol) at ./loading.jl:468
 [3] require(::Symbol) at ./loading.jl:320

julia> versioninfo()
Julia Version 0.7.0-DEV.1878
Commit 6e439ef750* (2017-09-20 21:41 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin16.7.0)
  CPU: Intel(R) Core(TM) i7-4850HQ CPU @ 2.30GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, haswell)
Environment:
```

richardreeve:
Will fix #101. I was having problems with one of my packages on nightly because `using Unitful` errors on nightly/0.7, so I've fixed the errors. In particular, minor issues (mostly warnings) were:

 - `Range` -> `AbstractRange`
 - `(a...)` -> `(a...,)`
 - escape * in strings
 - `e` -> `ℯ`
 - `parse` -> `Meta.parse`
 - `current_module()` -> `@__MODULE__`
 - `Base.Test` -> `Compat.Test` and require Compat v0.33.0 to allow this
 - `sqrt_llvm_fast` no longer exists - use `sqrt_llvm`
 
And then there were major fixes to several macros:
 - `@unit`
 - `@prefixed_unit_symbols`
 - `@unit_symbols`

Unfortunately 0.6 is broken now - contrary to what I thought when making this PR as I messed up my testing. I can't get consistent behaviour between 0.6 and nightly. When one generates the right symbols, the other always produces nonsense. It's very frustrating - suggestions are welcome!

There are also still some errors when doing `Pkg.test("Unitful")` on nightly, particularly when creating new units. This is the biggest change of behaviour I see (though there are others):
```julia
@unit altL "altL" altLiter 1000*cm^3 true
```
What gets created now is `Unitful.altL` not `altL`, so I've added `import Unitful.altL` to the macro.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.6
+Compat 0.33.0
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -1,5 +1,6 @@
 __precompile__(true)
 module Unitful
+using Compat
 
 import Base: ==, <, <=, +, -, *, /, //, ^
 import Base: show, convert
--- a/src/dimensions.jl
+++ b/src/dimensions.jl
@@ -1,5 +1,8 @@
 """
-    *(a0::Dimensions, a::Dimensions...)
+```
+*(a0::Dimensions, a::Dimensions...)
+```
+
 Given however many dimensions, multiply them together.
 
 Collect [`Unitful.Dimension`](@ref) objects from the type parameter of the
--- a/src/dimensions.jl
+++ b/src/dimensions.jl
@@ -1,5 +1,5 @@
 """
-    \*(a0::Dimensions, a::Dimensions...)
+    *(a0::Dimensions, a::Dimensions...)
 Given however many dimensions, multiply them together.
 
 Collect [`Unitful.Dimension`](@ref) objects from the type parameter of the
@@ -54,7 +54,7 @@ true
         end
     end
 
-    d = (c...)
+    d = (c...,)
     :(Dimensions{$d}())
 end
 
--- a/src/dimensions.jl
+++ b/src/dimensions.jl
@@ -72,9 +72,17 @@ end
 # Exponentiation is not type-stable for `Dimensions` objects in many cases
 ^(x::Dimensions{T}, y::Integer) where {T} = *(Dimensions{map(a->a^y, T)}())
 ^(x::Dimensions{T}, y::Number) where {T} = *(Dimensions{map(a->a^y, T)}())
-@generated function Base.literal_pow(::typeof(^), x::Dimensions{T}, ::Type{Val{p}}) where {T,p}
-    z = *(Dimensions{map(a->a^p, T)}())
-    :($z)
+
+@static if VERSION < v"0.7.0-DEV.843"
+    @generated function Base.literal_pow(::typeof(^), x::Dimensions{T}, ::Type{Val{p}}) where {T,p}
+        z = *(Dimensions{map(a->a^p, T)}())
+        :($z)
+    end
+else
+    @generated function Base.literal_pow(::typeof(^), x::Dimensions{T}, ::Val{p}) where {T,p}
+        z = *(Dimensions{map(a->a^p, T)}())
+        :($z)
+    end
 end
 
 # Since exponentiation is not type stable, we define a special `inv` method to enable fast
--- a/src/display.jl
+++ b/src/display.jl
@@ -59,18 +59,6 @@ function show(io::IO, x::Quantity)
 end
 
 """
-    show{T,D,U}(io::IO, ::Type{Quantity{T,D,U}})
-Show the type of a unitful quantity in a succinct way. Otherwise,
-array summaries are nearly unreadable.
-"""
-function show(io::IO, ::Type{Quantity{T,D,U}}) where {T,D,U}
-    print(io, "Quantity{", string(T),
-            ", Dimensions:{", string(D()),
-            "}, Units:{", string(U()), "}}")
-    nothing
-end
-
-"""
     show(io::IO, x::Unitlike)
 Call [`Unitful.showrep`](@ref) on each object in the tuple that is the type
 variable of a [`Unitful.Units`](@ref) or [`Unitful.Dimensions`](@ref) object.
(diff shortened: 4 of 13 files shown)


richardreeve:
0.6 now works pretty well, but only thanks to some VERSION-specific fixes. In particular 0.6 macros seem to require more escaping `$(esc(symb))` than 0.7 macros `$symb`, and 0.7 macros seem to add modules to objects as they are created, so `sr` becomes `Unitful.sr`, which is a problem when you are creating macros outside the `Unitful` package. This still causes errors in 0.6 tests right at the end:

```julia
julia> @dimension f "f" FakeDim12345
f

julia> @derived_dimension FakeDim212345 f^2

julia> @refunit fu "fu" FakeUnit12345 f false
ERROR: UndefVarError: f not defined
```

ajkeller34:
Thank you for doing this. I had been waiting for dust to settle in 0.7 before attempting to fix, since macro hygiene seems to have changed (possibly related to  https://github.com/JuliaLang/julia/issues/23221 ?), and getting the macros right can be subtle. I'll take a look and see what I can do regarding your last point, but I can't give up 0.6 support until 0.7 is released.

richardreeve:
Yes, that is definitely the problem that required escaping changes between 0.6 and 0.7. That's for spotting it - I felt like I was going mad! And I certainly don't want to stop 0.6 working - that would be much worse than the positives for 0.7. However if the scoping issue (or whatever it is) can be addressed then both would work for the time being I think. I'm far from a macro expert, so any suggestions would be great. At the end of the day I can just go back to not using Unitful for the time-being, but one of my PhD students is using it quite heavily both on its own and through AxisArrays and I was trying to make things easier for her as we slowly switch over to 0.7.

ajkeller34:
There are still some fixes remaining to be made but I think I may have sorted out the macro problems at least.

richardreeve:
I have to admit I’m glad you’re doing this, not me, thanks! I’m still struggling to understand why your fix for the macros means we don’t need a difference between the 0.6 and 0.7- code, but I see it works. Also, how are you identifying these PRs - did you know about them already or are you looking at the julia source to see when changes happened?

ajkeller34:
I don't know the new escaping rules very well, but fortunately in my case it turns out I can escape everything the macros emit. With the `esc` on the very outside of the expression, there's not much difference between 0.6 and 0.7. I just make sure to be explicit with module names where needed, like when I'm extending functions defined in Unitful. The comments on the macro thread I linked to previously suggest that things work better if the macro escaping is moved to the very outside of expressions (somewhat controversially, it seems). `macroexpand` helps for diagnostic purposes.

A lot of Unitful is just extending functions from Base by defining new methods that act on `Quantity` types. There's a good chance that if one of the methods I've defined no longer works, it is because a similar method defined for numeric types in Base has changed. On Github I usually just look for a similar method in Base, then click the blame button, which identifies which commit changed a given line of code. Most of the time this works. Sometimes you just have to search the merged PRs using keywords relevant to the problem you're encountering, like I did to find that macro issue.

Now, to get the `VERSION` number according to a merged PR in julia is not so obvious. If you have the merge commit hash of a PR and you have julia built from source, navigate to the julia folder, then for example run `contrib/commit-name.sh 259996c07e87fc3d2da8f05d4dfa405cdf45fbbe` and you will get `0.7.0-DEV.843` in return. This example was for PR 22475.

ajkeller34:
Thanks again for your help with this!

richardreeve:
Fantastic! Obviously that works great for me. Sorry not to have been more help...

Issue #111, opened 2017-12-05
Some range constructions don't work as expected


ajkeller34:
Example:

```
julia> using Unitful:°

julia> 0:10°:350°
ERROR: MethodError: no method matching colon(::Int64, ::Quantity{Int64, Dimensions:{}, Units:{°}}, ::Quantity{Int64, Dimensions:{}, Units:{°}})
```

timholy:
I'm sure you noticed this, but
```julia
julia> 0°:10°:350°
0 °:10 °:350 °
```
This seems to be about a special property of 0.

timholy:
Related (from https://github.com/Keno/SIUnits.jl/issues/113):
```julia
using Unitful: mm
range(1mm, 0.1mm, 50)
```

ajkeller34:
Resolved by https://github.com/ajkeller34/Unitful.jl/commit/3d5fbb79e6f4339358725e500f3884bbb469a71c

PR #113, opened 2017-12-06
Some fixes for #111

ajkeller34:
I haven't merged this immediately because it'd be nice to see if there is fallout from these changes in other packages that make use of unitful ranges. Existing tests pass and the issues noted in #111 are resolved, but the test coverage is likely incomplete.


ajkeller34:
Thanks for taking a look at this. I was at a loss to think of some general rule that made sense beyond just using the mechanisms of promotion. At least ranges should work generally now [, he said, tempting fate]

PR #114, opened 2017-12-06
Allow logarithmic display of any dimension

jebej:
Currently things like `@dB 1m/m` can be created, but cannot be displayed to to this missing method.


ajkeller34:
This wasn't included because I don't have a mechanism for deciding if a given ratio is of power quantities or root-power quantities. My worry is that the returned number in `dB` may be incorrect for some dimensions with a generic fallback like this.

PR #118, opened 2018-01-07
Promote some units to Int64

ChrisRackauckas:
I assume that all combinations are built at compile time? I typed the largest integers to see what would happen.


ChrisRackauckas:
Hmm, is there any way to introspect the macros to find out which combination is failing?

ajkeller34:
Tbh, I usually just "bisect" by commenting out half the file, which is probably beyond dumb, but I'm not sure what the correct way to do it is.

PR #120, opened 2018-02-13
fix a usage error

jingpengw:


ajkeller34:
I think the original text was an unfortunate language choice; the "using" was outside the backticks, so I meant to say that the user could just use whatever was inside the backticks. I didn't mean to imply that you should type `using`. Could you put the `const` inside the backticks to make that clear?

jingpengw:
ah, make sense. done!

ajkeller34:
Thanks!

PR #129, opened 2018-04-05
Proposal of additional units used in electrical engineering

christiankral:
First of all, thanks for providing the package Unitful. I use it a lot, even when teaching my classes. I am using Julia and Unitful for the programming and scripting of electrical engineering applications. So I am not a high level programmer of Julia. I yet hope my proposal is formally correct and it makes it to one of the next versions of Unitful.  

I would like to propose some more units. Most of them are frequently used in electrical engineering, particularly for electric drives. I even added two more units "percent" and "permille" which make sense to non engineers as well. 

Electrical engineers use three different power terms when it comes to AC systems. I think it is worth explaining these different terms:  

- Real power, unit "W", see IEC, http://www.electropedia.org/iev/iev.nsf/display?openform&ievref=131-11-42
- Apparent power, unit "VA", see IEC, http://www.electropedia.org/iev/iev.nsf/display?openform&ievref=131-11-41 
- Reactive power, unit "var", see IEC, http://www.electropedia.org/iev/iev.nsf/display?openform&ievref=131-11-44

The problem with unit "var" is here, that the function ```var``` is in Julia already used to calculate the variance of a vector or array. So there is obviously a name conflict. Possible solutions are:
1. Use unit "VA" instead of "var" as suggested by IEC, see http://www.electropedia.org/iev/iev.nsf/display?openform&ievref=131-11-44; this is the simplest solution and it is fully backwards compatible to any future implementation of "var", whenever it may come.  
2. Use the non-standard unit "VAR" instead of "var", knowing that this is not standard compliant; this is possibly no good idea, since this change were non-backwards compatible to a future implementation of the unit "var"
3. Propose to rename function ```var``` to a different name, e.g., ```variance```. This will possibly cause a lot of trouble for the developers of the function ```var``` since this were a non-backwards compatible change. 

I would conclude to go for option 1 and thus to not implement a separate unit of the reactive power. There should then be a note explaining why "VA" shall be used for the reactive power, not "var".

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -60,9 +60,12 @@ end
 # SI and related units
 @unit Hz     "Hz"       Hertz       1/s                     true
 @unit N      "N"        Newton      1kg*m/s^2               true
+@unit Nm     "Nm"       NewtonMeter 1N*m                    true
 @unit Pa     "Pa"       Pascal      1N/m^2                  true
 @unit J      "J"        Joule       1N*m                    true
 @unit W      "W"        Watt        1J/s                    true
+@unit VA     "VA"       VoltAmpere  1J/s                    true
+# @unit var     "var"     VoltAmpereReactive  1J/s            true
 @unit C      "C"        Coulomb     1A*s                    true
 @unit V      "V"        Volt        1W/A                    true
 @unit Ω      "Ω"        Ohm         1V/A                    true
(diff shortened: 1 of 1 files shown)


ajkeller34:
Thank you for your contribution. There are some subtleties to what you are proposing that I'd like to mention, along with some more general comments.

1. It could be a little dangerous to define `VA` to be equivalent to `W` (both equal 1 J/s). Unitful has no way of knowing that `VA` and `W` are not equivalent, and will implicitly convert between the two without knowledge of power factors. You could try to make an array of watt and VA quantities and it will convert them all to common units, for example. Have you considered defining an electrical engineering units package along the lines of UnitfulUS.jl, UnitfulAstro.jl, UnitfulAngles.jl, etc.? You may have specialized code to handle W and VA more carefully but if VA were added to Unitful I suspect it'd be pretty easy to get incorrect answers silently.

2. I don't think it makes sense to have `Nm` be its own unit. You can already multiply `N` and `m`, and unlike `VA`, I'm not aware of any special meaning that is implied in a particular context. By that I mean `VA` can mean *apparent* power in EE, which would not be obvious from knowing what `V` and `A` mean and multiplying those. If there is a context-specific meaning, then probably`Nm` should be implemented in a separate units package.

3. I am not sure how to think about units like `rpm`. At first, it seemed more natural to me to define `revolution` and then just divide by seconds or minutes. Presumably revolution would be dimensionless. I guess it would be equal to `2pi` radian, since one revolution of a circle would be a turn of 2pi radians. But then, if I convert 1rps to Hz, I would get 2pi Hz, which is a consequence of having defined 1rad equal to 1. I'm not really sure what the best way to handle this is. Maybe what you've done is fine. If you made a PR with just rpm, rps, percent, and permille, I could add those and see if anyone has issues.

christiankral:
This PR is now obsolete, creating a new pull request #144 with just rpm, rps, percent, and permille.

PR #132, opened 2018-04-17
Sort output display by exponent

rafaqz:
After using this for a week I think it's really worth including in Unitful as a the default. Its easier on eyes and brains trained to think in English "units per unit".


ajkeller34:
Thanks for your contribution. I agree after trying this out for a while.

PR #141, opened 2018-06-20
Update to 0.7

timholy:


timholy:
Argh, clicking too fast for github to update the parent repo...

PR #142, opened 2018-06-20
Make the leap to 0.7

timholy:
This was mostly femtocleaner, so it removes some backwards compatibility. Consequently you should merge this only when you're ready to commit to 0.7.

Requires https://github.com/JuliaLang/julia/commit/3aef0e53858c82d5751a991a30d967b27519e13f.

The last two patches are most interesting. 654c17d contains the stuff I needed to do manually after femtocleaner was done, CC @Keno. I'll make a line note about one weird issue that is surely a bug in julia.

The last patch is due to more subtle changes in Julia (mostly, I think, no longer falling back to calling `convert` for constructors).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,11 @@
+- v0.8.0
+  - Add Rydberg constant, unified atomic mass unit, mils, rpm, rps, percent, permille.
+  - Introduce/rename derived dimensions: ElectricalConductivity, ElectricalResistivity,
+    ElectricalConductance, ElectricalResistance.
+  - Fix some Julia 0.7 deprecations.
+  - This will probably be the last release that supports Julia 0.6.
+- v0.7.1
+  - Bug fixes, mainly.
 - v0.7.0
   - Implement `mod2pi` for degrees, cleanup display of degree units.
   - Tweak implementation of `Gain` types for usability.


src/fastmath.jl
```diff
@@ -173,10 +173,8 @@ for f in (:cos, :sin, :tan)
     end
 end
 
-atan2_fast(x::Quantity{Float32,D,U}, y::Quantity{Float32,D,U}) where {D,U} =
-    ccall(("atan2f",libm), Float32, (Float32,Float32), x.val, y.val)
-atan2_fast(x::Quantity{Float64,D,U}, y::Quantity{Float64,D,U}) where {D,U} =
-    ccall(("atan2",libm), Float64, (Float64,Float64), x.val, y.val)
+atan_fast(x::Quantity{T,D,U}, y::Quantity{T,D,U}) where {T,D,U} =
+    atan_fast(x.val, y.val)
```
timholy: This change is a good idea on its own merits, but curiously it was also giving the wrong answer. `@code_llvm` suggested it was using a different pointer address than the one it uses if you write it this way. I'll try to isolate this.

timholy:
You may also want to consider whether to make any other 0.6 releases before merging this.

ajkeller34:
Thanks, as always, for your help with Unitful! I'm pretty swamped this week but hope to take a closer look soon, likely this weekend.

ajkeller34:
I've rebased this against master and fixed a few other things. I see this is your master branch, do you mind if I force push to it? Otherwise I can make a new branch + PR but I plan on merging immediately if the tests pass.

timholy:
Go for it! Thanks for polishing this up!

ajkeller34:
Hmm... it seems to have closed itself when I force-pushed to your branch, and I cannot reopen it. I must admit I wasn't expecting that. I'll just open a new PR I guess.

PR #144, opened 2018-06-30
Proposal of additional units rpm, rps, percent, and permille

christiankral:
This PR is based on https://github.com/ajkeller34/Unitful.jl/pull/129#issuecomment-399709893


christiankral:
Thanks for considering my contribution.

PR #146, opened 2018-07-03
fix some mapreduces in v0.7

ChrisRackauckas:


ajkeller34:
This will prevent master from working on the 0.7-beta release but I'm fine with merging it if you'd prefer, since we have a tagged version that works.

ChrisRackauckas:
I've given up on cross compatibility.

PR #152, opened 2018-09-03
Fix AppVeyor build script for Julia ≥ 0.7

sostock:
This fixes the AppVeyor build script which currently fails for Julia 1.0 since `versioninfo` has been moved to a stdlib package.


sostock:
AFAICT, this is why the tests on the other two pull requests fail.

giordano:
It's probably better to switch to https://github.com/JuliaCI/Appveyor.jl

ajkeller34:
Thanks for bringing this to my attention—closing in favor of #156.

Issue #133, opened 2018-04-17
Getting compilation warning during precompile


djsegal:
I use the following line of code in one of my packages:

```julia 
const atm_conversion =
  uconvert(u"atm", 1e20u"keV/m^3") / 1u"atm"
```

However, this gives the following warnings on the initial compile:

```julia
WARNING: eval from module Unitful to FusionSystems:
Expr(:call, :^, Unitful.FreeUnits{(Unitful.Unit{:Meter, Unitful.Dimensions{(Unitful.Dimension{:Length}(power=Base.Rational{Int64}(num=1, den=1)),)}}(tens=0, power=Base.Rational{Int64}(num=1, den=1)),), Unitful.Dimensions{(Unitful.Dimension{:Length}(power=Base.Rational{Int64}(num=1, den=1)),)}}(), 3)::Any
  ** incremental compilation may be broken for this module **

WARNING: eval from module Unitful to FusionSystems:
Expr(:call, :/, Unitful.FreeUnits{(Unitful.Unit{:eV, Unitful.Dimensions{(Unitful.Dimension{:Length}(power=Base.Rational{Int64}(num=2, den=1)), Unitful.Dimension{:Mass}(power=Base.Rational{Int64}(num=1, den=1)), Unitful.Dimension{:Time}(power=Base.Rational{Int64}(num=-2, den=1)))}}(tens=3, power=Base.Rational{Int64}(num=1, den=1)),), Unitful.Dimensions{(Unitful.Dimension{:Length}(power=Base.Rational{Int64}(num=2, den=1)), Unitful.Dimension{:Mass}(power=Base.Rational{Int64}(num=1, den=1)), Unitful.Dimension{:Time}(power=Base.Rational{Int64}(num=-2, den=1)))}}(), Unitful.FreeUnits{(Unitful.Unit{:Meter, Unitful.Dimensions{(Unitful.Dimension{:Length}(power=Base.Rational{Int64}(num=1, den=1)),)}}(tens=0, power=Base.Rational{Int64}(num=3, den=1)),), Unitful.Dimensions{(Unitful.Dimension{:Length}(power=Base.Rational{Int64}(num=3, den=1)),)}}())::Any
  ** incremental compilation may be broken for this module **
```

How do I prevent it?

ajkeller34:
[...]

PR #161, opened 2018-09-25
Try u_str without eval

tshort:
This is to help with precompiling (#133). It's generally frowned upon to use `eval` in a macro. This is an attempt to return the expression rather than `eval`ing it. The main difference is that `u"N*m"` resolves to `Unitful.N * Unitful.m` whereas now, it resolves to the result of that multiplication.

Making `u_str` work with precompilation will make this macro useful in other packages (most other packages are going to want precompilation).

I tested this in a small package, and it does get rid of the `eval` warnings related to precompilation.

### DIFF
--- a/src/user.jl
+++ b/src/user.jl
@@ -482,7 +482,7 @@ function replace_value(ex::Expr)
                 ex.args[i]=replace_value(ex.args[i])
             end
         end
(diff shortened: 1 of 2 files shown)


ajkeller34:
Sorry for the delay, I think this is a reasonable approach (I rebased and force pushed).

PR #163, opened 2018-09-30
fix one specialization

ChrisRackauckas:
Either I'm not understanding something here or there's a bug. MWE:

```julia
using Unitful
a = [1.0u"s",1u"kg"]
eltype(a) <: Quantity{T,D,U} where {U,D,T} # True
(Quantity{T,D,U} where {T,D,U}) <: Number # True

Base.one(x::Type{Quantity{T,D,U}}) where {T,D,U} = one(T)
@which one(eltype(a)) # one(::Type{T}) where T<:Number in Base at number.jl:297
```

I don't understand how that could end up at the dispatch for Number since Quantity is a subtype

`eltype(a) # Quantity{Float64,D,U} where U where D`

`(Quantity{Float64,D,U} where U where D) <: Quantity{T,D,U} where {T,D,U} # true`

Anyways, this PR gives a workaround which makes sure that `one` is correctly specialized.

### DIFF
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -254,7 +254,8 @@ zero(x::Quantity) = Quantity(zero(x.val), unit(x))
 zero(x::Type{Quantity{T,D,U}}) where {T,D,U} = zero(T)*U()
 
 one(x::Quantity) = one(x.val)
-one(x::Type{Quantity{T,D,U}}) where {T,D,U} = one(T)
+get_T(::Type{Quantity{T}}) where T = T
+one(x::Type{<:Quantity}) = one(get_T(x))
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
@JeffreySarnoff has a better solution:

```julia
julia> using Unitful
julia> oneunit = 1u"s"; twounits = [1u"s",1u"kg"]; twosame = [1u"s", 2u"s"];

julia> test(::Type{T}) where {T<:Number} = "Number";
julia> test(::Type{Q}) where {T,D,U,Q<:Quantity{T,D,U}} = "Quantity";

julia> test(eltype(oneunit)), test(eltype(twosame)), test(eltype(twounits))
("Quantity", "Quantity", "Number")
```

Issue #164, opened 2018-09-30
Unit type computations 


ChrisRackauckas:
It would be nice to be able to compute on the type directly:

```julia
using Unitful
uType = typeof(1u"kg")
tType = typeof(1u"s")
uType/tType
```

This would allow the computation of where types as well:

```julia
using Unitful
uBottomEltype = Unitful.Quantity{Float64,D,U} where U where D
tType = typeof(1u"s")
uBottomEltype/tType
```

Currently, handling recursive arrays is incompatible with Unitful due to the inability to do this calculation.

ajkeller34:
In principle I agree that such a calculation should be possible at compile time. However, it feels like a pun to define division on types, and may also count as type piracy. Jeff has signaled that overloading `Base.show(::IO, ::Type{<:Quantity})` is somehow not great for reasons I'm not clear on (see e.g. https://github.com/Keno/SIUnits.jl/issues/113#issuecomment-348036422), so I imagine `Base. /(::Type{<:Quantity}, ::Type{<:Quantity})` would be objectionable as well. Can you provide more information on what you're trying to do with recursive arrays?

ChrisRackauckas:
I am trying to compute the element type of an array of arrays, and the element type may be a non-concrete Quantity if the array has multiple different unit types.

>However, it feels like a pun to define division on types, and may also count as type piracy. 

No, it's not type-piracy because it would have dispatches on Quantity types only. Overloading Base.show is different and to do it is complicated and that's what's mentioned there, not anything to do with overloading other functions with Type{<:Quantity}.

I think there are better precidents to looks at. You can take an `eltype` of an array type. You can get the `real` part of a type too. For example:

```julia
julia> A = Array{Complex{Float64}}
Array{Complex{Float64},N} where N

julia> eltype(A)
Complex{Float64}

julia> real(eltype(A))
Float64
```

Being able to then similarly do calculations on units would allow one to build arrays before computing all of the values.

ajkeller34:
Yes, those are better examples. I'll try it out.

PR #165, opened 2018-10-01
Begin looking at #164.

ajkeller34:
@ChrisRackauckas A work in progress. Multiplication and division implemented so far (possibly with bugs). Seems like it satisfies your two examples though.

### DIFF
--- a/src/typemath.jl
+++ b/src/typemath.jl
@@ -1,0 +1,26 @@
+for f in (:/, :*)
+    @eval function $f(::Type{S}, ::Type{T}) where {S<:Quantity, T<:Quantity}
+        A = typeof($f(oneunit(S), oneunit(T)))
+        return A
+    end
+    @eval function $f(::Type{Quantity{S}}, ::Type{T}) where {S, T<:Quantity}
+        A = typeof($f(oneunit(S), oneunit(numtype(T))))
+        return Quantity{A}
+    end
+    @eval function $f(::Type{T}, ::Type{Quantity{S}}) where {S, T<:Quantity}
+        A = typeof($f(oneunit(numtype(T)), oneunit(S)))
+        return Quantity{A}
+    end
+    @eval function ($f(::Type{Quantity{T1,D1}}, ::Type{Quantity{T2,D2,U2}})
+            where {T1,T2,D1,D2,U2})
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
I'm not sure why that would not be inferring.

ajkeller34:
I could also do this one monster method, which seems like overkill, but covers all the cases and is inferrable:

```
for f in (:/, :*)
    @eval @generated function ($f(::Type{Q1}, ::Type{Q2}) where
            {Q1 <: Quantity, Q2 <: Quantity})
        Q1u = unwrap_unionall(Q1) # I've just locally defined what is in base to avoid using an unexpected function
        Q2u = unwrap_unionall(Q2)
        T1, T2 = Q1u.parameters[1], Q2u.parameters[1]
        D1, D2 = Q1u.parameters[2], Q2u.parameters[2]
        U1, U2 = Q1u.parameters[3], Q2u.parameters[3]
        if T1 isa TypeVar || T2 isa TypeVar
            return :($(Quantity))
        else
            A = typeof($f(oneunit(T1), oneunit(T2)))
            if D1 isa TypeVar || D2 isa TypeVar
                return :($(Quantity{A}))
            else
                B = typeof($f(D1(), D2()))
                if U1 isa TypeVar || U2 isa TypeVar
                    return :($(Quantity{A,B}))
                else
                    C = typeof($f(oneunit(Q1), oneunit(Q2)))
                    return :($C)
                end
            end
        end
    end
end
```

src/typemath.jl
```diff
@@ -0,0 +1,90 @@
+# Same as Base.unwrap_unionall, but I don't want to rely on unexported functions.
+function unwrap_unionall(@nospecialize(a))
+    while isa(a, UnionAll)
+        a = a.body
```
ajkeller34: I doubt the `body` field will change in the near future

ajkeller34:
Let me know if operations besides `+`,`-`,`*`,`/` would be useful for your application, up to that this will work.

It would be uglier in some sense to do this without the `@generated` function / `UnionAll` manipulation because the "clean" way I showed above gives inference trouble. In the original way I tried doing this, you would need to write a lot of methods and probably miss some, whereas here the reasoning is at least clear even if the implementation is ugly.

PR #172, opened 2018-10-16
Support isequal with NaN quantities

### LINKED ISSUE #171: Support isequal
`isequal(NaN*u"m", NaN*u"m")` is `false` at the moment, but it should be true.

```julia
help?> isequal
search: isequal issetequal DimensionlessQuantity

  isequal(x, y)

  Similar to ==, except for the treatment of floating point numbers and of
  missing values. isequal treats all floating-point NaN values as equal to each
  other, treats -0.0 as unequal to 0.0, and missing as equal to missing. Always
  returns a Bool value.
```

cstjean:
Fix #171

### DIFF
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -1,6 +1,6 @@
 module Unitful
 
-import Base: ==, <, <=, +, -, *, /, //, ^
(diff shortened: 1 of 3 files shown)


ajkeller34:
Thanks for the fix!

Issue #72, opened 2017-03-28
(0°C)^n should not equal 0*(°C)^n


ajkeller34:
Same for `°F`. Better to convert `°C` to an absolute scale like `K` before taking the power, if it is the only unit on the quantity.

Not sure if better to convert `°F` to `°Ra` or `K` since both are absolute scales, and `°Ra` is more closely related to `°F`...

deszoeke:
Are you proposing that taking a power should always use an absolute temperature? Then the power operator would not have the expected relationship to multiplication for `Temperature`s.

This should also be done for dividing (the power is -1) or multiplying by temperature in many cases. In the ideal gas law, for example,
```
density=p/Rspecific*T^-(1//1), or
volume=Nmoles*Runiversal*T^(1//1)/p,
```
where T must be an absolute temperature. Yet this might not be universally desirable.

The proposed behavior is good for physics, but could lead to unexpected results, e.g. for evaluating polynomial fits linearized about a different temperature scale, like an empirical thermal response function of the form `r = a0+ a1*T + a2*T^2 + a3*T^3...` with T in °C.

Perhaps Unitful should always use absolute temperature upon (multiplication, division, or) taking a power and this empirical formula should be constructed from powers of `NoUnits` as 
`r = a0 + a1*NoUnits(T/u"°C") ...`. This would work as long as the ratio `T/u"°C" = u"°C"(T)/u"°C"` is not interpreted as absolute `u"K"(T)/u"°C"`, which this proposal seems dangerously close to doing.

ajkeller34:
I would write your example as `r = ((1u"°C")/a_0) + ((T-T_0)/a_1)^1 + ((T-T_0)/a_2)^2 + ...`. In this way you can make an empirical fit around `T_0` where `T_0` is any temperature. By dividing by a coefficient before you take the power, the exponentiation would work properly. An added benefit is that all your coefficients now have the same units of temperature^1 and can be stored efficiently in an array. Finally, this expression will work regardless of the units of `T` for given `a_i`.

I am concerned that there may be cases where the wrong answer is given, including the one you point out, but ultimately I need Unitful to be useful for physics. It may be possible to make a custom `°C` unit with some associated logic to indicate whether it should be viewed as a temperature interval or as a temperature on a scale.

deszoeke:
[...]

PR #176, opened 2018-10-19
Separate absolute and relative temperature

### LINKED ISSUE #137: Need better handling of temperatures
Right now, it seems that temperature conversions have been made correct on a case by case basis, but its easy to find cases where things go wrong. For instance,

```julia
julia> uconvert(u"K", 1.0u"°C")
274.15 K

julia> uconvert(u"K^-1", 1.0u"°C^"^-1)
1.0 K^-1
```
if 1°C is 274.15 K, then it should not be the case that 1°C^-1 is 1 K^-1. 

It seems that the handling of temperature conversions needs to be made more general. However, this is a really fraught area. I think this package needs to choose a philosophy and stick to it consistently for offset units. This is a hard choice to make though. For instance, what does it mean to multiply an offset unit by a number? Is 5°C really 5 times warmer than 1°C, or is it only 1.015 times warmer? If one thinks that it should be treated as 5 times warmer, then one then has to commit to the position that 1°C is infinitely warmer than 0°C which soon becomes a problematic position. It's really not obvious what the optimal choice is to make here, but I think that whatever choice is made should at least be well documented and consistent. 

My suggestion would be to always internally represent offset units on an absolute scale but then display them as offset units such that 5°C is only 1.015 times warmer than 1°C but provide an opt-in option to treat them as if they were absolute if that's what a user needs.

cstjean:
Following the discussion in #137, I separated absolute and relative temperature units, similar to @burnpanck's suggestion. It works well so far. Notable new test cases:

```
            @test uconvert(abs°F, 0abs°C) == 32abs°F
            @test uconvert(°F, 0°C) == 0°F
            @test 1.0abs°C + 45°F == 26.0abs°C
            @test 1abs°C - 1K == 0abs°C
            @test 40abs°C - 95abs°F == 5°C
            @test_throws DimensionError 1abs°C + 1abs°F
```

This PR is a proposal for discussion. If the general approach is acceptable, then we might find a less ugly name than `35abs°C`. And of course: documentation.

Regarding binary arithmetic on absolute temperatures, we can either disallow promotion of two absolute temperatures in general, and explicitly allow some operations (`-, ==, ...?`) or allow all operations by default and disallow some (`+, *, /, ...?`) by throwing an exception. I did the latter in this PR, but I also tested the former and it works too.

Fix #137 and #72

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -7,6 +7,7 @@
 @dimension 𝐓 "𝐓" Time
 @dimension 𝐈 "𝐈" Current
 @dimension 𝚯 "𝚯" Temperature    # This one is \mbfTheta
+@dimension abs𝚯 "abs𝚯" AbsTemperature
 @dimension 𝐉 "𝐉" Luminosity
 @dimension 𝐍 "𝐍" Amount
 
@@ -49,6 +50,7 @@ include("temperature.jl")
 @refunit  s       "s"      Second    𝐓           true
 @refunit  A       "A"      Ampere    𝐈            true
 @refunit  K       "K"      Kelvin    𝚯           true
+@refunit  absK    "absK"   AbsKelvin abs𝚯        true
 @refunit  cd      "cd"     Candela   𝐉            true
 @refunit  g       "g"      Gram      𝐌           true
 @refunit  mol     "mol"    Mole      𝐍           true
@@ -89,8 +91,9 @@ end
 @unit permille "‰"      Permille    1//1000                 false
 
 # Temperature
-@unit °C     "°C"       Celsius     1K                      true
-Unitful.offsettemp(::Unitful.Unit{:Celsius}) = 27315//100
+@unit °C     "°C"       Celsius             1K                      true
+@unit abs°C  "abs°C"    AbsCelsius          1absK                   true
+Unitful.offsettemp(::Unitful.Unit{:AbsCelsius}) = 27315//100
 
 # Common units of time
 @unit minute "minute"   Minute                60s           false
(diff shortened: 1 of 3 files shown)


cstjean:
Here's one issue: 

```julia
julia> mean([1abs°C, 2abs°C])
ERROR: DimensionError: 1 abs°C and 2 abs°C are not dimensionally compatible.
Stacktrace:
 [1] +(::Quantity{Int64,Unitful.Dimensions{(Unitful.Dimension{:AbsTemperature}(1.
```

We can fix it manually, but it's a bit worrying.

RuiRojo:
(A nobody giving his opinion)

I wouldn't pay much attention to the `mean` issue. It just happens that the general implementation of `mean` assumes you can add the stuff, while that's not a requirement for a mean to make sense. `mean_v2(x) = mean(x .- x[1]) + x[1]`. 

I think that it would introduce less silent mistakes to make the traditional names (°C, °F, K) default to the absolute ones which IIUC would error more often when misused (you can't add them, multiply, etc). As is, I bet stuff like `uconvert(°F, 0°C)` returning smells like trouble `0°F`. I do get the temptation to make the new weirdos in town (units with offsets) bear the prefix. Maybe it's even worth it to have a short prefix for both relative and absolutes.

cstjean:
Thank you for the comments. I agree that a general solution would be better. 

> The dimension of temperature ought to be the same regardless of absolute / relative scales. 

Conceptually, I agree, but pragmatically, I can define functions with dimensions in the signature, like `is_boiling(::Temperature) = t < 100°C`. That function should not accept relative temperature, but I can't specify that if absolute and relative temperatures share the same dimension.

> I think we could generalize the Units types to permit affine transformations. Perhaps an extra type parameter that could either be nothing (no affine transformation) or a value ("absolute zero" on the scale).

Sounds good. 

The current PR is good enough for our purposes, so I don't know when/if I'll have time to push this further. If anyone wants to take a shot at this, please do.

burnpanck:
This is exciting! I'm not even a Julia user yet, but it's great to see how the language allows library writers to offer tools that do the right thing™, without sacrificing expressiveness. And of course, kudos to you guys! Now if only arrays would start at zero...

ajkeller34:
Closed in favor of #177, discussion can continue there.

PR #177, opened 2018-10-26
[RFC] Prototyping better temperature handling.

ajkeller34:
Alternate approach to #176, will merge in a day if no further objections.

To-do:

- [x] ~~bikeshed symbol names (is `abs°C` tolerable?)~~
  No need for new symbol names. Declare `°C`, `°F` are relative scales and shouldn't be used to represent temperature differences (that's what `K` and `Ra` are for)
- [x] ~~bikeshed how to display quantities that affine transform under unit conversion (e.g. what we've been calling absolute temperatures). Right now they just show `(affine)` after the unit symbols, which is ugly.~~
  No need for different display.
- [x] restricted nonsensical operations for affine quantities (i.e. absolute temperatures): `*`, `/`, `+`, `^`, `inv`, `sqrt`, `cbrt`, `zero` (could be allowed if different type could be returned), `one`, `oneunit`.
- [x] fix operations between absolute, relative temperatures
- [x] make things sensible when `ContextUnits`, `FixedUnits`, etc. are used
- [x] write documentation
- [x] make a macro for defining affine units more easily


ajkeller34:
Ha, I was borrowing the terminology that #176 used and not thinking. You're absolutely right about that, though.

ajkeller34:
Alright, now `°C` and `°F` are affine units; `K` and `Ra` are not. Promotion will often prefer `K` as an answer result but that can be overridden with `ContextUnits` as usual. However, that might not be a good idea because some operations are prohibited with affine quantities. Overall, I would say this seems less fiddly and feels right.

I've switched the terminology around (`K` is an absolute scale). It's late in my timezone so I'm sure there are mistakes to find...

docs/src/index.md
```diff
@@ -64,6 +64,8 @@ DocTestSetup = quote
     using Unitful
     °C = Unitful.°C
     °F = Unitful.°F
+    abs°C = Unitful.abs°C
+    abs°F = Unitful.abs°F
```
cstjean: Those lines are not needed now.

src/quantities.jl
```diff
@@ -253,13 +286,21 @@ for f in (:floor, :ceil, :trunc, :round)
 end
 
 zero(x::Quantity) = Quantity(zero(x.val), unit(x))
-zero(x::Type{Quantity{T,D,U}}) where {T,D,U} = zero(T)*U()
+zero(x::AffineQuantity) =
+    throw(AffineError("no additive identity element of type $(typeof(x))."))
+zero(x::Type{Quantity{T,D,U}}) where {T,D,U<:ScalarUnits} = zero(T)*U()
+zero(x::Type{Quantity{T,D,U}}) where {T,D,U<:AffineUnits} =
+    throw(AffineError("no additive identity element of type $(typeof(x))."))
```
cstjean: Shouldn't that be `0K`? I don't think that `zero(T)` has to be of type T; `one(1u"m")` isn't of the same type either.

ajkeller34: The docstring for `zero` is a bit ambiguous: "get the additive identity *element* for the type of x" (emphasis mine). If it just said additive identity, I would agree. That it uses the word "element" suggests that it is considering the set of allowed values of the type as elements of an additive group with this particular value for the type as the additive identity. It may sound like I'm reading too much into it but I've seen people speak in these terms before in other discussions regarding `zero` and `one`. If you look for usage examples in Base, it is often assumed that `typeof(zero(T)) <: T` in generic code. 

On the other hand, the docstring for `one` mentions explicitly that the result may not be the same type for dimensional quantities. That's why the function `oneunit` also exists.

docs/src/index.md
ajkeller34: Good catch, forgot to update `index.md`.

src/quantities.jl
cstjean: Ok, that's reasonable. FWIW, Base doesn't strictly adhere to that:

```julia
julia> using LinearAlgebra

julia> q = qr(randn(2,2));

julia> typeof(q.Q)
LinearAlgebra.QRCompactWYQ{Float64,Array{Float64,2}}

julia> zero(q.Q)
2×2 Array{Float64,2}:
 0.0  0.0
 0.0  0.0
```
Likewise for [tranpose](https://github.com/JuliaLang/julia/issues/29839). But I can see that going from Celsius to Kelvin would be a bigger leap, in some sense.

ajkeller34: Interesting. I wonder if container types are treated differently. Maybe I’ll ask around...

burnpanck:
One comment on the swap of nomenclature: While I appreciate that "absolute Temperature" has a specific meaning in physics, calling temperature points/coordinates in `°C` _relative_ temperature (as opposed to temperature differences) seems confusing to me. Maybe we should nor use _relative_ vs. _absolute_ to refer to the distinction between points (coordinates) and differences (vectors) at all.

ajkeller34:
Ok, I've tried to be more explicit to say "relative scale" instead of "relative." I think "relative scale" is commonly used and understood terminology but "relative" is indeed an ambiguous term. I'd like to avoid introducing new jargon though and I worry `TemperatureCoordinate` may not convey enough information for all to understand. 

I will merge once tests pass.

PR #187, opened 2018-11-06
Add Velocity to derived dimensions

non-Jedi:
This seems like an obvious omission. Am I missing some context?


ajkeller34:
I never exhaustively went through the possible dimensions. The current list reflects what previous users have needed. Want to add `Acceleration` while you're at it?

non-Jedi:
Sure would. Threw in an extra bonus dimension for you, too, while I was at it.

PR #191, opened 2018-11-14
WIP: Unitful arrays

cstjean:
I believe that the datastructure is fine, but I am unhappy with everything else. My use case is with `StaticArrays`, and there doesn't seem to be a nice way of accommodating both `Array` and `SArray` objects in the unit conversion routines. At least, not without also requiring `StaticArrays`, and at that point, we should probably start a new package entirely. Suggestions are welcome.

cc. @mschauer

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.7
+StaticArrays
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -45,6 +45,7 @@ include("conversion.jl")
 include("range.jl")
 include("fastmath.jl")
 include("logarithm.jl")
+include("arrays.jl")
 include("pkgdefaults.jl")
 
 function __init__()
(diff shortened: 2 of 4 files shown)


alhirzel:
I recently started to encounter issues with arrays of Unitful values, and wanted to present one example. @cstjean, do you think this would be addressable?

```julia
# fit a quadratic to three points, y = 5x^2 + 0.1, x in meters, y in seconds
using Unitful.DefaultSymbols

x = [0.0s  1.0s   2.0s]'
y = [0.2m  5.2m  20.2m]'

A = [x[1]^2 x[1]^1 x[1]^0
     x[2]^2 x[2]^1 x[2]^0
     x[3]^2 x[3]^1 x[3]^0]

# ideally, units would be calculated rather than assigned here
coeffs = ustrip.(A)\ustrip.(y) .* [m/s^2; m/s; m]
@assert all(ustrip.(abs.(coeffs - [5.0m/s^2; 0.0m/s; 0.2m])) .< 1.0e-12)
coeffs
```

cstjean:
It's should be doable with another special array type, but this PR only supports units of the form 
```
[a*x a*y a*z
 b*x b*y b*z
 c*x c*y c*z]
```
There might already be a package that generates arrays of the form your `A` has.

PR #198, opened 2018-11-30
Add derived dimensions for concentration

non-Jedi:
Similar to #187. Let me know if I start getting too obscure with these, but I don't think I'm there yet.

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -44,6 +44,9 @@ const AbsoluteScaleTemperature = Quantity{T, 𝚯, <:ScalarUnits} where T
 @derived_dimension ElectricDipoleMoment     𝐋*𝐓*𝐈
 @derived_dimension ElectricQuadrupoleMoment 𝐋^2*𝐓*𝐈
 @derived_dimension MagneticDipoleMoment     𝐋^2*𝐈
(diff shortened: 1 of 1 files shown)


src/pkgdefaults.jl
```diff
@@ -44,6 +44,9 @@ const AbsoluteScaleTemperature = Quantity{T, 𝚯, <:ScalarUnits} where T
 @derived_dimension ElectricDipoleMoment     𝐋*𝐓*𝐈
 @derived_dimension ElectricQuadrupoleMoment 𝐋^2*𝐓*𝐈
 @derived_dimension MagneticDipoleMoment     𝐋^2*𝐈
+@derived_dimension Concentration            𝐌/𝐋^3
```
ajkeller34: There are many ways to express concentration, some of which don't have these dimensions. Isn't molarity also a measure of concentration?

non-Jedi: Sure is. In my field at least, when someone just talks about concentration, they usually mean mass/volume, so I was including it based on that. If you think this definition creates more ambiguity than it's worth, I can certainly remove that one or rename it `MassConcentration`.

non-Jedi: Removed.

Issue #193, opened 2018-11-16
Occasional seg faults on 0.7/1.0


nicoleepp:
Seg fault when `using ExtensionPkg` before `using Unitful` in another module:

This came up here https://github.com/lanl-ansi/PowerModelsAnnex.jl/issues/29 and was resolved for that package specifically (turning precompilation off for the package that extends Unitful solves the error).

But I noticed that if that package did a `using UnitfulUS` before `using Unitful` it would also cause seg faults when loading `using PowerModelsAnnex`. Not sure the cause exactly but I suspect it might be related to Unitful itself/precompilation in general as this appears not to be specific to that extension of Unitful.

cstjean:
I've had a ridiculous amount of segfaults since I started using Unitful. I don't think there's anything that  can be done about it, besides filing reports with minimal code examples to julialang, unfortunately. Eg. https://github.com/JuliaLang/julia/issues/29936 came from Unitful, and sounds very similar.

nicoleepp:
Even with precompilation disabled, using Unitful on 0.7/1.0 will occasionally result in seg faults.

ajkeller34:
I'd like to offer help, but usually segfaults are a Julia problem, unless it turns out I'm doing something unsafe in Unitful. I can audit my use of unsafe macros like `@inbounds`, etc. but to be honest I haven't had any segfaults. I also don't use separate unit packages that often, so maybe that points to the source of the problem?

ajkeller34:
I'd be curious to know if the segfault situation has improved after merging https://github.com/PainterQubits/Unitful.jl/pull/243. I don't have evidence to claim the issues are related although there were some unsafe uses of generated functions prior to that PR.

cstjean:
FWIW I haven't had segfaults in a long time...

omus:
We've still been seeing segfaults on our private CI. I'll set it up to make use of the `master` branch for now and report back.

omus:
Yesterday we saw a segfault on the CI using the Unitful `master` branch. We were also using the [PowerSystemsUnits.jl](https://github.com/invenia/PowerSystemsUnits.jl) which was using [v0.1.0](https://github.com/invenia/PowerSystemsUnits.jl/releases) which may have been the cause of the segfault. We'll try to confirm this by using the v0.2.0 release.

omus:
We haven't seen any segfaults with updated version of PowerSystemsUnits.jl. I think it's safe to close this issue. We can re-open if they appear again.

PR #201, opened 2018-12-12
Make u_str work with precompilation

c42f:
I was a little disheartened to see that the (hugely convenient and awesome) `u_str` is not recommended for use with precompilation when used with unit plugin modules. I *think* the following approach fixes that: I've just added a check that the desired unit extension module is explicitly loaded into the module before the u_str macro is invoked.

Demo:

```julia
julia> using UnitfulUS

julia> module A
           using Unitful
           foo() = 1u"gal_us"
       end
ERROR: LoadError: Symbol `gal_us` was found in unit module UnitfulUS, but was not loaded into Main.A. Consider `using UnitfulUS` within `Main.A`?
Stacktrace:
 [2] replace_value(::Module, ::Symbol) at /home/tcfoster/.julia/dev/Unitful/src/user.jl:540
 [3] @u_str(::LineNumberNode, ::Module, ::Any) at /home/tcfoster/.julia/dev/Unitful/src/user.jl:495
in expression starting at REPL[2]:3

julia> module A
           using Unitful, UnitfulUS
           foo() = 1u"gal_us"
       end
WARNING: replacing module A.
Main.A

julia> A.foo()
1 galᵘˢ
```

### DIFF
--- a/src/user.jl
+++ b/src/user.jl
@@ -471,9 +471,6 @@ will be used.
 Note that what goes inside must be parsable as a valid Julia expression.
 In other words, u"N m" will fail if you intended to write u"N*m".
 
-Since `@u_str` dynamically looks through registered modules, it does not play well with
-precompilation. You shouldn't use this macro in a precompiled module.
-
 Examples:
 
 ```jldoctest
@@ -495,25 +492,25 @@ julia> u"ħ"
 """
 macro u_str(unit)
     ex = Meta.parse(unit)
-    esc(replace_value(ex))
+    esc(replace_value(__module__, ex))
 end
 
 const allowed_funcs = [:*, :/, :^, :sqrt, :√, :+, :-, ://]
-function replace_value(ex::Expr)
+function replace_value(targetmod, ex::Expr)
     if ex.head == :call
         ex.args[1] in allowed_funcs ||
             error("""$(ex.args[1]) is not a valid function call when parsing a unit.
              Only the following functions are allowed: $allowed_funcs""")
         for i=2:length(ex.args)
             if typeof(ex.args[i])==Symbol || typeof(ex.args[i])==Expr
-                ex.args[i]=replace_value(ex.args[i])
+                ex.args[i]=replace_value(targetmod, ex.args[i])
             end
         end
         return ex
     elseif ex.head == :tuple
         for i=1:length(ex.args)
             if typeof(ex.args[i])==Symbol
-                ex.args[i]=replace_value(ex.args[i])
+                ex.args[i]=replace_value(targetmod, ex.args[i])
             else
                 error("only use symbols inside the tuple.")
             end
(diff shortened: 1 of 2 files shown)


c42f:
Updated tests to fix a difference in the exception type thrown in 0.7 vs 1.0.

ajkeller34:
Thanks for looking at this. It seems like it should work, so after the tests pass I'll merge and we can see if there's any unexpected fallout :)

ajkeller34:
This seems to have [broken a few packages](https://s3.ap-south-1.amazonaws.com/juliarun-ci/9eeee137888c2048ca16db387e11bcdc706b40b3/Unitful_reverse_dependency_tests.html). Closer to the end of the week I hope to get a chance to look closer but it looks like your error message is popping up frequently, so some packages must not be doing what you think they should.

c42f:
Hmm, darn. Nothing like trying it out to find your assumptions are a little broken... or alternatively that your implementation has become part of your public API against your best intentions. There's a bit of both here I think ;-)

The `UnitfulAtomic` error is due to having the unit definitions in a submodule - they actually come from `UnitfulAtomic.RegisteredUnits` rather than `UnitfulAtomic` itself. But the tests only do `using UnitfulAtomic`, and quite reasonably so. I think we could patch up the current approach in `Unitful` to also allow unit definitions in submodules of `using`'d modules. Fixing upstream is also possible though, what do you think @sostock?

The `PowerSystemsUnits` errors are due to the tests doing `using Unitful: @u_str`, so the name `Unitful` isn't actually in the namespace at all. Easily patched up in `PowerSystemUnits` itself.

The `PowerModelsAnnex` failure is due to the fact that `PowerSystemsUnits` doesn't register itself in its `__init__`, and `PowerModelsAnnex` does it instead as a workaround, but not in the tests... so I don't know how this ever worked properly.  Quick workaround would be to manually register `PowerSystemsUnits` in the `PowerModelsAnnex` tests, CC @nicoleepp. There's a crossref to #193 in the `PowerSystemsUnits` source.

Can't tell what's wrong with `Modea` at a quick glance... looks like they're duplicating `@u_str` with their own `@U_str`, perhaps partly for the reasons this PR was trying to fix. So perhaps this change would help them to not do that anymore (or maybe #161 has fixed it already). @tshort any thoughts?

sostock:
Fixing `UnitfulAtomic` is simple: The reason for using the submodule `UnitfulAtomic.RegisteredUnits` is that not all units defined in `UnitfulAtomic` are registered, because their names clash with those in `Unitful`. Simply changing the names of the units in `UnitfulAtomic` (e.g. `ħ` → `ħ_au`) would make it possible to register all units and thus get rid of the submodule. I should probably have done that in the first place, since having both `Unitful.ħ` (a quantity) and `UnitfulAtomic.ħ` (a unit) is kind of confusing. I will update my package accordingly.

Also, I think it would be good to mention this new restriction on `@u_str` usage somewhere in the documentation (even though the error message is pretty clear).

sostock:
I just tagged UnitfulAtomic 0.2.0 (which is compatible with the new `@u_str`) and restricted UnitfulAtomic 0.1.0 to Unitful ≤ 0.13.0.

Issue #186, opened 2018-11-05
[Wishlist] Define an AbstractQuantity type


giordano:
A few months ago I've created a new package to provide physical constants: https://github.com/JuliaPhysics/PhysicalConstants.jl  This package uses `Unitful.jl` and defines a new type which is mostly based on my proposal at #81.  In order to be able to automatically use a `Constant` object in mathematical operations, I think it would be great to have an `AbstractQuantity` type here, that can be subclassed by `PhysicalConstants.jl` (see https://github.com/JuliaPhysics/PhysicalConstants.jl/issues/2).  `Quantity` would be another concrete subtype of `AbstractQuantity`.  I don't think the change would be very disruptive, it'd only be necessary to define the interface of `AbstractQuantity` and use it throughout this package (rather than accessing named fields).

ajkeller34:
Should this issue stay open given that #204 was merged, and if so what remains to be done for your needs?

giordano:
It'd be good to have a getter function for the `val` field just as a nicer interface than `a.val`.  This function would be semantically different from `ustrip`, and probably won't be exported, I don't expect users to need it.

Apart from this, I think this ticket can be closed.

PR #204, opened 2019-01-27
Define AbstractQuantity

giordano:
Start addressing #186 by defining an `AbstractQuantity` type.  <s>I've marked this PR as work-in-progress because I'm not sure I've used the new abstract type wherever it's needed, at least tests pass locally.</s>

### DIFF
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -17,16 +17,16 @@ Quantity(x::Number, y::Units{()}) = x
 *(x::Number, y::Units, z::Units...) = Quantity(x,*(y,z...))
 *(x::Units, y::Number) = *(y,x)
 
-*(x::Quantity, y::Units, z::Units...) = Quantity(x.val, *(unit(x),y,z...))
-*(x::Quantity, y::Quantity) = Quantity(x.val*y.val, unit(x)*unit(y))
+*(x::AbstractQuantity, y::Units, z::Units...) = Quantity(x.val, *(unit(x),y,z...))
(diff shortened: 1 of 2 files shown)


giordano:
I've played with this branch in `PhysicalConstants.jl` and it seems to be sufficient for my current needs: https://github.com/JuliaPhysics/PhysicalConstants.jl/issues/2#issuecomment-457912121

The only thing I would recommend doing is to define an API function (`value`?) to access the value of an `AbstractQuantity`, which for the concrete type `Quantity` could be `value(x::Quantity) = x.val`.  In my tests for `PhysicalConstants.jl` I've implemented `Base.getproperty(c::Constant, sym::Symbol)`

c42f:
> The only thing I would recommend doing is to define an API function to access the value of an AbstractQuantity.

Just wondering - is this just `ustrip`, or does it mean something different?

giordano:
Ustrip is defined as
```julia
@inline ustrip(x::Quantity) = ustrip(x.val)
```
So strictly speaking it doesn't simply access to the val field. However, I'm having difficulties to find a case where the results would be different

c42f:
Yes I haven't read enough of the internals to know why it's recursive, but it seems like it might have the meaning you want.

giordano:
Yeah, anyway I'm not too concerned about that, for me it's sufficient this PR is merged

ajkeller34:
I'm late on the explanation but `ustrip` is recursive because of types like `Quantity{<:Level}`, i.e. `(20u"dBm/Hz").val == 20u"dBm"`, but presumably you want `ustrip(20u"dBm/Hz") == 20`. There is some special sauce needed to make logarithmic quantities work right.

PR #205, opened 2019-01-29
Safer `ustrip`

### LINKED ISSUE #48: Stripping of units
Unitful is an amazing package and seems to me to be the one superseding all the others - I'd like to consider it (after some further maturity) to get part of the Julia "standard packages" ecosystem.
I'm used to the unit system of CLHEP library used in high-energy physics codes (e.g. [manual](http://cmd.inp.nsk.su/old/cmd2/manuals/cernlib/CLHEP/UserGuide/Units/units.html), [git](https://gitlab.cern.ch/CLHEP/CLHEP)), where the workflow is following:

At the input, all values are multiplied by the input unit, all computations are done with units and at the output (or binary interface or ccall to a library), the variable is divided by the desired unit:
```
u = 15u"V"
d = 5u"mm"
E = u/d
outE = E/u"V/m"
>3.0 mm^-1 m
```
I'd like outE v to be an unitless value of  3000.
This would be much more readable than having to write `upreferred(outE)`, or using convert and ustrip functionality. Compare the following easy readable code (as would be in CLHEP)
```
println("The field intensity is $(E/u"V/m") [V/m] in a gap of $(d/u"m") [m]")
>The field intensity is 3000.0 [V/m] in a gap of 0.005 [m]
```
with the current verbosity
```
println("The field intensity is $(ustrip(float(uconvert(u"V/m", E)))) [V/m] in a gap of $(ustrip(float(uconvert(u"m", d)))) [m]")
>The field intensity is 3000.0 [V/m] in a gap of 0.005 [m]
```
Is it possible to specialize somehow the `/` operator in such a way that if dimensions of both operands are same, the units are correctly stripped, giving just a float number?
At least for cases where just different unit multiples of the same basic unit are uses, such as `mm` and `m`, etc.
Thank you for the advice.

c42f:
One approach to fix #48 by providing a safer form for `ustrip`, while not introducing extra conversion rules or special cases.

Also a couple of fixes to documentation generation, which seems to have been broken by recent `Documenter` updates (I guess?)

As a separate step we could consider deprecating the one argument form of `ustrip`, though I guess it's not so easy to deprecate `ustrip` for arrays without copying.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,6 +1,8 @@
 using Documenter, Unitful
 
-makedocs()
+makedocs(
+    sitename = "Unitful",
+)
 
 deploydocs(
     deps   = Deps.pip("Tornado>=4.0.0,<5.0.0", "mkdocs==0.17.5", "mkdocs-material==2.9.4", "python-markdown-math"),
--- a/src/logarithm.jl
+++ b/src/logarithm.jl
@@ -479,11 +479,12 @@ struct InvalidOp end
 Base.show(io::IO, ::InvalidOp) = print(io, "†")
 macro _doctables(x)
     return esc(quote
+        sprint(show,
         try
             $x
         catch
             Unitful.InvalidOp()
-        end
+        end)
     end)
 end
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -7,15 +7,35 @@
 @inline dimtype(u::Unit{U,D}) where {U,D} = D
 
 """
-    ustrip(x::Number)
-    ustrip(x::Quantity)
-Returns the number out in front of any units. The value of `x` may differ from the number
-out front of the units in the case of dimensionless quantities, e.g. `1m/mm != 1`. See
-[`uconvert`](@ref) and the example below. Because the units are removed, information may be
-lost and this should be used with some care.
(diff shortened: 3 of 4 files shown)


c42f:
Bump. I'd like to progress this if possible?

BeastyBlacksmith:
Would it be a good or bad idea to have it take `Number`s instead of `Quantity`s? Since if you have generic code you might as well encounter unit-less numbers and then it errors.

c42f:
Yes, good point; in particular, the following shouldn't be an error:

```
julia> ustrip(NoUnits, 1.0u"s"/u"s")
ERROR: MethodError: no method matching ustrip(::Unitful.FreeUnits{(),NoDims,nothing}, ::Float64)
```

especially since there's no problem for the slightly less trivial version:
```
julia> ustrip(NoUnits, 1.0u"ms"/u"s")
0.001
```

c42f:
Actually it seems this was already fixed in #212, it's just not in the released version.

@ajkeller34 perhaps we could have a new release given all the handy fixes you've merged recently?

ajkeller34:
Yes, that's next up on my list. I should switch over from REQUIRE to Project.toml while I'm at it.

PR #212, opened 2019-03-05
Remove type annotation on last argument of ustrip methods

giordano:
<s>With the new safe `ustrip(unit, quantity)` method (which I like!) it's not possible to do something like `ustrip(NoUnits, 3.0m/1.0m)` (which, I think, should be safe enough).  I've added another method to make `ustrip(NoUnits, 3.0m/1.0cm)` fail, since it would be unsafe</s>

This enables using this function on dimensionless quantitites and missing objects too.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -25,8 +25,8 @@ julia> ustrip(Float64, u"m", 2u"mm") == 0.002
 true
 ```
 """
-@inline ustrip(u::Units, x::Quantity) = ustrip(uconvert(u, x))
-@inline ustrip(T::Type, u::Units, x::Quantity) = convert(T, ustrip(u, x))
+@inline ustrip(u::Units, x) = ustrip(uconvert(u, x))
+@inline ustrip(T::Type, u::Units, x) = convert(T, ustrip(u, x))
 
 """
     ustrip(x::Number)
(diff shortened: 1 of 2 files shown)


cstjean:
> I've added another method to make ustrip(NoUnits, 3.0m/1.0cm) fail, since it would be unsafe

Why? Isn't the answer 300 in all known universes?

giordano:
Yes but I don't know how to convince Unitful.jl that 3m/1cm is dimensionless

giordano:
> maybe by using uconvert?

Yes, this did the trick!

cstjean:
You're missing a method: 

```julia
julia> ustrip(Float64, unit(1.0), 1.0)
ERROR: MethodError: no method matching ustrip(::Type{Float64}, ::Unitful.FreeUnits{(),NoDims,nothing}, ::Float64)
```

src/utils.jl
```diff
@@ -27,6 +27,8 @@ true
 """
 @inline ustrip(u::Units, x::Quantity) = ustrip(uconvert(u, x))
 @inline ustrip(T::Type, u::Units, x::Quantity) = convert(T, ustrip(u, x))
```
cstjean: Wouldn't it be easier to just change `x::Quantity` for `x::Number` (or just `x`) in the two definitions above, rather than adding new definitions?

cstjean: If it was just `x`, it would have the extra benefit of correctly handling `missing` as well, which would be nice.

giordano:
I've simply removed the type annotation on the last argument of `ustrip` methods, as suggested by @cstjean, and added more tests.  Thank you so much!

ajkeller34:
Merged into [rigetti/Unitful.jl](https://github.com/rigetti/Unitful.jl/pull/1).

PR #213, opened 2019-03-05
Add show method for MIME"text/plain"

giordano:
Add a `show` method to print the value and unit parts in case they have custom representations with MIME `text/plain`


ajkeller34:
Merged into [rigetti/Unitful.jl](https://github.com/rigetti/Unitful.jl/pull/2).

PR #218, opened 2019-04-03
Defined 4 new isrootpower_dim methods

m-wells:
Added `isrootpower_dim` methods for length, time, temperature and mass.
```julia
isrootpower_dim(::typeof(dimension(m)))         = false
isrootpower_dim(::typeof(dimension(s)))         = false
isrootpower_dim(::typeof(dimension(K)))         = false
isrootpower_dim(::typeof(dimension(g)))         = false
```
These (I believe) are not controversial and more readily enable the creation of new log units.


m-wells:
Can someone explain how to fix the `codecov` checks?

Issue #38, opened 2016-10-18
dispatch on units of angle


mweastwood:
I am interested in adding Unitful as a dependency for a package I maintain, but it is important that I am able to dispatch on units of angle. For example I have a 3-vector type, and it would be nice to provide constructors like (without units all of the arguments are `Float64` and you don't get the convenience of having both constructors):

``` julia
f(x::Unitful.Length, y::Unitful.Length, z::Unitful.Length) = ...
f(r::Unitful.Length, theta::Unitful.Angle, phi::Unitful.Angle) = ...
```

I am also going to argue that "Angle" should be a fundamental (not derived dimension). The idea that angles are "dimensionless" seems to originate from the idea that it is a `length / length` (see table 3 http://physics.nist.gov/cuu/Units/units.html), but this is only correct in the small angle approximation. In general you need to use a trig function to get from an angle to something that is truly dimensionless. Or rather you need an inverse trig function to get an angle from something that is dimensionless.

Finally it would also be really cool if `sin(1°)` dispatched to `sind(1)`.
(see https://github.com/Keno/SIUnits.jl/blob/d25c43483828c954cd63c19cbe3ff209a614bbd1/src/nonsiunits.jl#L15-L22)

ajkeller34:
Wow, small world, someone else at Caltech found my package...

First, let me argue that no small angle approximation is required. Consider that the circumference of a circle is `(2π radians)*(radius)`. Surely the circumference has dimensions of length, not dimensions of length*angle. The dimension of `2π radians` must then be that of circumference over radius, or length/length, which is dimensionless. So, angle dimensions are kind of problematic for consistency when doing dimensional analysis, which is probably why there is no angle dimension in SI.

To do the dispatch you want, I offer two solutions:

1. You can specify `theta::Number, phi::Number` and then do a dimensions check in the function body (`dimension(theta) != NoDims && throw(Unitful.DimensionError())`). There will be no ambiguity with the other method you defined, and in the past I have found that the compiler can totally elide the dimensions check: even though it looks like a run-time check, you shouldn't see a branch in the generated code when you pass in a valid argument. This isn't something guaranteed by Unitful, rather it just appears to work that way based on some black magic going on inside Julia. I think it works because the dimension can be decided based on the type of theta, rather than its value. In other words, there should be no performance penalty for doing this.

[...]

PR #221, opened 2019-04-09
Implementation of an Angles Dimension

m-wells:
I've implemented an Angles Dimension.  This means that angles are no longer treated as dimensionless units.  I think there is a strong case to be made that if `Mole` is a dimensional unit then `Radian` is as well.

This opens up the ability to create derived dimensions for angular quantities like momentum, velocity, . . . and also torque.

Some of the features
```julia
julia> using Unitful

julia> u = Unitful
Unitful

julia> a = 0.5u.rad
0.5 rad

julia> b = uconvert(u.°,a)
28.64788975654116°

julia> sin(a) == sin(b)
true

julia> sind(a) == sind(b)
true

julia> sin(a) == sind(b)
true

julia> sin(a)
0.479425538604203

julia> sind(a)
0.479425538604203

julia> sin(b)
0.479425538604203

julia> sind(b)
0.479425538604203
```
The inverse functions (non-exported due to name conflicts) actually return the correct units.
```julia
julia> u.asin(sqrt(2)/2)
0.7853981633974484 rad

julia> u.asind(sqrt(2)/2)
45.00000000000001°
```
Dispatch on `Angle` quantities is consistent with other dimensions.
```julia
julia> f(x::u.Angle) = sin(x) + 2
f (generic function with 1 method)

julia> f(a)
2.479425538604203

julia> f(1)
ERROR: MethodError: no method matching f(::Int64)
Closest candidates are:
  f(::Union{Quantity{T,𝚽,U}, Level{L,S,Quantity{T,𝚽,U}} where S where L} where U where T) at REPL[26]:1
Stacktrace:
 [1] top-level scope at none:0
```
Currently radians drop only when they encounter a quantity with a length dimension via multiplication such that
```julia
julia> a
0.5 rad

julia> a*1u"m"
0.5 m

julia> a*1u"m"^2
0.5 m^2 rad

julia> a/1u"m"
0.5 rad m^-1
```

### DIFF
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -46,6 +46,7 @@ include("range.jl")
 include("fastmath.jl")
 include("logarithm.jl")
 include("pkgdefaults.jl")
+include("trigonometry.jl")
 
 function __init__()
     # @u_str should be aware of units defined in module Unitful
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -5,7 +5,8 @@ import Base: show, convert
 import Base: abs, abs2, angle, float, fma, muladd, inv, sqrt, cbrt
 import Base: min, max, floor, ceil, real, imag, conj
 import Base: exp, exp10, exp2, expm1, log, log10, log1p, log2
-import Base: sin, cos, tan, cot, sec, csc, atan, cis
+import Base: sin, cos, tan, csc, sec, cot, cis
+import Base: sind, cosd, tand, cscd, secd, cotd
 
 import Base: eps, mod, rem, div, fld, cld, trunc, round, sign, signbit
 import Base: isless, isapprox, isinteger, isreal, isinf, isfinite, isnan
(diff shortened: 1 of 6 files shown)


ajkeller34:
I agree it would be convenient to dispatch on Angle types but this isn't a logically consistent position, at least within the design constraints of this package. See previous discussion in https://github.com/PainterQubits/Unitful.jl/issues/38.

Regarding moles and the "number" dimension, I'm following the SI conventions.

bhalonen:
Comment on here: the use of `Number` would be an issue as it doesn't prevent someone from forgetting to put units on something they want to be a length, and having it dispatch silently to a function that processes it as an angle.

cmichelenstrofer:
@m-wells checkout [DimensionfulAngles.jl](https://github.com/cmichelenstrofer/DimensionfulAngles.jl).

PR #226, opened 2019-04-29
implement Base.complex, Base.widen, Base.reim

steinQF:
This PR helps dealing with complex, Unitful quantities by implementing missing methods.

Complex and yet Unitful quantities do occur e.g. as generalized frequencies or in the Ray Transfer matrix treatment of Gaussian beams.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 deps/
 docs/build/
 docs/site/
+*~
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -45,6 +45,7 @@ include("conversion.jl")
 include("range.jl")
 include("fastmath.jl")
 include("logarithm.jl")
+include("complex.jl")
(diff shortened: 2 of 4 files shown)


steinQF:
I'm surprised at the two test failures (appveyor x86 for Julia v0.7 and v1.0 but not for x64), especially since one of them failed in only 17 seconds and the another one timed out. Is it possible that these failures are unrelated to the code, perhaps due to some networking problem?

.gitignore
```diff
@@ -4,3 +4,4 @@
 deps/
 docs/build/
 docs/site/
+*~
```
giordano: Please *always* end a file with a newline

steinQF: Yes, sorry, I'll change that.

steinQF:
I'm abandoning this PR as testing revealed that the `Base.reim` method has a bug. I'll post a new PR when that is fixed.

PR #227, opened 2019-05-02
implement Base.complex, Base.widen, Base.reim (fix #223)

### LINKED ISSUE #223: Dealing with Real and Complex Quantities
It's tempting to think only real quantities matter, e.g. an imaginary meter would be strange and unphysical. In fact, generalized frequencies can be complex. Gaussian beam propagation using the [Ray Transfer Matrix](https://en.wikipedia.org/wiki/Ray_transfer_matrix_analysis) approach is another example where there are complex but unitful quantities involved.

I'm trying to generalize code for Gaussian beam propagation using Ray Transfer Matrices ([Issue 6 on ABCDBeamTrace.jl](https://github.com/ngedwin98/ABCDBeamTrace.jl/issues/6) in particular) written using dimensionless `Real` and `Complex` (with implied units) such that it works with Unitful.

Does anyone have suggestions for how to deal with the problem that Unitful.Quantity is merely a number and neither a subtype of `Real` or a `Complex`?

I think it would make sense to define at least a `Base.complex(realpart::Unitful.AbstractQuantity, imagpart::Unitful.AbstractQuantity=zero(realpart))` to ease defining complex, unitful quantities based on real valued ones. Ideally, I'd obviously wish for a Unitful.Quantity that somehow subtypes `Real` or `Complex` depending on the underlying type.

Would it make sense to define `AbstractQuantity` to be a subtype of `Real` and have `Complex{AbstractQuantity}` allowed? I'm not quite sure about that, since unitful quantities aren't strictly Real numbers, but it would lead to more code reuseability.

I realize there was an earlier discussion about subtyping, #197 and #9, and if I understand correctly, the verdict was that `Unitful.Quantity` should not subtype `Real` or `Complex`. I'm not sure I agree, but if that's the route to go, then how should one handle migrating code that tries to stay generic yet would benefit from being able to handle Unitful numbers?

steinQF:


src/complex.jl
```diff
@@ -0,0 +1,39 @@
+# This file is meant to provide all the methods in
+# https://github.com/JuliaLang/julia/blob/master/base/complex.jl that
+# are defined for Real or Complex numbers, just for
+# AbstractQuantities{T,D,U} where T is Real or Complex, respectively.
+#
+# It is currently incomplete.
+
+complex(z::Quantity{T,D,U}) where {T<:Complex,D,U} = z
+function complex(x::Quantity{T,D,U}, y = zero(x)) where {
```
ajkeller34: I suggest writing this instead as:

```
function complex(x::Quantity{T}, y = zero(x)) where {T<:Real}
    r, i = promote(x, y)
    return Quantity(complex(ustrip(r), ustrip(i)), unit(r))
end
```

The reason is because as you have written it, the following will happen:
```
julia> complex(1u"m",1.5u"m")
ERROR: InexactError: Int64(1.5)
```

If you can update your PR to reflect this change and also add some tests for the `complex(::Quantity)` functions you defined, I will merge it promptly. Thanks for the PR!

steinQF: Thanks for spotting this! I implemented the requested changes in a new commit.

ajkeller34:
Thanks, I appreciate the PR!

PR #239, opened 2019-06-03
Update Readme's badges and CI

briochemc:
The first commit updates Unitful's badges and links for its documentation to point to PainterQubits.

Also tried to update the badges for Travis, Appveyor, Codecov, and Coveralls, but I think those need to be set up on PainterQubits side. This probably means a few more changes to make these badges reflect the package's status properly.


briochemc:
Sorry I did not see this was one of many duplicates

Issue #207, opened 2019-02-04
round to units


rafaqz:
While single argument `round(1.5m)` and `trunc(3.4s)` etc can give unpredictable results, they wouldn't if we always specified the dimension with the type to round to using the two argument syntax :

```julia
using Unitful
using Unitful: m, cm
import Base: round
round(::Type{T}, x::Quantity) where {T<:Quantity} = T(round(typeof(zero(T).val), uconvert(unit(T), x).val))

julia> round(typeof(1m), 50.78cm)
1 m
```

Although there is probably a cleaner way to write that. 

I'm finding round() to be needed all the time for plotting series of environmental data. Being able to just round to the start of a day is very useful when using evenly spaced months or years where days need to be fractional.

Edit: I'll put this into a pull request if its useful.

c42f:
This seems sensible to me; it's basically the same thing as #205 / #48 but for a different function.

Related to plotting, did you see `UnitfulPlots.jl`?

rafaqz:
Cool. I've seen UnitfulPlots, most of the commits in the last year are mine :)

But it needs some love too. I've got another pull request sitting around for it, as it doesn't handle functions or DiffEq output. Function are an easy fix but debugging DiffEq output is awful.

rafaqz:
I have a pull request ready to go for this (and trunc etc) but there's already a backlog of my pull requests in the queue :/

ajkeller34:
Sorry, other work is unusually overwhelming right now; I will try to review PRs soon. Thanks for your patience.

ajkeller34:
A rounding PR came up again (https://github.com/PainterQubits/Unitful.jl/pull/246) and I wonder if this might be the time to revisit this issue. @rafaqz did you have a branch on your fork for this?

rafaqz:
PR is in.

ajkeller34:
Closed by #249, #246

Issue #78, opened 2017-04-23
Behavior of `floor`, `ceil` is surprising


ajkeller34:
`floor` and `ceil` pretend the units aren't there. Not sure why I did it this way, but I think this warrants reconsideration:

```
julia> 20u"µm"/(110u"nm")
0.18181818181818182 nm^-1 μm

julia> floor(ans)
0.0 nm^-1 µm

julia> Int(ans)
0
```

... which is different from the expected answer of 181. My inclination is to define `floor` and `ceil` only for dimensionless quantities and have them convert to a unitless number first.

giordano:
Also `isinteger` really depends on the actual units and can give unexpected result for non simplified quantities:
```julia
julia> isinteger(1u"m" / 10u"mm")
false

julia> isinteger(uconvert(NoUnits, 1u"m" / 10u"mm"))
true
```

ajkeller34:
`trunc` and `round` also had the same problems, but I think the latest patch should fix these issues.

PR #246, opened 2019-06-20
update rounding code

BeastyBlacksmith:
This updates the rounding code to the post 1.0 era.
To me rounding to significant digits is also well defined for Quantities.
I need to adjust the tests.


BeastyBlacksmith:
probably you want to keep the old methods for backwards compability?

ajkeller34:
Thanks, I agree a 1.0 update is useful, but I don't think rounding to significant digits is well defined for dimensionful quantities. It seems like things work if the units differ by powers of ten, but otherwise, the operation is not invariant under a change of units. Using your branch:

```
julia> l = 1u"cm"
1 cm

julia> floor(l; sigdigits=1) == floor(uconvert(u"km", l); sigdigits=1)
true

julia> floor(l; sigdigits=1) == floor(uconvert(u"inch", l); sigdigits=1)
false
```

For some context, this kind of operation was first prohibited after https://github.com/PainterQubits/Unitful.jl/issues/78. There was some discussion on making safe methods of `round` for dimensionful quantities in https://github.com/PainterQubits/Unitful.jl/issues/207.

c42f:
I think the two-argument form of #207 is reasonable, but I do wonder whether the inch-vs-km example is related to units as such. That is, given

```
julia> l = 1.23
julia> factor = 1.7
```

it's not surprising that the following versions using `Base.floor` are not equal:

```
julia> floor(l*factor, sigdigits=2)/factor
1.1764705882352942
julia> floor(l, sigdigits=2)
1.2
```

To the extent that these kind of functions are used for display of data to humans, I find this completely reasonable behavior.  On the other hand, surprises like #78 are no good.

BeastyBlacksmith:
I think the issue with
```
floor(l; sigdigits=1) == floor(uconvert(u"inch", l); sigdigits=1) # false
```
is rather promotion of rounded quantities than rounding to significant digits itself.
Personally I wouldn't suspect this equality to hold but to prevent issues it would be good to have a `RoundedQuantity` type and then define
```julia
promote_rule(x::RoundedQuantity,y::Any) = error("Rounded quantities cannot be promoted due to loss of precision")
```
or something like that.

ajkeller34:
@c42f I get what you're saying and I agree there are usability compromises that must be made eventually. But in your example, you are passing `round` different values. On the other hand, from a physical perspective, `1inch` and `2.54cm` are different representations of the same value. It's the same length, and I think you're taking the view that `round` should ignore that these are lengths and just consider them as real numbers, which might be practical but is not in the spirit of other design choices made in the package (for instance we don't have `Quantity{Float64} <: Real`). I think a closer analogy would be that `round(1.23f0) == round(1.23)`; representations of the same value in different types ought to return the same result.

c42f:
> On the other hand, from a physical perspective, 1inch and 2.54cm are different representations of the same value. It's the same length, and I think you're taking the view that round should ignore that these are lengths and just consider them as real numbers

To be clear, I'm not so much arguing for the changes in this PR as trying to understand the shape or the problem better by thinking out loud ;-) The point I was trying to make is that rounding is inherently dependent on the numerical representation because it's defined by finding the closest representable number in the base 10 positional numeral system with a given number of significant digits. Choose a different radix or numeral system and the set of representable numbers isn't even the same. The choice of representation in inch vs km gives a different lattice of numbers for the rounding operation, in some ways not dissimiiar from changing the radix.

In any case thanks for explaining the larger design considerations surrounding this. I think your analogy to `round(1.23f0) == round(1.23)` is quite accurate.

BeastyBlacksmith:
so I did rebase and added the two and three argument forms as discussed in https://github.com/PainterQubits/Unitful.jl/pull/249.
So give it a try and see if it works as expected.

PR #247, opened 2019-06-25
REQUIRE -> Project.toml; CI tweaks

c42f:
Here's some tweaks to help with the next release. I've kept the declared version at 0.15.0 for now so that will need to be updated as appropriate before using JuliaRegistrator.

* Convert REQUIRE to Project.toml using the gen_project.jl script
* Add version 1.2 to CI
* Modernize tweak script a bit


ajkeller34:
This is great, thank you!

PR #249, opened 2019-06-26
add two arg rounding methods

rafaqz:
Adds two-argument methods for round, trunc, ceil and floor,for rounding to specific units.

kwargs are also cleaned up for existing methods. I don't think we need to specify `digits=0` as that's the default.

### DIFF
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -276,11 +276,16 @@ _dimerr(f) = error("$f can only be well-defined for dimensionless ",
         "different results.")
 isinteger(x::AbstractQuantity) = _dimerr(isinteger)
 isinteger(x::DimensionlessQuantity) = isinteger(uconvert(NoUnits, x))
+
+_rounderr(f) = error("Specify the type of the quantity to convert to with $f. ",
+                     "Example: $f(typeof(1u\"m\"), 137u\"cm\") ")
+
 for f in (:floor, :ceil, :trunc, :round)
-    @eval ($f)(x::AbstractQuantity; digits=0) = _dimerr($f)
-    @eval ($f)(x::DimensionlessQuantity; digits=0) = ($f)(uconvert(NoUnits, x); digits=digits)
-    @eval ($f)(::Type{T}, x::AbstractQuantity) where {T <: Integer} = _dimerr($f)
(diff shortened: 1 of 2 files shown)


rafaqz:
I didn't realise there were new methods for round that we should support, this is just my old work rebased.

ajkeller34:
It may not have the new methods, but I think this PR is a better starting point for writing new methods, as it makes it clearer what operations are and are not considered safe / well-defined on `Quantity` types (well, at least in my view and I have yet to be persuaded otherwise). I am am in favor of merging this, but given ongoing rounding discussion maybe let's wait one day and see what happens. If we merge this, I'll ask @BeastyBlacksmith if he would be so kind as to rebase his PR on top of this one so that both PRs have a chance to be merged.

c42f:
No compliant from me; I think the two argument form is good.

Whether there's any way to define a sensible one argument form is a separate question.

BeastyBlacksmith:
I am also totally fine with a two argument form for rounding.
Two points I think that could be improved:

1. I would like to directly pass units like (optionally the unit defaults to the preferred unit)
```julia
round(u"mm", 1.5u"m")
```
2. It should work with `Measurements`, this errors with `ERROR: AssertionError: isinteger(a)`
```julia
round( typeof(1u"mm"), (1.5±0.2)u"m" )
```

rafaqz:
It rounds to the exact type you specify, so 1mm and 1.0mm are different. It's useful to be able to do that in one operation, but I think it would be fine to add another method that accepts Unit instead of Type{Quantity} if you want to.

It would also be good if you could get it to work with Measurements.

BeastyBlacksmith:
Ah, I see, so 
```julia
round( typeof((1.0±0.1)u"mm"), (1.5±0.2)u"m" )
```
works. So we could add
```julia
round(u::Units,q::AbstractQuantity) = round(Quantity{numtype(q),dimension(q),typeof(u)}, q)
```

c42f:
It might make sense to also have the three argument form analogous to `ustrip(Float64, u"m", 2u"mm")`.

rafaqz:
The reason I didn't do this is that `round` normally returns a value of the type you specify in the first argument. So `round(Int16, 4.5)` returns an `Int16`. Your version would still return a `Quantity` even though the first argument is of type `Units`. But that might not matter.

PR #251, opened 2019-07-02
Auto-register basefactors for precompiled unit extension modules

c42f:
The other day I got bitten by trying to use `@unit` in a custom extension module, but not reading the docs properly and getting confused when things didn't work.

Here's a change which removes the need for unit extension modules to know about the existence of `basefactors`.

### DIFF
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -30,7 +30,20 @@ export @logscale, @logunit, @dB, @B, @cNp, @Np
 export Level, Gain
 
 const unitmodules = Vector{Module}()
-const basefactors = Dict{Symbol,Tuple{Float64,Rational{Int}}}()
+
+function _basefactors(m::Module)
+    # A hidden symbol which will be automatically attached to any module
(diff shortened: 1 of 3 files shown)


ajkeller34:
This change looks great! I guess after merging I should make a PR to the general registry to upper-bound the Unitful version available to unit packages?

c42f:
Well I think this should actually be compatible even if unit extension modules continue to do it the old for a little while. Requires testing, but if all they do is to `merge!` in their local copy of this dict then what we'll get after this change will be multiple merges of the same data which presumably is harmless.

ajkeller34:
Seems to work as you expected. Thanks!

c42f:
Great!

I just read the docs section in more detail and realized that there's a distinction made between calling `Unitful.register` vs merging in `basefactors`, in regard to whether the new units are usable from `u_str` vs just being symbols in their originating package.

Personally I don't care about this distinction, but we could have an option to `register` which can turn off `u_str` support if desired.

Docs need updating too...

PR #255, opened 2019-07-19
Fix display() of angular degrees to be SI compliant

### LINKED ISSUE #253: Degree printing
Something has changed in the way degrees are printed that causes my doctests to fail. See e.g. https://travis-ci.org/bramtayl/LightQuery.jl/jobs/555926202 . I think the old way is better?

c42f:
As noticed in #253, `show` and `display` are inconsistent for degrees. This cleans things up to fix that (and could be simply extended if we ever supported arc minutes or arc seconds).

Closes #253

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -48,6 +48,13 @@ function show(io::IO, x::Unit{N,D}) where {N,D}
     show(io, FreeUnits{(x,), D, nothing}())
 end
 
+# Space between numerical value and unit should always be included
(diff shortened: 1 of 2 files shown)


giordano:
https://github.com/yakir12/UnitfulAngles.jl has arc minutes and arc seconds

c42f:
Cool I suppose it can overload `has_unit_spacing` then, since it displays them with ′ and ″.

PR #257, opened 2019-07-19
Doc for using local units in another local module

tianrluo:
It took me sometime figuring out how this could be done.
So Adding this updates to the doc might be helpful.


ajkeller34:
If I understand you correctly, it seems like what you were hung up on was how to bring an unregistered package into scope, rather than something specific to Unitful. I don't think you need to `include` a file as proposed in this PR, rather you can just use something like PkgTemplates.jl to start making a package as usual, and do `using MyUnits`. It doesn't have to be in the registry if you have done `pkg> dev MyUnits` to develop the package.

tianrluo:
I see, I'll close this.

PR #258, opened 2019-07-22
Add molar unit

jlumpe:
This is a really important one in chemistry and biology, I feel it should be included by default (especially since molarity is already included as a derived dimension).

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -118,6 +118,9 @@ for p in (:y, :z, :a, :f, :p, :n, :μ, :µ, :m, :c, :d,
     Core.eval(Unitful, :(const $(Symbol(p,:l)) = $(Symbol(p,:L))))
 end
 
+# Molarity
+@unit M      "M"        Molar       1mol/L                  true
+
 # Energy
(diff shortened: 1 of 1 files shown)


ajkeller34:
I agree and I'm happy to include this, but `Molar` has to be defined after `Liter` given the definition. I'll merge this once you make that change.

jlumpe:
Oops, fixed. I don't know if there are any tests I should incorporate it into, I didn't see any.

I'm also interested in adding a "count" unit with dimension 𝐍 which is equal to 1/6.022e23 of a mole, but I'm not really sure what its proper name or symbol should be. This would be very useful for converting from bulk (per mole) measurements to measurements per particle, like grams -> Daltons or in my case converting from concentration in a small volume like a cell to the actual number of molecules.

ajkeller34:
Thanks for the PR. I'm hesitant to include a unit that doesn't have an established name or symbol, though I can see the utility. Of course you can easily define units in your own packages.

erlebach:
I tried the following with Unitful installed: 

> julia> 1*u"ml" / 1*u"L"
> 1.0 mL L

I am surprised that you do not convert all units to SI and then the results would have been 10^{-3}. What is the rationale for not doing so? Thanks.

PR #260, opened 2019-08-10
`uconvert` shortcut for unitless `Number`.

tianrluo:
As `AbstractQuantity <: Number`, a shortcut for `uconvert` on unitless `Number`.

### DIFF
--- a/src/conversion.jl
+++ b/src/conversion.jl
@@ -75,6 +75,8 @@ function uconvert(a::Units, x::Quantity{T,D,U}) where {T,D,U}
     end
 end
 
+uconvert(::typeof(NoUnits), x::Union{Real,Complex}) = x
+
 function uconvert(a::Units, x::Number)
     if dimension(a) == NoDims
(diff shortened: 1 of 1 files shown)


ajkeller34:
Thank you for the PR. I'm not totally clear on what this shortcut was intended to do, but I suspect you are concerned about performance. The compiler is helpful and optimizes everything out even without your shortcut being defined. Below is without the PR:

```
julia> @code_llvm uconvert(NoUnits, 4.0)

;  @ /Users/andrew.keller/.julia/dev/Unitful/src/conversion.jl:79 within `uconvert'
define double @julia_uconvert_12497(double) {
top:
;  @ /Users/andrew.keller/.julia/dev/Unitful/src/conversion.jl:80 within `uconvert'
  ret double %0
}

julia> using BenchmarkTools; @btime uconvert(NoUnits, 4.0)
  0.033 ns (0 allocations: 0 bytes)
4.0
```

I'm going to close this but please let me know if I'm missing something!

PR #268, opened 2019-09-01
Fix conversions between `rpm`, `rps`, and other rotational units.

### LINKED ISSUE #261: conversion from RPM to rad/s is incorrect
Because there is no default `turn` or `rotation` unit defined, conversion from rpm or rps to rad/s gives incorrect results. As an example `1rps |> rad/s` should result in `2*pi rad/s` but the conversion pipeline results in `1//1 rad s^-1` which is incorrect. Changing the definition of RPM to `(2*pi)rad/minute` fixes the problem but requires rad be defined and may cause other problems I'm not familiar with.

ajkeller34:
Closes #261.

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -101,8 +101,8 @@ end
 @unit hr     "hr"       Hour                  3600s         false
 @unit d      "d"        Day                   86400s        false
 @unit wk     "wk"       Week                  604800s       false
-@unit rps    "rps"      RevolutionsPerSecond  1/s           false
-@unit rpm    "rpm"      RevolutionsPerMinute  1/minute      false
+@unit rps    "rps"      RevolutionsPerSecond  2π*rad/s      false
+@unit rpm    "rpm"      RevolutionsPerMinute  2π*rad/minute false
 
 # Area
 # The hectare is used more frequently than any other power-of-ten of an are.
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -71,8 +71,7 @@ for (_x,_y) in ((:sin,:sind), (:cos,:cosd), (:tan,:tand),
 end
 
 # SI and related units
-@unit cycle  "cycle"    Cycle       2π*rad                  false
-@unit Hz     "Hz"       Hertz       1cycle/s                true
+@unit Hz     "Hz"       Hertz       1/s                     true
 @unit N      "N"        Newton      1kg*m/s^2               true
 @unit Pa     "Pa"       Pascal      1N/m^2                  true
 @unit J      "J"        Joule       1N*m                    true
@@ -102,8 +101,8 @@ end
 @unit hr     "hr"       Hour                  3600s         false
 @unit d      "d"        Day                   86400s        false
 @unit wk     "wk"       Week                  604800s       false
-@unit rps    "rps"      RevolutionsPerSecond  1cycle/s      false
-@unit rpm    "rpm"      RevolutionsPerMinute  1cycle/minute false
+@unit rps    "rps"      RevolutionsPerSecond  2π*rad/s      false
+@unit rpm    "rpm"      RevolutionsPerMinute  2π*rad/minute false
 
 # Area
 # The hectare is used more frequently than any other power-of-ten of an are.
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -71,7 +71,8 @@ for (_x,_y) in ((:sin,:sind), (:cos,:cosd), (:tan,:tand),
 end
 
 # SI and related units
-@unit Hz     "Hz"       Hertz       1/s                     true
+@unit cycle  "cycle"    Cycle       2π*rad                  false
+@unit Hz     "Hz"       Hertz       1cycle/s                true
 @unit N      "N"        Newton      1kg*m/s^2               true
 @unit Pa     "Pa"       Pascal      1N/m^2                  true
 @unit J      "J"        Joule       1N*m                    true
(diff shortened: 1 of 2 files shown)


ajkeller34:
I'd like to keep this open for a couple days to see if there are any objections. In particular, I think the previous definition of `Hz` was also wrong. I'm surprised nobody complained about it before given how common of a unit `Hz` is, which is why I'm nervous to merge this.

src/pkgdefaults.jl
```diff
@@ -71,7 +71,8 @@ for (_x,_y) in ((:sin,:sind), (:cos,:cosd), (:tan,:tand),
 end
 
 # SI and related units
-@unit Hz     "Hz"       Hertz       1/s                     true
+@unit cycle  "cycle"    Cycle       2π*rad                  false
+@unit Hz     "Hz"       Hertz       1cycle/s                true
```
giordano: Wait, what? https://en.wikipedia.org/wiki/Hertz

ajkeller34: > The hertz (symbol: Hz) is the derived unit of frequency in the International System of Units (SI) and is defined as one cycle per second.[1] 

> ... a disc rotating at 60 revolutions per minute (rpm) is said to be rotating at either 2π rad/s or 1 Hz.

With this PR:

```
julia> 2π*u"rad/s" == 1u"Hz" == 60u"rpm"
true
```

on master:

```
julia> 2π*u"rad/s" == 1u"Hz" == 60u"rpm"
false
```

giordano:
```julia
julia> uconvert(u"s^-1", 1u"Hz")
6.283185307179586 s^-1
```
This is just wrong

giordano: On master:
```julia
julia> using Unitful, PhysicalConstants.CODATA2018

julia> uconvert(u"Hz / K", BoltzmannConstant / PlanckConstant)
2.0836619123327576e10 Hz K^-1
```
With this PR:
```julia
julia> using Unitful, PhysicalConstants.CODATA2018

julia> uconvert(u"Hz / K", BoltzmannConstant / PlanckConstant)
3.3162509308006988e9 Hz K^-1
```
Correct result: https://physics.nist.gov/cgi-bin/cuu/Value?kshhz

ajkeller34: I routinely consider `1Hz` as `1/s` myself, so don't get me wrong, I'm also a bit surprised. I'm just not sure how to avoid a contradiction unless things are defined this way. 

It is as if `1/s` cannot be read as "once per second," because without explicit angular units, radians are implicit (circumference in meters = 2π * length in meters). If we're explicit with `1cycle/s`, it is hard to disagree that a cycle is 2π radians, and so I think everything is correct. That's not to say this wouldn't break a lot of user code...

ajkeller34: I would very much welcome alternative suggestions to fix the reported issue without breaking other conversions.

ajkeller34: The root of the problem is that `2π == 2π * rad`; from units alone we cannot distinguish angular frequency from Hz.

giordano: I think there is a misunderstanding with the meaning of "cycle": https://en.wikipedia.org/wiki/Cycle_per_second.  Numerically it's just `1::Int`.  It doesn't measure the length of an arc.

giordano: > ... a disc rotating at 60 revolutions per minute (rpm) is said to be rotating at either 2π rad/s or 1 Hz.

Isn't 2π rad/s an angular velocity?  Why it should be equal to a frequency?

ajkeller34: I'll revert the commit for `Hz`, but something is still broken. We're now back to 

```
julia> 1u"Hz" == 2π * u"rad/s"
false
```

giordano: Which I think is right?  Isn't angular velocity 2pi times the frequency?
```julia
julia> uconvert(u"Hz", 2π*u"rad/s")
6.283185307179586 Hz
```

ajkeller34: Perhaps I'm not thinking about this clearly today given the US holiday weekend so feel free to mark this as resolved if you're sure it's right now, but I'm concerned that people will expect that the conversion to angular frequency will happen automatically as part of a unit conversion:

```
julia> uconvert(u"rad/s", 1u"Hz")
1//1 rad s^-1
```

Shouldn't this be 2pi times the frequency?

ajkeller34: > > ... a disc rotating at 60 revolutions per minute (rpm) is said to be rotating at either 2π rad/s or 1 Hz.
> 
> Isn't 2π rad/s an angular velocity? Why it should be equal to a frequency?

Dimensionally, angular velocity and frequency are the same, if you subscribe to the view that angles are dimensionless. This issue has cropped up before in Unitful discussions. I think it may be a limitation of the package design, in that there is no way to distinguish by type an angular velocity from a frequency.

giordano: Yes, exactly, angular velocity and frequency of a circular motion share the same physical dimensions but are scaled by a constant factor.

I don't find the result of `uconvert(u"rad/s", 1u"Hz")` surprising. I see 1 Hertz as one (dimensionless) "thing" per second. One radian happens to be a dimensionless quantity, so 1 radian per second is just that one "thing" per second.

I don't think it's this package job's to convert between different physical quantities that refer to the same motion and by chance share the same physical dimensions, this is more on the user side.

ajkeller34: > I don't think it's this package job's to convert between different physical quantities that refer to the same motion and by chance share the same physical dimensions, this is more on the user side.

Ok, I buy that. I'll mark this as resolved and squash+merge the PR. I think the confusion came from our differing presumptions about what `uconvert` should do, but your view seems more consistent. Thanks for the input.

gustaphe:
I know I'm a bit late to the party but I do object to this. I would expect `60rpm == 1Hz == 1/s`. To get an angular frequency (in radians/minute) corresponding to 1 *revolution* per minute, you take `ω = 2π*f`. This tripped me up bad recently.

PR #271, opened 2019-09-24
Add unit "angstrom"

singularitti:
I was going to contribute to [UnitfulAtomic.jl](https://github.com/sostock/UnitfulAtomic.jl), but the [author asks me to try here first](https://github.com/sostock/UnitfulAtomic.jl/pull/4).

I did not find any abbreviation of "angstrom" as "mi" does. But requiring input `Å` as a unit is harder than just typing `angstrom`. If you have better ideas, I would like to adopt.

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -182,7 +182,7 @@ const R∞ = 10_973_731.568_160/m     # (21) Rydberg constant
 @unit ft        "ft"       Foot                 12inch                  false
 @unit yd        "yd"       Yard                 3ft                     false
 @unit mi        "mi"       Mile                 1760yd                  false
-@unit angstrom  "Å"        Angstrom             (1//10)*Unitful.nm      false
+@unit angstrom  "Å"        Angstrom             (1//10)*nm      false
 
 # Area
 @unit ac        "ac"       Acre                 (316160658//78125)*m^2  false
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -182,6 +182,7 @@ const R∞ = 10_973_731.568_160/m     # (21) Rydberg constant
 @unit ft        "ft"       Foot                 12inch                  false
 @unit yd        "yd"       Yard                 3ft                     false
 @unit mi        "mi"       Mile                 1760yd                  false
+@unit angstrom  "Å"        Angstrom             (1//10)*Unitful.nm      false
 
 # Area
 @unit ac        "ac"       Acre                 (316160658//78125)*m^2  false


singularitti:
Wiki for angstrom: https://en.wikipedia.org/wiki/Angstrom

ajkeller34:
I would suggest adding a line `const Å = angstrom` so that people who want to type it are able to do so but otherwise I agree this is a well-known and common unit that should probably go here.

Was it necessary to write `Unitful.nm` in the macro?

giordano:
Note that Angstrom is already in `UnitfulAstro.jl`: https://github.com/JuliaAstro/UnitfulAstro.jl/blob/5598f8357944eeb6e5c1885adba444f3cc1f86a9/src/UnitfulAstro.jl#L55

singularitti:
> Note that Angstrom is already in `UnitfulAstro.jl`: https://github.com/JuliaAstro/UnitfulAstro.jl/blob/5598f8357944eeb6e5c1885adba444f3cc1f86a9/src/UnitfulAstro.jl#L55

That's a good point. I do not know why they need this unit since "angstrom" is so small to be astronomical. I do not know whether it would impact people who write
```julia
using UnitfulAstro: angstrom
```
if we merge this PR into Unitful.jl.

But it sounds weird to me that I need to install UnitfulAstro.jl in my field: dealing with atoms and molecules.

giordano:
>  I do not know why they need this unit since "angstrom" is so small to be astronomical

Wavelengths 🙂

> if we merge this PR into Unitful.jl and ask UnitfulAstro.jl maintainers to remove it.

I'm sure we'll find a way to not break users' code if angstrom is added here 😉

singularitti:
> I would suggest adding a line `const Å = angstrom` so that people who want to type it are able to do so but otherwise I agree this is a well-known and common unit that should probably go here.
> 
> Was it necessary to write `Unitful.nm` in the macro?

The reason I used `Unitful.nm` is that it will keep the fractional-structure when converting
```julia
julia> @unit angstrom "Å" Angstrom (1 // 10) * Unitful.nm false

julia> Unitful.register(Main)

julia> uconvert(u"m", 1u"angstrom").val
1//10000000000
```
However, if I use UnitfulAstro.jl, it will be converted to floats:
```julia
julia> using UnitfulAstro

julia> uconvert(u"m", 1u"angstrom").val
1.0e-10
```
I do not know whether this is a good choice. But I saw your [prefer fractional-structures](https://github.com/PainterQubits/Unitful.jl/blob/9dfc702c5a2ea128afda453504abef5923e86d61/src/pkgdefaults.jl#L180-L181):
```julia
@unit inch      "inch"     Inch                 (254//10000)*m          false
@unit mil       "mil"      Mil                  (1//1000)*inch          false
```

src/pkgdefaults.jl
```diff
@@ -182,6 +182,7 @@ const R∞ = 10_973_731.568_160/m     # (21) Rydberg constant
 @unit ft        "ft"       Foot                 12inch                  false
 @unit yd        "yd"       Yard                 3ft                     false
 @unit mi        "mi"       Mile                 1760yd                  false
+@unit angstrom  "Å"        Angstrom             (1//10)*Unitful.nm      false
```
ajkeller34: `const Å = angstrom` should go immediately following the `@unit` definition.

ajkeller34:
> The reason I used Unitful.nm is that it will keep the fractional-structure...

My question was really about whether the `Unitful.` part of `Unitful.nm` was necessary. Why not just write `nm`, consistent with the style of the rest of the file? 

> Note that Angstrom is already in `UnitfulAstro.jl`

At least the `u_str` macro shouldn't fail silently if you have two different modules defining the same unit—I think the unit from the most recently registered units module is used, and importantly a warning should be emitted.

singularitti:
> > The reason I used Unitful.nm is that it will keep the fractional-structure...
> 
> My question was really about whether the `Unitful.` part of `Unitful.nm` was necessary. Why not just write `nm`, consistent with the style of the rest of the file?
> 
> > Note that Angstrom is already in `UnitfulAstro.jl`
> 
> At least the `u_str` macro shouldn't fail silently if you have two different modules defining the same unit—I think the unit from the most recently registered units module is used, and importantly a warning should be emitted.

Oh, I was just not aware of that. I have updated as you suggested.

singularitti:
Any updated information?

PR #278, opened 2019-10-15
Handle missing |> kg

cstjean:
Before this PR, `vec .|> kg` would fail if it contained any `missing`


cstjean:
Failure is on `nightly` and unrelated...

PR #280, opened 2019-10-29
Add constructorof for Quantity, and ConstructionBase dep

rafaqz:
This is just `constructorof` for `Quantify` at this stage, as it is definitely used already and is useful. I'm not sure `constructorof` would ever be used for other types that have fields, like `Level` of `Gain`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,9 @@ julia:
   - nightly
 notifications:
   email: false
-# matrix:
-#   allow_failures:
-#   - julia: nightly
+matrix:
+  allow_failures:
+  - julia: nightly
 # uncomment the following lines to override the default test script
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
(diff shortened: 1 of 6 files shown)


ajkeller34:
Would you mind putting an upper bound on the version of `ConstructorBase` in Project.toml? The next time I tag a version that will likely be required.

rafaqz:
Agreed, I'm just getting used to putting upper bounds on everything. We might bump ConstructionBase to 1.0 before that as well.

Unfortunately the ConstructionBase dependency breaks 0.7 compatibility. It that's ok I'll remove it from CI and Project.toml. The nightly build is also failing, but that seems to be from an unrelated error.

ajkeller34:
Honestly, it's probably about time to drop julia 0.7 support, and maybe also soon time to tag Unitful 1.0. I have been thinking that this package is in pretty wide use and I'd like the versioning to reflect some degree of maturity. There are some things in Unitful that would be nice to improve before 1.0, but hey, that's what 2.0 is for.

rafaqz:
Ok this is done and passing. Nightly failures are due to `@fastmath` errors.

Should we allow failures on nightly ?

ajkeller34:
Allowing failures on nightly seems like a good idea, but I should really fix whatever is going on as well. Hope to do that this week.

Issue #225, opened 2019-04-20
Bug: Array Division by Unit


bhgomes:
```julia
using Unitful
ones(3) / u"s"

ERROR: MethodError: no method matching /(::Array{Float64,1}, ::Unitful.FreeUnits{(s,),𝐓,nothing})
```

It appears that division by unit was not implemented here, just multiplication.

https://github.com/ajkeller34/Unitful.jl/blob/82f84969c2b78fcc3f66bc23ef997e81c0f0661e/src/quantities.jl#L35-L36

sostock:
This was fixed in #369.

PR #284, opened 2019-11-11
Resolve issue #225.

ajkeller34:
At some point, should consider using `broadcast_preserving_zero_d` instead of `broadcast` (ref: https://github.com/JuliaLang/julia/pull/32122)


sostock:
This was implemented in #369 instead.

Issue #125, opened 2018-03-14
Support automatic conversions from `TimeType`s


rofinn:
It would be nice to support automatic handling of julia base timetypes.

Example)
```
a = now()
b = a - Dates.Day(1)

hrs = (a - b)u"minute"
```

cstjean:
@ajkeller34 We need something like this too. I'm happy to make a PR, but a decision needs to be made about what form it will take. Is this desirable at all? If so, should Unitful import `Dates`, or use `Require`?

ajkeller34:
I intended to get around to this eventually, but who knows when. A PR would be very helpful. Importing `Dates` seems reasonable. Thanks!

PR #288, opened 2019-12-10
Add unit of year

briochemc:
I think that expressing quantities in units of years is so ubiquitous in science that it would be nice to have it in the default Unitful.

FWIW I used 365.25 days = 31557600 seconds for the conversion like in UnitfulAstro.jl.

There may have been previous discussions about this but I could not find them. (Maybe they disappeared when Unitful moved to PainterQubits?)

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -104,6 +104,7 @@ end
 @unit hr     "hr"       Hour                  3600s         false
 @unit d      "d"        Day                   86400s        false
 @unit wk     "wk"       Week                  604800s       false
+@unit yr     "yr"       Year                  31557600s     false
 @unit rps    "rps"      RevolutionsPerSecond  2π*rad/s      false
 @unit rpm    "rpm"      RevolutionsPerMinute  2π*rad/minute false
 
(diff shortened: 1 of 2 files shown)


briochemc:
I realize now that one might want to change the example exteding Unitful because it extends it with "yr".

cstjean:
I also defined year to be 365.25 days, but wouldn't it make sense to keep it application-specific, since different domains might have different requirements?

Furthermore, if https://github.com/PainterQubits/Unitful.jl/issues/125 is still planned, we would have `1year != Year(1)`. 

Those are minor points though, I wouldn't mind.

briochemc:
> wouldn't it make sense to keep it application-specific, since different domains might have different requirements?

My opinion is that it is still better for (most?) Unitful users to have a default `u"yr"` available. If a specific application really needs something different then I believe it should precisely care about making its own precise definition for it, no?

ajkeller34:
I think I agree with these changes. For calculations with physical quantities it seems like the most generically useful definition is the one in this PR. Other definitions could shadow if needed. Will close and reopen to retrigger CI.

PR #295, opened 2019-12-30
Add convenience tools for converting from/to magnitudes wrt. preferred units

j-fu:
Hi,

for a pde simulation toolbox potentially using several tools not able to work  with unitful (a number of sparse matrix tools etc. - I don't see that this can be fixed in a short timeframe) I would like to see some convenience tools to support working with magnitudes only. 

ustrip is of course doing the job, but forces the user to ensure that she is consistent  as we have ustrip(1m)+ustrip(1cm)==1 ...

So my proposal is to make a shortcut of ustrip(upreferred) named quantity2magnitude 

As a consequence, we need to ensure that preferunits does what the user intended, thus assert_preferunits (IMHO it would be sufficient that preferunits returns true if it was successful) and
once we are at that, assert_prefersibase.

For the other way round we need a consistent shortcut to uconvert(upreferred), this could be magnitude2quantity. 

With this setup the user can define multipliers wrt. preferred units and work consistently with them, and go back to quantities if necessary. 

... I know, this cripples a bit the idea of the package, OTOH I really believe Unitful _is great_ and should be the go-to database for working with units. We just need some more tooling around, e.g. to handle situations like the described one. 

Jürgen

### DIFF
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -31,6 +31,9 @@ export uconvertp, uconvertrp, convertr, convertrp, reflevel, linear
 export @logscale, @logunit, @dB, @B, @cNp, @Np
 export Level, Gain
 
+export quantity2magnitude, magnitude2quantity
+export assert_preferunits, assert_prefersibase
+
 const unitmodules = Vector{Module}()
 
 function _basefactors(m::Module)
--- a/src/user.jl
+++ b/src/user.jl
@@ -586,3 +586,76 @@ function basefactor end
 Returns a [`Unitful.Dimensions`](@ref) object describing the given unit `x`.
 """
 function dimension end
+
+
+
+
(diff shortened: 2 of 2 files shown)


ajkeller34:
Thanks for your PR. A few comments here. First, `@assert` is not intended to be used that way, I think it can be optimized out for some optimization levels, which presumably you don't want. 

Second, perhaps I don't understand your use case, but I think you could define functions in your own code based on the two-arg version `ustrip(u"m", 1.0u"cm")` for situations where you want the magnitude in a given unit.

Finally, I do agree that it is a real challenge to get Unitful quantities working well with all linear algebra / sparse matrix functionality, and so converting to Float64 is sometimes needed... I just think the existing functions are probably enough to handle your use cases.

PR #296, opened 2020-01-06
Support Float32/Float16 etc. type conversions

cstjean:
Eg. `Float32(1u"m") === 1.0f0u"m")`


cstjean:
Also, since Base has `float(x) = AbstractFloat(x)`, it would be slightly cleaner to implement `AbstractFloat` instead of `float`

Issue #18, opened 2016-09-16
Better display for units


ajkeller34:
[SIUnits.jl](https://github.com/Keno/SIUnits.jl) seemed to handle this pretty well. There's no superscript slash in Unicode, so printing rational exponents using Unicode only will be problematic.

mlhetland:
I see this is closed, but I thought I wouldn't open a new issue just for this: Might it be a good idea to differentiate between compact and non-compact display? Primarily removing the space in compact display, I guess, so `show(1cm)` gives you `1 cm` but `showcompact` yields `1cm`? (Similar to operator spacing for complex numbers, for example.)

PR #297, opened 2020-01-16
unicode exponents for display

briochemc:
Suggestion for https://github.com/PainterQubits/Unitful.jl/issues/18 to use `ᐟ` for displaying fractional dimensions.

FYI, I used SIUnits' `superscript` combined with unicode character `ᐟ` (Canadian Syllabics Final Acute `'\u141F'`), which looks like a superscript `/` for fractions.

It looks like
```julia
julia> 3u"m^4/s^(1//2)"
3 m⁴ s⁻¹ᐟ²
```
and I think/hope this is good enough?

### DIFF
--- a/docs/src/display.md
+++ b/docs/src/display.md
@@ -1,3 +1,8 @@
+By default, exponents on units or dimensions are indicated using Unicode superscripts on
+macOS and without superscripts on other operating systems. You can set the environment
+variable `UNITFUL_FANCY_EXPONENTS` to either `true` or `false` to force using or not using
+the exponents.
+
 ```@docs
 Unitful.abbr
 Unitful.prefix
--- a/src/display.jl
+++ b/src/display.jl
@@ -154,10 +154,6 @@ end
     superscript(i::Rational)
 Prints exponents.
 """
-"""
-    superscript(i::Rational)
-Prints exponents.
-"""
 function superscript(i::Rational)
     i.den == 1 ? superscript(i.num) : string(superscript(i.num), '\u141F', superscript(i.den))
 end
--- a/src/display.jl
+++ b/src/display.jl
@@ -155,8 +155,16 @@ end
 Prints exponents.
 """
 function superscript(i::Rational)
-    i.den == 1 ? superscript(i.num) : string(superscript(i.num), '\u141F', superscript(i.den))
+    v = @eval get(ENV, "UNITFUL_FANCY_EXPONENTS", $(Sys.isapple() ? "true" : "false"))
+    t = tryparse(Bool, lowercase(v))
+    k = (t === nothing) ? false : t
+    if k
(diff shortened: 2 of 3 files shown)


ajkeller34:
Great find, it looks perfect on my Mac. Unfortunately on Windows, the default console does not seem to display the character. If I launch Julia from the console that Git Bash runs in, the character displays and the fraction is readable, but it doesn't look as nice as one would hope.

This is so slick I'm tempted to merge it anyway, but my better judgment says that it is worth thinking about this a bit. I wonder if the best thing to do for now is display the old way on Windows?

briochemc:
Ah yes I'm sorry I tested only on my mac (and by "tested" I mean just tried it out, as you can see the failed CI tests 😅). 

I found the character in [this SO question](https://stackoverflow.com/questions/16239848/how-to-represent-a-superscript-slash-with-unicode/49325148), which suggests two other characters. First is **Right Raised Omission Bracket** `⸍` `'\u2E0D'`, which does not look as nice on my mac. But maybe it looks good on Windows and that would work? The other character suggested, **Musical Symbol Repeated Figure-1**, does not display on my mac at all.

ajkeller34:
It's a similar situation for the other suggested character on Windows. Anyone able to chime in regarding appearance on Linux systems?

briochemc:
What about having an environment variable or some other flag mechanism to toggle either of the display options? 

My guess is that the actual display of unicode characters is not exactly an OS issue but rather a question of which fonts are installed on that OS, right? Having a toggle/flag for swithcing would be good enough IMHO :) (Even better would be to detect if the characters are displayed somehow)

ajkeller34:
I am not sure how much of it is only a font issue. The default Windows console may be part of the problem.

An environment variable would work. Kind of wordy, but maybe `UNITFUL_UNICODE_DISPLAY`? It should probably default to `true` for macOS and `false` for Windows, still not sure about Linux or FreeBSD for that matter.

briochemc:
I asked on slack (mentioned you but not sure you saw) and Mark Kittisopikul uploaded this screenshot: 

image

giordano:
This is how it looks like on my system (Arch Linux):
image
Not great

briochemc:
@giordano Could you check the other characters?

ajkeller34:
I think it makes sense to default to this on macOS and otherwise have it be opt-in using an environment variable. It's legible on all of the default themes in Terminal (2/10 themes look a little funny but still pretty decent), and seems legible with default iTerm2 settings as well. There's probably a fair bit of variation across other operating systems.

ajkeller34:
I rebased onto master and did some force pushing, sorry for the noise. The environment variable `UNITFUL_FANCY_EXPONENTS` is annoying to type but presumably people won't be changing it all the time. Since nobody objected to the proposed behavior, I'll go ahead and merge this.

PR #298, opened 2020-01-17
Parsing of units from strings

### LINKED ISSUE #214: Add function for run-time parsing
The `@u_str` macro is useful when units are known in advance. In some cases, however, units must be determined at runtime. This requires parsing strings as numeric and Unitful values, which is currently sort of ugly:

```julia
val, unit = split(rawValue, " "; limit=2)
return parse(Float64, val)*(@eval @u_str $unit) # ugly
```
I'd like to propose that Unitful include a function which exposes this kind of behavior (i.e. parse quantity from string). It would be good if it were resilient to having a space (or not) between the numeric and the unit.

### LINKED ISSUE #272: @u_str macro hygiene?
I'm not sure if this is intentional or not, but I expected it to be sufficient to just `import Unitful: @u_str` in order to use e.g. u"m".

```julia
julia> import Unitful: @u_str

julia> u"m"
ERROR: LoadError: Symbol `m` was found in unit module Unitful, but was not loaded into Main. Consider `using Unitful` within `Main`?
Stacktrace:
 [2] replace_value(::Module, ::Symbol) at /Users/goretkin/.julia/packages/Unitful/ytsW0/src/user.jl:550
 [3] @u_str(::LineNumberNode, ::Module, ::Any) at /Users/goretkin/.julia/packages/Unitful/ytsW0/src/user.jl:505
in expression starting at REPL[4]:1

julia> import Unitful: m

julia> u"m"
ERROR: LoadError: Symbol `m` was found in unit module Unitful, but was not loaded into Main. Consider `using Unitful` within `Main`?
Stacktrace:
 [2] replace_value(::Module, ::Symbol) at /Users/goretkin/.julia/packages/Unitful/ytsW0/src/user.jl:550
 [3] @u_str(::LineNumberNode, ::Module, ::Any) at /Users/goretkin/.julia/packages/Unitful/ytsW0/src/user.jl:505
in expression starting at REPL[6]:1

julia> using Unitful

julia> u"m"
m

julia> m = "garbage"
ERROR: cannot assign a value to variable Unitful.m from module Main
Stacktrace:
 [1] top-level scope at REPL[9]:1
```

c42f:
Generalize the handling of `@u_str` and the unit extension module
context to allow the same machinery to be used for a new `parseunit`
function which does the same thing at runtime.

Fixes #214 in a basic kind of way, and fixes #272 as a side effect of generalizing `@u_str`.

I'm fairly happy with the implementation here, but I wonder whether there's a better naming for `parseunit`. For example, we could possibly do something like overload `Base.parse(::Unit, str)` instead if that's appropriate. Arguably this could extend more naturally to `Base.parse(::Quantity, str)`). Thoughts?

### DIFF
--- a/src/user.jl
+++ b/src/user.jl
@@ -514,26 +514,26 @@ macro u_str(unit)
 end
 
 """
-    parseunit([unit_module(s),] string)
+    uparse(string, [unit_context=ctx,])
 
-Parse a string as a unit type. The format for `string` must be a valid Julia
-expression, and any identifiers will be looked up in the context `unit_module`
+Parse a string as a unit or quantity. The format for `string` must be a valid
+Julia expression, and any identifiers will be looked up in the context `ctx`,
 which may be a `Module` or a vector of `Module`s. By default
-`unit_module=Unitful`.
+`unit_context=Unitful`.
 
 Examples:
 
 ```jldoctest
-julia> parseunit("m/s")
-1.0 m s^-1
+julia> uparse("m/s")
+m s^-1
+
+julia> uparse("1.0*dB")
+1.0 dB
 ```
 """
-parseunit(str::AbstractString) = parseunit(Unitful, str)
-
-parseunit(unitmodule::Module, str) = parseunit([unitmodule], str)
-function parseunit(unitmods, str)
+function uparse(str, unit_context=Unitful)
     ex = Meta.parse(str)
-    eval(lookup_units(unitmods, ex))
+    eval(lookup_units(unit_context, ex))
 end
 
 const allowed_funcs = [:*, :/, :^, :sqrt, :√, :+, :-, ://]
@@ -590,6 +590,7 @@ function lookup_units(unitmods, sym::Symbol)
                  We will use the one from $m."""
     return u
 end
+lookup_units(unitmod::Module, ex::Symbol) = lookup_units([unitmod], ex)
 
 lookup_units(unitmods, literal::Number) = literal
 
--- a/src/user.jl
+++ b/src/user.jl
@@ -514,7 +514,7 @@ macro u_str(unit)
 end
 
 """
-    uparse(string, [unit_context=ctx,])
+    uparse(string [; unit_context=ctx])
 
 Parse a string as a unit or quantity. The format for `string` must be a valid
 Julia expression, and any identifiers will be looked up in the context `ctx`,
@@ -531,7 +531,7 @@ julia> uparse("1.0*dB")
 1.0 dB
 ```
 """
-function uparse(str, unit_context=Unitful)
+function uparse(str; unit_context=Unitful)
     ex = Meta.parse(str)
     eval(lookup_units(unit_context, ex))
 end
--- a/src/user.jl
+++ b/src/user.jl
@@ -502,67 +502,96 @@ julia> u"ħ"
 """
 macro u_str(unit)
     ex = Meta.parse(unit)
-    esc(replace_value(__module__, ex))
+    unitmods = [Unitful]
+    for m in Unitful.unitmodules
+        # Find registered unit extension modules which are also loaded by
+        # __module__ (required so that precompilation will work).
+        if isdefined(__module__, nameof(m)) && getfield(__module__, nameof(m)) === m
+            push!(unitmods, m)
+        end
+    end
+    esc(lookup_units(unitmods, ex))
+end
+
+"""
+    parseunit([unit_module(s),] string)
+
+Parse a string as a unit type. The format for `string` must be a valid Julia
+expression, and any identifiers will be looked up in the context `unit_module`
+which may be a `Module` or a vector of `Module`s. By default
+`unit_module=Unitful`.
+
+Examples:
+
+```jldoctest
+julia> parseunit("m/s")
+1.0 m s^-1
+```
+"""
+parseunit(str::AbstractString) = parseunit(Unitful, str)
+
+parseunit(unitmodule::Module, str) = parseunit([unitmodule], str)
+function parseunit(unitmods, str)
+    ex = Meta.parse(str)
+    eval(lookup_units(unitmods, ex))
 end
 
 const allowed_funcs = [:*, :/, :^, :sqrt, :√, :+, :-, ://]
-function replace_value(targetmod, ex::Expr)
+function lookup_units(unitmods, ex::Expr)
     if ex.head == :call
         ex.args[1] in allowed_funcs ||
-            error("""$(ex.args[1]) is not a valid function call when parsing a unit.
-             Only the following functions are allowed: $allowed_funcs""")
+            throw(ArgumentError(
+                  """$(ex.args[1]) is not a valid function call when parsing a unit.
+                   Only the following functions are allowed: $allowed_funcs"""))
         for i=2:length(ex.args)
             if typeof(ex.args[i])==Symbol || typeof(ex.args[i])==Expr
-                ex.args[i]=replace_value(targetmod, ex.args[i])
+                ex.args[i]=lookup_units(unitmods, ex.args[i])
             end
         end
         return ex
     elseif ex.head == :tuple
         for i=1:length(ex.args)
             if typeof(ex.args[i])==Symbol
-                ex.args[i]=replace_value(targetmod, ex.args[i])
+                ex.args[i]=lookup_units(unitmods, ex.args[i])
             else
-                error("only use symbols inside the tuple.")
+                throw(ArgumentError("Only use symbols inside the tuple."))
             end
         end
         return ex
     else
-        error("Expr head $(ex.head) must equal :call or :tuple")
+        throw(ArgumentError("Expr head $(ex.head) must equal :call or :tuple"))
     end
 end
 
-function replace_value(targetmod, sym::Symbol)
-    f = m->()
-    inds = findall(unitmodules) do m
-        # Ensure that both the unit exists in the registered unit module, and
-        # that the target module (the one invoking `u_str`) has loaded it so
-        # that precompilation will work.
-        isdefined(m,sym)             && ustrcheck_bool(getfield(m, sym))  &&
-        isdefined(targetmod, nameof(m)) && getfield(targetmod,nameof(m)) === m
-    end
+function lookup_units(unitmods, sym::Symbol)
+    has_unit = m->(isdefined(m,sym) && ustrcheck_bool(getfield(m, sym)))
+    inds = findall(has_unit, unitmods)
     if isempty(inds)
         # Check whether unit exists in the global list to give an improved
         # error message.
-        f = m->(isdefined(m,sym) && ustrcheck_bool(getfield(m, sym)))
-        hintidx = findfirst(f, unitmodules)
+        hintidx = findfirst(has_unit, unitmodules)
         if hintidx !== nothing
             hintmod = unitmodules[hintidx]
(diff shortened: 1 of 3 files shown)


src/user.jl
```diff
@@ -502,67 +502,96 @@ julia> u"ħ"
...
+julia> parseunit("m/s")
+1.0 m s^-1
+```
+"""
+parseunit(str::AbstractString) = parseunit(Unitful, str)
+
+parseunit(unitmodule::Module, str) = parseunit([unitmodule], str)
+function parseunit(unitmods, str)
+    ex = Meta.parse(str)
+    eval(lookup_units(unitmods, ex))
```
c42f: By the way, I believe this use of `eval` is a rare circumstance where it's probably warranted: we want to evaluate the expression containing the small number of `allowed_funcs` where we've already sanitized the expression `ex` and fairly carefully resolved all symbols ourselves. In addition, we want to use `Unitful.eval` in particular, taking the `allowed_funcs` bindings from the scope of `Unitful`.

test/runtests.jl
```diff
@@ -387,22 +387,27 @@ end
...
-    @test_throws LoadError @eval u"basefactor"
-
-    # test ustrcheck(::Quantity)
-    @test u"h" == Unitful.h
-    @test u"π" == π              # issue 112
+@testset "Unit string parsing" begin
+    @test parseunit("m") == m
+    @test parseunit("m,s") == (m,s)
+    @test parseunit("1.0") == 1.0
+    @test parseunit("m/s") == m/s
```
Datseris: I guess a test about `ms` should exist as well, to see whether it is interpreted as miliseconds?

How would you type the units of meters times seconds? In general in the tests presented here I don't see any test regarding parsing a multiplication of units (which is a reasonable way to make units as well).

src/user.jl
```diff
@@ -502,67 +502,96 @@ julia> u"ħ"
...
+        end
+    end
+    esc(lookup_units(unitmods, ex))
+end
+
+"""
+    parseunit([unit_module(s),] string)
+
+Parse a string as a unit type. The format for `string` must be a valid Julia
+expression, and any identifiers will be looked up in the context `unit_module`
```
Datseris: I propose some doc improvement here: after `valid Julia expression` add "(i.e. multiplication between units is specified with `*`, division between units with `/` and the string has no space characters)". After this parenthesis add a full stop, to make the sentence easier to read, and continue with "Any identifiers...".

test/runtests.jl
Datseris: EDIT: I guess by using multiplication sign `*`, but still this could/should be one test case as well.

Datseris:
As a user or a developer of another Julia package that wants to add support for Unitful (namely NCDatasets.jl), how would I extend this function to handle cases more used in my field, e.g. using spaces to separate units that are multiplied?

test/runtests.jl
c42f: Yeah, we could add that. As you can see, I just converted the existing `u_str` tests (which use the same infrastructure).

c42f:
> how would I extend this function to handle cases more used in my field

So you are using NetCDF, yes? Do you know exactly what the rules are for parsing valid NetCDF unit strings?  My impression is that they're defined by the implementation of the udunits package, which supposedly implements the formal grammar available at https://www.unidata.ucar.edu/software/udunits/udunits-current/doc/udunits/udunits2lib.html#Grammar.

So the robust and complete way to do it would be to write a proper parser which follows that exact grammar. But this seems potentially NetCDF specific so it's not clear that it belongs in `Unitful` itself.

For some cases you might get away with simple string substitutions to turn your NetCDF unit strings into julia syntax, which could then be passed to `parseunit` with an appropriate unit module. Though I admit this is not very satisfying and could lead to confusing error messages for invalid unit strings.

Datseris:
I've commented this in the issue page, but here is an idea: make `parseunit` defined as :

```
parseunit([modules], str, preprocess = identity)
```
This would allow users and package developers to give a pre-processing function in a natural way, without Unitful.jl ever having to worry about what whackadoodle functions the users would use?

And for some pre-processing functions, if you find them helpful enough, you could add them in the docs of Unitful.jl?

c42f:
But I'm not sure how this is better than simply writing `parseunit(modules, preprocess(str))` which is more orthogonal and compositional?

Datseris:
I guess the first way explicitly shows how you would extend the parsing functionality, while the second has to be explicitly mentioned in the docs? But sure, we can move forward with your way.

Datseris:
orthogonal to the current discussion: can the first argument of `parseunit` be also made a `Vector{<:Unit}` ?

ajkeller34:
Thanks so much for this long-desired functionality!

> I'm fairly happy with the implementation here, but I wonder whether there's a better naming for parseunit. For example, we could possibly do something like overload `Base.parse(::Unit, str)` instead if that's appropriate. Arguably this could extend more naturally to `Base.parse(::Quantity, str)`. Thoughts?

The new implementation is really nice. It is a little funny that `parseunit` can return a `Units` or a `Quantity`, but that's not a problem to me. I suppose something along the lines of what you're describing would be a natural extension of `Base.parse`, and make parsing `Quantity`s feel a bit more first-class.

It seems like `Base.parse(type, str)` expects a concrete type as the first argument, for example `parse(Complex, "1.0 + 2.0im")` does not work. That raises two questions:
  1. Would we also require a concrete type, or would `parse(Quantity, "1.0m")` be expected to work?
  2. Should e.g. `parse(typeof(u"1.0cm"), "2.0km")` work?

I suppose since `parse(Complex{Float64}, "1+2im")` evidently works and converts to `Complex{Float64}`, probably the answer to the second question is that it should work and result in a `Quantity` in units of `cm`.  The answer to the first question is not as obvious to me, since we would want to maintain a similar level of convenience to `parseunit(str)` but maybe there were clearly articulated reasons why `parse(Complex, str)` was not implemented. I was not able to find any relevant discussion. I would lean towards implementing `parse(Quantity, str)` and `parse(Units, str)`.

Finally, would you imagine implementing three-arg versions, `Base.parse(::Module, Quantity, str)` or `Base.parse(::Module, Units, str)`, similar to the two-arg version of `parseunit`?

c42f:
> `parseunit` can return a `Units` or a `Quantity`

Oh right, somehow I missed this! I guess an alternative spelling could be `uparse` in somewhat-analogy to `ustrip` and `uconvert`; it can be a "unit aware" parsing function.

> maybe there were clearly articulated reasons why `parse(Complex, str)` was not implemented

I'm not sure about the reasoning though I think it's unclear what the concrete return type should be for cases like `parse(Complex, "1")`. Either it defaults to a fixed type like `ComplexF64` or it's inferred using some kind of rules. If it's inferred then `parse` becomes inherently type unstable (as is the case for `parseunit`).

> I would lean towards implementing `parse(Quantity, str)` and `parse(Units, str)`

Yes I think I may agree with this. `Units` is not exported which makes the latter somewhat inconvenient spelling. Should we export it?

> Finally, would you imagine implementing three-arg versions, `Base.parse(::Module, Quantity, str)` or `Base.parse(::Module, Units, str)`, similar to the two-arg version of `parseunit`?

I think the output type should always come first (that's the `Base.parse` convention), so it would be best to have `Base.parse(Quantity, str)` for consistency.  That way generic code can still arguably use `parse(T, str)` with `T` being a `Quantity` or not. Then we need to decide whether the unit module context is passed as a keyword or not. There's some precedent for that; Base supports `parse(Int, str, base=16)`, so I think we could have `parse(Quantity, str, reference_units=[Unitful, ...])` or some such thing.

test/runtests.jl
ajkeller34: The likelihood that someone would make a change causing `"ms"` to parse as meters times seconds is basically zero. You'd really have to go out of your way to make that happen. I don't mind if we add it though.

src/user.jl
ajkeller34: How about "The string will be parsed with standard Julia syntax" ? What you've described is not special to Unitful.

I was actually a bit more confused by the first sentence. Perhaps that could read "Parse a string for Unitful quantities, units, or dimensions."

ajkeller34:
Just some brief feedback:

- 👍 to a keyword argument for `parse`. Perhaps `unitful_modules`?
- 👍 to `uparse`, nice name.
- I was sort of wavering on `parse(::Type{Quantity}, str)`. Providing just one type-unstable way to parse `Quantity`s (`uparse`) seems clean and this way `parse` wouldn't have to stray too far from Base behavior, in that `Quantity` is not a concrete type. Ultimately though, I can see the need for a function like `uparse` that would complain if unexpected input were encountered. For example you could imagine trying to parse a `Quantity` and winding up with `Units` if the numbers were missing, etc. So I think we should do it.
- I'm on board with exporting `Units`.

Let me know how much you want to tackle here. I really appreciate what you've done already. Perhaps we can change `parseunit` to `uparse`, and then the `parse` stuff could go into a separate PR, if you prefer?

c42f:
I had a bit of a look at what it would take to do "something sensible" with `Base.parse`. I do think there's value in it, but I also think it's a fairly wide design space and it might be a lot more work to explore properly.

For example, what do we do about `MixedUnits`? I got a surprise when a half baked version of `parse(Units, "dB")` ended up with an error because `dB` is not a `Units`. Is there an appropriate abstract type here which could be used by the naive user?

>     * Perhaps `unitful_modules`

Right, though it might make sense to support more general "unit contexts" in the future other than fixing on modules? For example it could be nice to support @Datseris's idea of a vector as a collection of allowed units.

>     * Ultimately though, I can see the need for a function like `uparse` that would complain if unexpected input were encountered.

We could have `uparse(Units, str)` for that, and `uparse(str)` as the base version which just tries to figure out what you mean?

> Let me know how much you want to tackle here. I really appreciate what you've done already.

Thanks, I appreciate it! Sometimes a simple looking thing turns out to be a deeper rabbit hole than you might expect. For now I'd be fairly happy to simply rename `parseunit` to `uparse` and develop further from there. Maybe swapping the order of arguments to look more like `parse` though might be appropriate.

c42f:
Ok, I've renamed `parseunits` to `uparse` and I think that's an improvement.

This might be good to merge now. I did rearrange a bit and added a `unit_context` keyword argument for potential consistency with an associated version of `parse()`. Happy to tweak that or swap out for a better name if people can think of one.

ajkeller34:
Beautiful, thank you!

alhirzel:
HECK YEAH, thank you @c42f!

caseykneale:
uparse just saved my day!

pranshumalik14:
It'll also be useful to have this kind of type comparison: `uparse("1cm") <: u"m"`. Instead, what we get is: `ERROR: TypeError: in <:, expected Type, got a value of type Quantity{Int64,𝐋,Unitful.FreeUnits{(cm,),𝐋,nothing}}`

alhirzel:
@pranshumalik14 depending on what you want to do, you may be able to rely on Unitful to make a conversion for you assuming that the result of `uparse` is a `Length`. For instance, the following works and returns `true`:

```julia
dimension(uparse("1cm")) == dimension(1u"m")
```

and the following return approximately 0.01:

```
ustrip(Float64, u"m", uparse("100cm"))
ustrip(Float64, u"m", uparse("0.03280839895ft"))
```

henry2004y:
Is this described in the document? If not then it would be nice to mention it as it is very useful for other packages that depends on Unitful.jl! I found `uparse` only through a discourse question link.

sostock:
No, `uparse` is currently not documented.

PR #303, opened 2020-02-05
Remove code for Julia < 1.0

sostock:
Since newer Unitful versions support only Julia versions ≥ 1.0, the code for supporting older versions can be removed.


giordano:
Thanks!

PR #308, opened 2020-02-21
Fix rounding with digits/sigdigits

### LINKED ISSUE #307: Rounding with digits/sigdigits
`round` with the `digits`/`sigdigits` keyword only really works with quantities based on floating-point numbers:
```julia
julia> round(u"m", (120//1)u"cm", sigdigits=2)
5404319552844595//4503599627370496 m

julia> round(u"m", 120u"cm", sigdigits=2)
ERROR: InexactError: Int64(1.2)
```
For unitless numbers, `round` always returns a float when `sigdigits`/`digits` is used. Maybe Unitful should do the same.
```julia
julia> round(120//100, sigdigits=2)
1.2

julia> round(125, sigdigits=2)
120.0
```

sostock:
Fixes #307

This PR makes `round(u::Units, x::Quantity, digits/sigdigits=...)` return a quantity based on a floating-point number (as does rounding with `digits`/`sigdigits` for unitless numbers).

### DIFF
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -281,9 +281,9 @@ _rounderr() = error("specify the type of the quantity to convert to ",
     "when rounding quantities. Example: round(typeof(1u\"m\"), 137u\"cm\").")
 
 # convenience methods
-round(u::Units, q::Quantity, r::RoundingMode=RoundNearest; kwargs...) =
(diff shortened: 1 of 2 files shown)


ajkeller34:
Thanks!

PR #309, opened 2020-02-23
Support isone

islent:
Hello guys, I've implemented `isone`:
```julia
julia> isone(1.0u"m")
true

julia> isone(1.4u"m")
false
```

In latest version (Unitful@1.0.0):
```julia
julia> isone(1.0u"m")
false

julia> isone(2.0u"m")
false

julia> iszero(0.0u"m")
true

julia> iszero(1.0u"m")
false
```
`iszero` works fine (not implemented), whereas `isone` gives a weird result

### DIFF
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -9,7 +9,7 @@ import Base: exp, exp10, exp2, expm1, log, log10, log1p, log2
 import Base: sin, cos, tan, cot, sec, csc, atan, cis
 
 import Base: eps, mod, rem, div, fld, cld, trunc, round, sign, signbit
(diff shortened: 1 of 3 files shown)


sostock:
Note that “one” is not the same as “one meter”. In fact, `isone` currently does what I would expect it to:
```julia
julia> 1u"m" == 1
false

julia> 1u"cm/m" == 1
false

julia> 100u"cm/m" == 1
true

julia> isone(1u"m")
false

julia> isone(1u"cm/m")
false

julia> isone(100u"cm/m")
true
```

islent:
Got it! So I might have been wishing for `isoneunit`, which is practically useless.

It would be better that these design ideas could be documented! 😄

PR #315, opened 2020-03-15
Provide Base.isless() for LogScaled values

VikoRifeo:


test/runtests.jl
```diff
@@ -1507,6 +1507,11 @@ end
         end
     end
 
+    @testset "> Comparisons" begin
+        @test 3dB < 5dB
+        @test 3dBm < 5dBm
```
giordano, line 1512: I'd add also a test that we cannot test different log scales
```suggestion
        @test 3dBm < 5dBm
        @test_throws MethodError 3dB < 5dBm
```

giordano:
Thanks!

PR #317, opened 2020-03-25
Support mixed argument types to `div` and friends

Tokazama:
This simply makes it so that when something like `div(9m, 3)` it produces `3m`. Currently this errors instead.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Unitful"
 uuid = "1986cc42-f94f-5a68-af5c-568840ba703d"
-version = "1.0.1"
+version = "1.1.0"
 
 [deps]
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Unitful"
 uuid = "1986cc42-f94f-5a68-af5c-568840ba703d"
-version = "1.0.0"
+version = "1.0.1"
 
 [deps]
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
@@ -8,8 +8,8 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [compat]
-julia = "1"
 ConstructionBase = "1"
+julia = "1"
 
 [extras]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -63,9 +63,9 @@ for f in (:div, :fld, :cld)
             ($f)(z.val,y.val)
         end
 
-        ($f)(x::Number, y::AbstractQuantity) = Quantity(($f)(x, y.val), unit(y))
+        ($f)(x::Number, y::AbstractQuantity) = ($f)(x, ustrip(y)) / unit(y)
 
-        ($f)(x::AbstractQuantity, y::Number) = Quantity(($f)(x.val, y), unit(x))
+        ($f)(x::AbstractQuantity, y::Number) = unit(x) * ($f)(ustrip(x), y)
     end
 end
 
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -56,7 +56,7 @@ end
 # ambiguity resolution
 //(x::AbstractQuantity, y::Complex) = Quantity(//(x.val, y), unit(x))
 
-for f in (:div, :fld, :cld)
+for f in (:fld, :cld)
     @eval begin
         function ($f)(x::AbstractQuantity, y::AbstractQuantity)
             z = uconvert(unit(y), x)        # TODO: use promote?
@@ -67,6 +67,19 @@ for f in (:div, :fld, :cld)
 
         ($f)(x::AbstractQuantity, y::Number) = Quantity(($f)(ustrip(x), y), unit(x) / unit(y))
     end
+end
+
+function div(x::AbstractQuantity, y::AbstractQuantity, r::RoundingMode=RoundToZero)
+    z = uconvert(unit(y), x)        # TODO: use promote?
+    div(z.val,y.val, r)
+end
+
+function div(x::Number, y::AbstractQuantity, r::RoundingMode=RoundToZero)
+    Quantity(div(x, ustrip(y), r), unit(x) / unit(y))
+end
+
+function div(x::AbstractQuantity, y::Number, r::RoundingMode=RoundToZero)
+    Quantity(div(ustrip(x), y, r), unit(x) / unit(y))
 end
 
 for f in (:mod, :rem)
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -63,9 +63,9 @@ for f in (:div, :fld, :cld)
             ($f)(z.val,y.val)
         end
 
-        ($f)(x::Number, y::AbstractQuantity) = ($f)(x, ustrip(y)) / unit(y)
+        ($f)(x::Number, y::AbstractQuantity) = Quantity(($f)(x, ustrip(y)), unit(x) / unit(y))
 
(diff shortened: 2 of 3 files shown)


Tokazama:
I've also added a version bump for a patch release

src/quantities.jl
```diff
@@ -57,9 +57,15 @@ end
...
-        ($f)(z.val,y.val)
+    @eval begin
+        function ($f)(x::AbstractQuantity, y::AbstractQuantity)
+            z = uconvert(unit(y), x)        # TODO: use promote?
+            ($f)(z.val,y.val)
+        end
+
+        ($f)(x::Number, y::AbstractQuantity) = Quantity(($f)(x, y.val), unit(y))
+
+        ($f)(x::AbstractQuantity, y::Number) = Quantity(($f)(x.val, y), unit(x))
```
giordano: I'm not sure you should strip the units.  Also, to strip the units use `ustrip`

test/runtests.jl
```diff
@@ -561,7 +561,13 @@ end
         @test m / missing === missing        # Unit / missing
         @test missing / m === missing        # Missing / Unit (// is not defined for Missing)
         @test @inferred(div(10m, -3cm)) === -333
+        @test @inferred(div(10m, 3)) === 3m
+        @test @inferred(div(10, 3m)) === 3m
```
giordano: This result doesn't seem correct

test/runtests.jl
```diff
@@ -561,7 +561,13 @@ end
...
         @test @inferred(fld(10m, -3cm)) === -334
+        @test @inferred(fld(10m, 3)) === 3m
+        @test @inferred(fld(10, 3m)) === 3m
```
giordano: Nor this one

test/runtests.jl
```diff
@@ -561,7 +561,13 @@ end
...
+        @test @inferred(cld(10m, 3)) === 4m
+        @test @inferred(cld(10, 3m)) === 4m
```
giordano: Nor this one

src/quantities.jl
Tokazama: Do you mean I shouldn't strip the units in lines 61-64? That code is copied verbatim from what was there before. I can change it but I was also trying to not mess with whatever style was implemented before.

giordano: But the line before refers to quantities with the same dimension and what it's doing is to promote to common units.  Your code is completely stripping units, so you lose the physical dimensions of the quantities, and you get wrong results, like the tests below.

test/runtests.jl
Tokazama: This just assumes that the non unitful argument has the same units so it's equivalent to `div(10, 3)`. This seems to be what  multiplication does.

src/quantities.jl
giordano: More specifically, I should have put the comment on the other method, that's the most problematic

test/runtests.jl
giordano: 10 divided by 3m has the dimension of the inverse of a length, why the result of `div(10, 3m)` should have the dimension of a length?  Multiplication of a unitful quantities by a pure number preserves the physical dimension, that's not a good example.  You should look at division instead:
```julia
julia> 10 / 3u"m"
3.3333333333333335 m^-1
```

Tokazama: So the tests should return `3 m^-1`?

giordano: I think that would be the most meaningful result, yes

Tokazama: That makes sense.  I'll fix it. Do you want me to put `ustrip` on my code, the old code, or do a comprehensive PR later putting `ustrip` in all appropriate places?

giordano: In the lines you've added, let's not go too far by making unrelated changes :wink:

Tokazama: This should be fixed now.

Project.toml
```diff
@@ -1,15 +1,15 @@
 name = "Unitful"
 uuid = "1986cc42-f94f-5a68-af5c-568840ba703d"
-version = "1.0.0"
+version = "1.0.1"
```
giordano: I think new methods configure as a new feature, so this will need a minor version bump

src/quantities.jl
```diff
@@ -57,9 +57,15 @@ end
...
-        ($f)(z.val,y.val)
+    @eval begin
+        function ($f)(x::AbstractQuantity, y::AbstractQuantity)
+            z = uconvert(unit(y), x)        # TODO: use promote?
+            ($f)(z.val,y.val)
+        end
+
+        ($f)(x::Number, y::AbstractQuantity) = ($f)(x, ustrip(y)) / unit(y)
+
+        ($f)(x::AbstractQuantity, y::Number) = unit(x) * ($f)(ustrip(x), y)
```
giordano: I'd follow that division does here:
```suggestion
        ($f)(x::Number, y::AbstractQuantity) = Quantity(($f)(x, ustrip(y)), unit(x) / unit(y))

        ($f)(x::AbstractQuantity, y::Number) = Quantity(($f)(ustrip(x), y), unit(x) / unit(y))
```

Wikunia:
I just had a look at `div` again and it has a third parameter. 

```
julia> div(4, 3, RoundDown) # Matches fld(4, 3)
  1
  julia> div(4, 3, RoundUp) # Matches cld(4, 3)
  2
  julia> div(5, 2, RoundNearest)
  2
  julia> div(5, 2, RoundNearestTiesAway)
  3
  julia> div(-5, 2, RoundNearest)
  -2
  julia> div(-5, 2, RoundNearestTiesAway)
  -3
  julia> div(-5, 2, RoundNearestTiesUp)
  -2
```

You might want to have this supported as well.

Tokazama:
Good, point. The other thing is that in more recent versions of Julia we are only suppose to overload `div` and other operators are suppose to follow. I'm not sure if I should be addressing this too.

src/quantities.jl
```diff
@@ -69,6 +69,19 @@ for f in (:div, :fld, :cld)
...
+    z = uconvert(unit(y), x)        # TODO: use promote?
+    div(z.val,y.val, r)
+end
+
+function div(x::Number, y::AbstractQuantity, r::RoundingMode=RoundToZero)
+    Quantity(div(x, ustrip(y), r), unit(x) / unit(y))
+end
+
+function div(x::AbstractQuantity, y::Number, r::RoundingMode=RoundToZero)
+    Quantity(div(ustrip(x), y, r), unit(x) / unit(y))
```
Wikunia, line 82: `unit(y)` is a free unit so it is just `unit(x)` isn't it?

Tokazama: You're correct. I was just copying the `/` method above. I'll fixe it.

Tokazama:
Not sure why this worked for me on 1.4 but is broken on everything earlier. I'll look at it more later today.

Wikunia:
The feature I meant with the rounding mode is supported since 1.4 as far as I can see.

Wikunia:
Not sure how to deal with it maybe @giordano can help. You probably want to support Julia 1.0. 
Is this something for Compat?

src/quantities.jl
```diff
@@ -56,13 +56,32 @@ end
...
+        ($f)(x::Number, y::AbstractQuantity) = Quantity(($f)(x, ustrip(y)), unit(x) / unit(y))
+
+        ($f)(x::AbstractQuantity, y::Number) = Quantity(($f)(ustrip(x), y), unit(x))
     end
 end
 
+function div(x::AbstractQuantity, y::AbstractQuantity, r::RoundingMode=RoundToZero)
+    z = uconvert(unit(y), x)        # TODO: use promote?
+    div(z.val,y.val, r)
+end
```
giordano, line 75: I'd do
```suggestion
function div(x::AbstractQuantity, y::AbstractQuantity, r...)
    z = uconvert(unit(y), x)        # TODO: use promote?
    div(z.val,y.val, r...)
end
```
so you don't even have to care to forcibly set a default value

Tokazama:
Oh, then I just need to put the div code with the extra parameter inside a `if VERSION > v"1.3"`` block.

src/quantities.jl
Tokazama: That sounds better than my idea.

PR #319, opened 2020-04-20
Use integer for conversion if ratio is whole

briochemc:
I find that when the conversion factor is whole, it is surprising that a rational is used instead of an integer. For example, I would expect

```julia
using Unitful: minute, s
1minute |> s
```

to give `60s` instead of the current behaviour, `(60//1)s`. 

This very simple PR remedies this by replacing whole conversion factors with their numerator. This keeps Unitful.jl's exact-conversion feature AFAIU but provides a more natural result for very simple cases like the one above.

### DIFF
--- a/src/conversion.jl
+++ b/src/conversion.jl
@@ -38,6 +38,10 @@ Find the conversion factor from unit `t` to unit `s`, e.g. `convfact(m,cm) = 0.0
         end
     end
 
+    if denominator(ex) == 1 # Use an integer for conversion if the ratio is whole
+        ex = numerator(ex)  # so that, e.g., uconvert(s, 1minute) = 60s
+    end
+
     a ≈ 1.0 ? (inex = 1) : (inex = a)
     y = inex * ex
     :($y)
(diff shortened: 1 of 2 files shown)


cstjean:
Is it type-stable?

briochemc:
> Is it type-stable?

AFAIU, yes, but I'm not entirely sure. 

This PR just concerns `convfact`. And my understanding is that `convfact(s::Units, t::Units)`, which would until now return a `Float` or `Rational`, will also return an `Int` for some combinations of `s` and `t` units. I'm assuming that the value of `ex`, [the conversion factor](https://github.com/PainterQubits/Unitful.jl/blob/master/src/conversion.jl#L22), only depends on the types of `s` and `t` (because it depends on `t()` and `s()`?) so that it can be inferred at compile time. If so, that would propagate to `confactor`'s output, and make this type stable, right?

Now, obviously, I am not a professional developer, so maybe my understanding is wrong. I don't know for sure that's how type-stability is assessed. And I also don't know what `@generated` does to `convfact`.

cstjean:
If it's inside a `@generated` function, then it's certainly fine! I hadn't seen it.

briochemc:
@cstjean Could you explain why the `@generated` is sufficient to make it fine? And also does that mean I got the rest right?

cstjean:
The calculation is done while generating the code which will be run. That's done only once, so it doesn't matter if it's type-stable or not. It shouldn't affect the type-stability of the generated code.

ajkeller34:
This looks great to me. I'm going to tidy up the commit history in this PR by cherry-picking the relevant commit from your branch into a new branch based on PainterQubits/master, and merge that.

ajkeller34:
Closed in favor of pull request #323

PR #320, opened 2020-04-21
Allow prefixes for years

briochemc:
So that `u"kyr"`, `u"Myr"`, etc., work.


briochemc:
Can someone review/merge this? (FWIW, I could personally use a new Unitful release with this tiny new feature to cleanly drop dependency on UnitfulAstro, which has prefix-able years.)

PR #322, opened 2020-04-29
Delete show(::IO, ::Type) methods.

### LINKED ISSUE #321: Consider deleting custom show methods for types
I.e. https://github.com/PainterQubits/Unitful.jl/blob/7fb2dfb68c63c0a77b8c859c294d8b7f03f337d6/src/display.jl#L80-L85. There are two reasons for this.

The first is that, according to the compiler team, the base methods are not supposed to be overloaded. It's not documented and not obvious but it's been assumed that `show(::IO, ::Type)` isn't overloaded by user code so doing that can lead to very odd issues. E.g. the definitions in this package might have caused Documenter to hang for one of my packages because Documenter prints the signature of methods with missing docs.

The second reason is that it can be super confusing when types print differently from what they really are. I've had some issues with that when using this package because I based some function signatures on how the types were printed in the REPL only to get an error because the signature didn't match.

andreasnoack:
Closes #321

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1216,9 +1216,9 @@ end
 @testset "Display" begin
     withenv("UNITFUL_FANCY_EXPONENTS" => false) do
         @test string(typeof(1.0m/s)) ==
-            "Quantity{Float64,𝐋*𝐓^-1,FreeUnits{(m, s^-1),𝐋*𝐓^-1,nothing}}"
+            "Quantity{Float64,𝐋 𝐓^-1,FreeUnits{(m, s^-1),𝐋 𝐓^-1,nothing}}"
         @test string(typeof(m/s)) ==
-            "FreeUnits{(m, s^-1),𝐋*𝐓^-1,nothing}"
+            "FreeUnits{(m, s^-1),𝐋 𝐓^-1,nothing}"
         @test string(dimension(1u"m/s")) == "𝐋 𝐓^-1"
         @test string(NoDims) == "NoDims"
     end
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -1216,9 +1216,9 @@ end
 @testset "Display" begin
     withenv("UNITFUL_FANCY_EXPONENTS" => false) do
         @test string(typeof(1.0m/s)) ==
-            "Quantity{Float64,𝐋*𝐓^-1,FreeUnits{(m, s^-1),𝐋*𝐓^-1,nothing}}"
+            "Quantity{Float64,𝐋 𝐓^-1,FreeUnits{(m, s^-1),𝐋 𝐓^-1,nothing}}"
         @test string(typeof(m/s)) ==
-            "FreeUnits{(m, s^-1),𝐋*𝐓^-1,nothing}"
+            "FreeUnits{(m, s^-1),𝐋 𝐓^-1,nothing}"
```
ajkeller34, line 1221: If you were to type this string at the REPL, since `L T^-1` would not be recognized as multiplication, the type would not be valid. Perhaps we should also change how dimensions are displayed, to always show the multiplication operators and avoid this problem?

andreasnoack: I'm not sure what the best solution is but it's probably a good idea to avoid printing something that looks like type signatures but aren't. My main aim in this PR, though, is to get rid of two show methods since they break the compiler on downstream problems

andreasnoack:
It would help us a lot if this could make it into a release.

Issue #326, opened 2020-05-11
round(n; sigdigits) without explicit units


mcabbott:
While being able to round in a unit-controlled way is neat, what's the reason to disallow simple rounding? I expected the second of these to work, and if it did, then code with & without units would behave the same:
```julia
len = (120//1)u"cm"
round(u"m", len, sigdigits=2)
ustrip(len) # plain number

tim = 3.0000001u"s"
ustrip(tim)
round(tim, sigdigits=3) # ERROR: specify the type of the quantity to convert to when rounding
```

KronosTheLate:
I feel like perfect is the enemy of the good here.

When I want to round a `Quantity`, if I do not specify the output unit, I expect it to be assumed to be the input unit. It is just an extremely sensible default, and the *only* sensible default, as we would never want the output to have a different unit than the input by default. 

The issues that arise with inequality when changing the order of rounding and unit conversion is a bummer, but there is no way around it - the information is lost during rounding. I honestly think that you are crazy if you expect strict commutativity between rounding and unit conversion, and that you would be foolish to use rounding and unit conversion without extreme care if you are doing anything important. And in the extreme care case, simply specify the output unit! But please allow the quick-and-dirty "let me remove some visual noise" rounding to be done without specifying the output unit, when such a sensible default exists.

KronosTheLate:
As for requiring the keyword argument `sigdigits`, I do not see why it should not default to the defaults from `Base.round`. If the user wants anything else they are free to specify it, but consistency with `Base` is really something that reduces friction in the ecosystem, and something that I greatly appreciate.

david-macmahon:
I'm writing a Plots recipe and would like to support user specified rounding for tick labels, but because of this issue it ends up being somewhat complicated:

```julia
julia> x = 1.2345678
1.2345678

julia> xs = x * 1u"s"
1.2345678 s

julia> round(x, digits=2)
1.23

julia> round(xs, digits=2)
ERROR: specify the type of the quantity to convert to when rounding quantities. Example: round(typeof(1u"m"), 137u"cm").

julia> round(typeof(xs), xs, digits=2)
1.23 s

julia> round(typeof(x), x, digits=2)
ERROR: MethodError: no method matching round(::Type{Float64}, ::Float64; digits=2)
```

[...]

PR #328, opened 2020-05-12
Allow rounding with sigdigits

mcabbott:
Minimal fix for #326

### DIFF
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -296,8 +296,8 @@ _dimerr(f) = error("$f can only be well-defined for dimensionless ",
 isinteger(x::AbstractQuantity) = _dimerr(isinteger)
 isinteger(x::DimensionlessQuantity) = isinteger(uconvert(NoUnits, x))
 
-_rounderr() = error("specify the type of the quantity to convert to ",
-    "when rounding quantities. Example: round(typeof(1u\"m\"), 137u\"cm\").")
+_rounderr() = error("when rounding quantities, you must either specify the quantity to ",
+    "convert to, or provide keyword sigdigits. Example: round(typeof(1u\"m\"), 137u\"cm\").")
 
 # convenience methods
 round(u::Units, q::AbstractQuantity, r::RoundingMode=RoundNearest; kwargs...) =
@@ -307,8 +307,12 @@ round(::Type{T}, u::Units, q::AbstractQuantity, r::RoundingMode=RoundNearest;
     round(Quantity{T, dimension(u), typeof(u)}, q, r; kwargs...)
 
 # workhorse methods
-round(x::AbstractQuantity, r::RoundingMode=RoundNearest; kwargs...) =
+round(x::AbstractQuantity, r::RoundingMode; kwargs...) =
     _rounderr()
+function round(x::AbstractQuantity; sigdigits::Union{Nothing,Integer}=nothing, kwargs...)
+    sigdigits === nothing && _rounderr()
+    round(unit(x), x; sigdigits=sigdigits, kwargs...)
+end
 round(x::DimensionlessQuantity; kwargs...) = round(uconvert(NoUnits, x); kwargs...)
 round(x::DimensionlessQuantity, r::RoundingMode; kwargs...) =
     round(uconvert(NoUnits, x), r; kwargs...)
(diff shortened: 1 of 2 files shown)


gustaphe:
This is a nobrainer to me, and I was looking into trying to fix whatever made the build fail, but that coveralls message is incomprehensible.

mcabbott:
These things always seem noisy. It does say "4 of 4 new or added lines in 1 file covered. (100.0%)"

JeffFessler (approved):

JeffFessler:
bump :)

src/quantities.jl
```diff
@@ -307,8 +307,12 @@ round(::Type{T}, u::Units, q::AbstractQuantity, r::RoundingMode=RoundNearest;
     round(Quantity{T, dimension(u), typeof(u)}, q, r; kwargs...)
 
 # workhorse methods
-round(x::AbstractQuantity, r::RoundingMode=RoundNearest; kwargs...) =
+round(x::AbstractQuantity, r::RoundingMode; kwargs...) =
     _rounderr()
+function round(x::AbstractQuantity; sigdigits::Union{Nothing,Integer}=nothing, kwargs...)
+    sigdigits === nothing && _rounderr()
+    round(unit(x), x; sigdigits=sigdigits, kwargs...)
```
sostock, line 314:
```suggestion
function round(x::AbstractQuantity, r::RoundingMode=RoundNearest; sigdigits::Union{Nothing,Integer}=nothing, kwargs...)
    sigdigits === nothing && _rounderr()
    round(unit(x), x, r; sigdigits=sigdigits, kwargs...)
```
So that we allow `round(::AbstractQuantity, ::RoundingMode; sigdigits)` as well.

sostock:
I’m not sure I like this, since it is not invariant under unit conversion:
```julia
julia> x = 2.5u"inch"
2.5 inch

julia> y = uconvert(u"cm", x)
6.35 cm

julia> x == y
true

julia> round(x, sigdigits=2) ≈ round(y, sigdigits=2)
false
```

mcabbott:
After rounding, shouldn't the test be approximate equality at the resolution you chose? The default is 8 digits and you've declared that you don't care about those. Of course `sigdigits` isn't a very precise specification of `rtol`, but I hope we're not keeping too many digits on the number of digits we're keeping. 
```
julia> @test x ≈ round(y, sigdigits=2) rtol=0.01
Test Passed
  Expression: ≈(x, round(y, sigdigits = 2), rtol = 0.01)
   Evaluated: ≈(2.5 inch, 6.4 cm; rtol = 0.01)

julia> @test y ≈ round(x, sigdigits=2) rtol=0.01
Test Passed
  Expression: ≈(y, round(x, sigdigits = 2), rtol = 0.01)
   Evaluated: ≈(6.35 cm, 2.5 inch; rtol = 0.01)

julia> @test round(x, sigdigits=2) ≈ round(y, sigdigits=2) rtol=0.01
Test Passed
  Expression: ≈(round(x, sigdigits = 2), round(y, sigdigits = 2), rtol = 0.01)
   Evaluated: ≈(2.5 inch, 6.4 cm; rtol = 0.01)
```

sostock:
It’s a goal of this package that all functionality is unit-invariant. That’s why `round` takes a unit argument in the first place.

For `x == y`, you will always have `round(u"m", x, sigdigits=2) == round(u"m", y, sigdigits=2)` (ignoring floating-point effects) even when `x` and `y` have different units. This is not the case when you do not specify the first argument.

However, there are already some non-invariant methods in this package, so I guess one more doesn’t hurt. For example, I really don’t like that `div(5000u"m", 2) != div(5u"km", 2)`, but other maintainers don’t think it’s too bad. I’m not generally opposed to this PR. I just wanted to mention that there is a reason that `round` without a unit argument was forbidden, and this PR ignores that reason. Even though the non-invariance may be less significant when using `sigdigits`, I don’t think it is completely unsignificant.

test/runtests.jl
```diff
@@ -950,6 +950,7 @@ end
     @test ceil(u"m", (125//1)u"cm", sigdigits=2) === 1.3u"m"
     @test trunc(u"m", -125u"cm", sigdigits=2) === -1.2u"m"
     @test trunc(u"m", (125//1)u"cm", sigdigits=2) === 1.2u"m"
+    @test round(3.7001u"m", sigdigits=3) === 3.7u"m"
```
sostock, line 953:
```suggestion
    @test round(3.7001u"m", sigdigits=3) === 3.7u"m"
    @test round(3.7001u"m", RoundUp, sigdigits=3) === 3.71u"m"
```
Since specifying a `RoundingMode` is now allowed, it should be tested as well.

PR #329, opened 2020-05-12
Converting Dates.FixedPeriod to Quantity{...,𝐓}

### LINKED ISSUE #125: Support automatic conversions from `TimeType`s

mcabbott:
Fixes #125, or at least makes a start. 

Edit: I moved this to a new file, as it needs to come after the relevant units have been defined. 
```julia
julia> uconvert(s, Microsecond(1))
1//1000000 s

julia> convert(typeof(1f0s), Minute(1))
60.0f0 s

julia> promote(Minute(1), 1s) 
(60 s, 1 s)
```
Should things like this be made to work?
```julia
julia> Minute(1) + 1s
ERROR: MethodError: no method matching +(::Minute, ::Quantity{Int64,𝐓,Unitful.FreeUnits{(s,)
```

### DIFF
--- a/src/dates.jl
+++ b/src/dates.jl
@@ -13,7 +13,7 @@ Quantity(x::Dates.Week) = x.value * u"wk"
 uconvert(a::Units, x::Dates.FixedPeriod) = uconvert(a, Quantity(x))
 
 Base.promote_rule(::Type{Quantity{T,D,U}}, ::Type{<:Dates.FixedPeriod}) where {T,D,U} =
-    Quantity{promote_type(T,Int),D,U}
+    Quantity{promote_type(T,Int64),D,U}
 
 convert(type::Type{<:Quantity}, x::Dates.FixedPeriod) = convert(type, Quantity(x))
 
--- a/src/dates.jl
+++ b/src/dates.jl
@@ -1,0 +1,20 @@
+
+
+# Dates.FixedPeriod == Union{Day, Hour, Microsecond, Millisecond, Minute, Nanosecond, Second, Week}
+Quantity(x::Dates.Nanosecond) = x.value * u"ns"
+Quantity(x::Dates.Microsecond) = x.value * u"μs"
+Quantity(x::Dates.Millisecond) = x.value * u"ms"
+Quantity(x::Dates.Second) = x.value * u"s"
+Quantity(x::Dates.Minute) = x.value * u"minute"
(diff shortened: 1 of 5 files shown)


sostock:
I also started to implement this some time ago, I have uploaded it here so you can maybe use some of my code:

https://gist.github.com/sostock/68e5eda1abee911116dcc08970843d0c

The broken tests should work on Julia ≥ 1.5, since https://github.com/JuliaLang/julia/issues/34645 was fixed recently.

src/dates.jl
```diff
@@ -0,0 +1,29 @@
...
+Quantity(x::Dates.Second) = x.value * u"s"
+Quantity(x::Dates.Minute) = x.value * u"minute"
+Quantity(x::Dates.Hour) = x.value * u"hr"
+Quantity(x::Dates.Day) = x.value * u"d"
+Quantity(x::Dates.Week) = x.value * u"wk"
+
+uconvert(a::Units, x::Dates.FixedPeriod) = uconvert(a, Quantity(x))
+
+Base.promote_rule(::Type{Quantity{T,D,U}}, ::Type{<:Dates.FixedPeriod}) where {T,D,U} =
+    Quantity{promote_type(T,Int),D,U}
```
sostock, line 16:
```suggestion
    Quantity{promote_type(T,Int64),D,U}
```
The types from `Dates` use `Int64` internally, so this promotion rule should probably be changed to reflect that.

mcabbott: OK, I hadn't checked.

mcabbott:
OK, perhaps I should close this in favour of that, you've taken this far more seriously. 

Would the plan be to just not run those tests on Julia < 1.5?

sostock:
> OK, perhaps I should close this in favour of that, you've taken this far more seriously.

If you want, I can make a PR with my changes. I think I didn’t do that because I wanted to implement some more stuff (like support for `CompoundPeriod`s) before making a PR.

> Would the plan be to just not run those tests on Julia < 1.5?

Yes, that’s how I would do it.

mcabbott:
OK, no great rush, but having even the basics would be nice.

sostock:
I opened #331 as a replacement.

PR #331, opened 2020-05-14
Add support for `Dates.FixedPeriod` and `Dates.CompoundPeriod`

sostock:
This is an attempt at #125.

Things to (maybe) add:
- [x] Documentation
- [x] Support for `CompoundPeriod`s (if they only contain `FixedPeriod`s)

### DIFF
--- a/test/dates.jl
+++ b/test/dates.jl
@@ -34,7 +34,7 @@
             @test Int64(3)u"m" * Day(1) === Int64(3)u"m*d"
             @test 1.0f0u"m" * Microsecond(-2) === -2.0f0u"m*μs"
             @test Hour(4) * Int32(2)u"m" === Int64(8)u"hr*m"
-            @test Second(5) * (3//2)u"Hz" === Rational{Int64}(15//2)u"s*Hz"
+            @test Second(5) * (3//2)u"Hz" === Rational{Int64}(15,2)u"s*Hz"
             @test Second(5) * 2u"1/s" === Int64(10)
             @test 2.5u"1/s^2" * Second(2) === 5.0u"1/s"
             @test_throws AffineError Second(1) * 1u"°C"
@@ -59,8 +59,8 @@
             @test 5u"m" / Hour(2) === 2.5u"m/hr"
             @test 5.0f0u"m" / Hour(2) === 2.5f0u"m/hr"
             # //
-            @test Nanosecond(10) // 2u"m" === Rational{Int64}(5//1)u"ns/m"
-            @test 5u"m" // Hour(2) === Rational{Int64}(5//2)u"m/hr"
+            @test Nanosecond(10) // 2u"m" === Rational{Int64}(5,1)u"ns/m"
+            @test 5u"m" // Hour(2) === Rational{Int64}(5,2)u"m/hr"
             # div
             @test div(Second(1), 2u"ms") == div(1u"s", Millisecond(2)) == 500
             @test div(Second(11), 2u"s") == div(11u"s", Second(2)) == 5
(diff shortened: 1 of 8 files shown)


sostock:
I now added support for `CompoundPeriod`s. They can be converted to `Quantity`s using `convert`/`uconvert`, but don’t support all operations that are supported by `FixedPeriod`s. For example:

* `unit(::CompoundPeriod)` errors, while `unit(::FixedPeriod)` returns the appropriate unit (e.g., `unit(Nanosecond(5)) == u"ns"`).
* `Quantity(::CompoundPeriod)` errors, while `Quantity(::FixedPeriod)` returns the appropriate quantity (e.g., `Quantity(Nanosecond(5)) == 5u"ns"`).
* Arithmetic operations with `CompoundPeriod`s error if the resulting unit would depend on the input units, e.g.:
  - `+(::Quantity, ::CompoundPeriod)` and `+(::CompoundPeriod, ::Quantity)` error.
  - `div(::Quantity, ::CompoundPeriod)` and `div(::CompoundPeriod, ::Quantity)` work, since the result is dimensionless.
  - `mod(::CompoundPeriod, ::Quantity)` works while `mod(::Quantity, ::CompoundPeriod)` does not, since the second argument determines the output unit.

We could either leave it like that, or decide between two options to enable all arithmetic operations between `Quantity`s and `CompoundPeriod`s:

* Define `unit(::CompoundPeriod) = u"s"` and treat it as such in all arithmetic operations. This is consistent with the addition between quantities with different time units: `5u"ms" + 1u"hr" == (720001//200)u"s"`
* Don’t define `unit(::CompoundPeriod)`, but convert the `CompoundPeriod` to the unit of the `Quantity` when used in an arithmetic operation between a `CompoundPeriod` and `Quantity`.

test/dates.jl
```diff
@@ -0,0 +1,786 @@
...
+            @test Int64(3)u"m" * Day(1) === Int64(3)u"m*d"
+            @test 1.0f0u"m" * Microsecond(-2) === -2.0f0u"m*μs"
+            @test Hour(4) * Int32(2)u"m" === Int64(8)u"hr*m"
+            @test Second(5) * (3//2)u"Hz" === Rational{Int64}(15,2)u"s*Hz"
+            @test Second(5) * 2u"1/s" === Int64(10)
+            @test 2.5u"1/s^2" * Second(2) === 5.0u"1/s"
+            @test_throws AffineError Second(1) * 1u"°C"
+            @test_throws AffineError 1u"°C" * Second(1)
+            # Multiplication with unit
+            Week(5) * u"Hz" === 5.0u"wk*Hz"
```
cstjean, line 47:
```suggestion
            @test Week(5) * u"Hz" === 5.0u"wk*Hz"
```
?

sostock: Of course, thanks :)

sostock:
I now added documentation. If we are ok with the behavior described in https://github.com/PainterQubits/Unitful.jl/pull/331#issuecomment-640040009, this should be good to review and merge.

OliverEvans96:
What's the current status of this PR? Is there more work that needs to be done before it can be merged? If some of the above issues are deal-breakers, I could take a shot at addressing them.

sostock:
If we are fine with the current behavior (see https://github.com/PainterQubits/Unitful.jl/pull/331#issuecomment-640040009), then this should be good to merge.

One issue remains: since `Period`s and `Quantity`s can now compare equal, they should also hash the same. But hashing is broken for this package anyway, so I didn’t address that for now. Hashing can only be repaired by changing the promotion behavior of logarithmic quantities, which I would consider breaking (see #402). For previous discussion on the hashing issue see #378, #379.

OliverEvans96:
It seems like fixing arithmetic would be good. I would definitely be surprised to find that it doesn't work.

I'm having a hard time deciding which of your solutions I like better. On one hand, it seems a little strange to say `unit(::CompundPeriod)` = seconds since it doesn't really just have a single unit. And would this option lead to any `InexactError`s for `CompoundPeriod`s which are not integer multiples of a second?

But on the other hand, converting to the units of the `Quantity` feels a little bit too magical, specifically in the case of `mod`. That definitely seems like an edge-case, but good to get it right.

Does anyone else have thoughts?

cstjean:
> It seems like fixing arithmetic would be good. I would definitely be surprised to find that it doesn't work.

My 2 cents is that this is already a big PR, which is overdue for a review and merge. The proposed arithmetic implementation doesn't look "obviously right" to me, so it seems wiser to leave it to a future PR than to risk having to deal with a breaking change if a mistake is made.

giordano (approved):
Looks good to me.  I think it's better to not add more complexity here.  If there are no objections, I'll merge tomorrow

ericphanson:
mind tagging a release with this feature?

giordano:
Done, this is in [`v1.7.0`](https://github.com/PainterQubits/Unitful.jl/releases/tag/v1.7.0)

ericphanson:
thanks!

PR #333, opened 2020-05-14
Test on Julia 1.4

sostock:
This adds Julia 1.4 to the CI matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ julia:
   - 1.1
   - 1.2
   - 1.3
+  - 1.4
   - nightly
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,7 @@ os:
 #  - osx    # locally test on my MacBook Pro; OS X tests take too long on Travis CI
 julia:
   - 1.0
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -4,6 +4,7 @@ environment:
   - julia_version: 1.1
   - julia_version: 1.2
   - julia_version: 1.3
```
giordano, line 6: I'd personally remove these versions, they're unsupported.  Also, the line above should be `1.0`, not `1`, we're testing v1.4 twice.

.travis.yml
```diff
@@ -8,6 +8,7 @@ julia:
   - 1.1
   - 1.2
   - 1.3
```
giordano, line 10: Same as the AppVeyor, I'd remove all these versions

appveyor.yml
sostock: OK, I removed them. We are now testing on `1.0` and `1`, so this (hopefully) does not have to be updated until the next Julia LTS release.

.travis.yml
sostock: Done :)

Issue #327, opened 2020-05-11
denominator error when converting fractional units


tomyun:
On Unitful 1.1.0,

```julia
julia> 1u"√m" |> u"√cm"
10.0 cm¹ᐟ²
```

After Unitful 1.2.0 with #323 (#319),

```julia
julia> 1u"√m" |> u"√cm"
ERROR: MethodError: no method matching denominator(::Float64)
Closest candidates are:
  denominator(::Integer) at rational.jl:232
  denominator(::Rational) at rational.jl:233
Stacktrace:
 [1] #s37#154 at /Users/tomyun/.julia/packages/Unitful/9XsxK/src/conversion.jl:41 [inlined]
 [2] #s37#154(::Any, ::Any, ::Any) at ./none:0
 [3] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any,N} where N) at ./boot.jl:526
 [4] uconvert(::Unitful.FreeUnits{(cm¹ᐟ²,),𝐋¹ᐟ²,nothing}, ::Quantity{Int64,𝐋¹ᐟ²,Unitful.FreeUnits{(m¹ᐟ²,),𝐋¹ᐟ²,nothing}}) at /Users/tomyun/.julia/packages/Unitful/9XsxK/src/conversion.jl:78
 [5] FreeUnits at /Users/tomyun/.julia/packages/Unitful/9XsxK/src/types.jl:92 [inlined]
 [6] |>(::Quantity{Int64,𝐋¹ᐟ²,Unitful.FreeUnits{(m¹ᐟ²,),𝐋¹ᐟ²,nothing}}, ::Unitful.FreeUnits{(cm¹ᐟ²,),𝐋¹ᐟ²,nothing}) at ./operators.jl:823
 [7] top-level scope at REPL[28]:1
```

giordano:
CC: @briochemc

briochemc:
I made a PR (https://github.com/PainterQubits/Unitful.jl/pull/334) that adds

```julia
Base.denominator(::Float64) = 1
Base.numerator(x::Float64) = x
```

as a quick hack but this but I'm pretty sure this is not the right way... I'll try something a bit cleaner...

briochemc:
I made a second PR (https://github.com/PainterQubits/Unitful.jl/pull/335) that instead adds a very simple `convertwholeratio` function. (I think while `convertwholeratio` is not type-stable, it's OK because it is called inside the generated `convfact`.)

[...]

PR #334, opened 2020-05-16
Update conversion.jl

briochemc:
Suggestion to fix https://github.com/PainterQubits/Unitful.jl/issues/327


giordano:
Can you please add also a regression test so that we don't break this again in the future?

briochemc:
> Can you please add also a regression test so that we don't break this again in the future?

Sorry I made a mess 😅 FYI, I opened another PR (https://github.com/PainterQubits/Unitful.jl/pull/335) with a test for issue #327

PR #335, opened 2020-05-16
Fix conversion of fractional units

### LINKED ISSUE #327: denominator error when converting fractional units

briochemc:
Fix #327

### DIFF
--- a/src/conversion.jl
+++ b/src/conversion.jl
@@ -38,8 +38,8 @@ Find the conversion factor from unit `t` to unit `s`, e.g. `convfact(m,cm) = 0.0
         end
     end
 
-    if ex isa Rational && denominator(x) == 1
-        ex = numerator(x)
+    if ex isa Rational && denominator(ex) == 1
+        ex = numerator(ex)
     end
     
     a ≈ 1.0 ? (inex = 1) : (inex = a)
--- a/src/conversion.jl
+++ b/src/conversion.jl
@@ -38,14 +38,20 @@ Find the conversion factor from unit `t` to unit `s`, e.g. `convfact(m,cm) = 0.0
         end
     end
 
-    if denominator(ex) == 1 # Use an integer for conversion if the ratio is whole
-        ex = numerator(ex)  # so that, e.g., uconvert(s, 1minute) = 60s
-    end
+    ex = convertwholeratio(ex)
 
     a ≈ 1.0 ? (inex = 1) : (inex = a)
     y = inex * ex
     :($y)
 end
+
+"""
+    convertwholeratio(x)
+
+Removes the denominator of whole rationals.
+"""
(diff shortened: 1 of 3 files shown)


briochemc:
@ajkeller34 could you check this?

BTW, I think the comments in line 14–15 that say:

```julia
    # fact1 is what would need to be multiplied to get to base SI units
    # fact2 is what would be multiplied to get from the result to base SI units
```

should be updated to refer to `inex1` and `inex2` instead? (But I'm not sure, so I did not suggest a correction for that.)

src/conversion.jl
```diff
@@ -38,10 +38,10 @@ Find the conversion factor from unit `t` to unit `s`, e.g. `convfact(m,cm) = 0.0
         end
     end
 
-    if denominator(ex) == 1 # Use an integer for conversion if the ratio is whole
-        ex = numerator(ex)  # so that, e.g., uconvert(s, 1minute) = 60s
+    if ex isa Rational && denominator(ex) == 1
+        ex = numerator(ex)
     end
-
+
```
giordano, line 44:
```suggestion

```

PR #354, opened 2020-06-11
Throw error for div etc. with affine quantities

### LINKED ISSUE #350: div, mod etc. for affine quantities
IMO, the following results are wrong:
```julia
julia> div(5u"°C", 1u"°C") # Should be 1, or throw an error
5

julia> mod(5u"°C", 1u"°C") # Should be 4 °C, or throw an error
0 °C
```
As most other arithmetic operations on affine quantities error, these should probably too.

sostock:
Throws an `AffineError` when calling `div`, `mod`, `rem` etc. with one or more `AffineQuantity`s.

### DIFF
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -8,7 +8,7 @@ import Base: complex, widen, reim # handled in complex.jl
 import Base: exp, exp10, exp2, expm1, log, log10, log1p, log2
 import Base: sin, cos, tan, cot, sec, csc, atan, cis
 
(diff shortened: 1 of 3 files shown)


sostock:
I just noticed that something like `div(5cm, 2)` also doesn’t throw an error. Therefore I will close this PR and upload one that addresses this problem as well.

sostock:
Reopened (see https://github.com/PainterQubits/Unitful.jl/issues/350#issuecomment-643731486).

PR #355, opened 2020-06-11
Enable Travis on macOS

giordano:
The comment says that tests take too long on macOS, I'm a bit surprised by this, let's see how it goes


giordano:
It's true that is a bit slower, but we have to wait for AppVeyor, too, anyway

Issue #285, opened 2019-11-16
Printing of Complex Unitful quantities is confusing. 


MasonProtter:
When one does
```julia
julia> using Unitful

julia> (1 + im)u"m"
1 + 1im m
```
it looks like this is saying `1 + (1im m)` instead of `(1 + 1im) m`. It'd be good to have a special case for `show` so that this is printed more clearly.

c42f:
Seems quite related to https://github.com/PainterQubits/Unitful.jl/issues/199.

Perhaps we should flip the default `show` to just add brackets in the style of https://github.com/PainterQubits/Unitful.jl/pull/256, with special cases to opt out?

PR #364, opened 2020-07-06
Fix printing of mixed quantities

### LINKED ISSUE #342: Printing of “mixed quantities”
The documentation states (in logarithm.md):
> We refer to a quantity with both logarithmic "units" and linear units as a mixed quantity. For mixed quantities, the numeric value associates with the logarithmic unit, and the quantity is displayed in a way that makes this explicit:
> ```
> julia> (0u"dBm")/u"Hz"
> [0.0 dBm] Hz^-1
> 
> julia> (0u"dB")/u"Hz"
> [0 dB] Hz^-1
> 
> julia> 0u"dB/Hz"
> [0 dB] Hz^-1
> ```

However, these quantities are not actually printed like that:
```julia
julia> (0u"dBm")/u"Hz"
0.0 dBm Hz^-1

julia> (0u"dB")/u"Hz"
0 dB Hz^-1

julia> 0u"dB/Hz"
0 dB Hz^-1
```

What should be changed, the documentation or the printing?

Edit: When converting to `String`, the brackets actually show up:
```julia
julia> (0u"dBm")/u"Hz"
0.0 dBm Hz^-1

julia> string(ans)
"[0.0 dBm] Hz^-1"
```

sostock:
Fixes #342.

### DIFF
--- a/src/logarithm.jl
+++ b/src/logarithm.jl
@@ -323,6 +323,16 @@ function Base.show(io::IO, x::Quantity{<:Union{Level,Gain},D,U}) where {D,U}
     nothing
 end
 
+function Base.show(io::IO, mime::MIME"text/plain", x::Quantity{<:Union{Level,Gain}})
+    print(io, "[")
+    show(io, mime, x.val)
+    print(io, "]")
+    if !isunitless(unit(x))
+        print(io," ")
+        show(io, mime, unit(x))
+    end
+end
+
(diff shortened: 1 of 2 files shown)


sostock:
Since it would be good if complex quantities were also printed with brackets (#285), I will open a new PR that addresses both complex and mixed quantities.

PR #366, opened 2020-07-11
Fix printing of complex and logarithmic quantities

### LINKED ISSUE #285: Printing of Complex Unitful quantities is confusing.

### LINKED ISSUE #342: Printing of “mixed quantities”

sostock:
Fixes #285. Fixes #342.

This adds brackets to the printing of complex and mixed logarithmic quantities:
```julia
julia> 1.2u"m"
1.2 m

julia> (2.5+1.0im)u"V"
(2.5 + 1.0im) V

julia> 3u"dB/Hz"
[3 dB] Hz^-1
```

This is accomplished by adding a new trait `BracketStyle` which specifies which kind of brackets (or no brackets) are used for a given type.

### DIFF
--- a/docs/src/display.md
+++ b/docs/src/display.md
@@ -6,10 +6,12 @@ variable `UNITFUL_FANCY_EXPONENTS` to either `true` or `false` to force using or
 the exponents.
 
 ```@docs
+Unitful.BracketStyle
 Unitful.abbr
 Unitful.prefix
 Unitful.show(::IO, ::Quantity)
 Unitful.show(::IO, ::Unitful.Unitlike)
 Unitful.showrep
+Unitful.showval
 Unitful.superscript
 ```
--- a/src/display.jl
+++ b/src/display.jl
@@ -48,6 +48,57 @@ function show(io::IO, x::Unit{N,D}) where {N,D}
     show(io, FreeUnits{(x,), D, nothing}())
 end
 
+abstract type BracketStyle end
+
+struct NoBrackets <: BracketStyle end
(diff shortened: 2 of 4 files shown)


giordano:
Nice!

PR #390, opened 2020-10-24
Macro hygeine

bramtayl:
I made some fixes for more hygienic macros. It's good to be careful inside `esc` to interpolate in as much as possible, otherwise, you could accidentally end up referring to something in the user's namespace. In unescaped code, you can refer to things in your namespace safely.


bramtayl:
Bump?

PR #397, opened 2020-11-13
Added magnetic CGS units

KronosTheLate:
In this PR, I have added the CGS units for:
Magnetic flux                                                    (Maxwell)
Magnetic flux density                                       (Gauss)
Magnetic intensity/magnetic auxiliary field     (Oersted).

The gauss has the symbol `Gs`, which is taken by the SI-unit gigaseconds. I found no abbreviation for gauss I was happy with, and landed on "Gauss" as the symbol. While inelegant, it is clear. Please suggest an improvement if you can think of one.

I have little understanding of the inner workings of Unitful.jl, so feel free to point out mistakes if you see any.


KronosTheLate:
Now, a month and a half later, I would love to see some feedback if there is an issue with this PR, or to see it merged. I know that as an open-source project it is not fair to have any expectations or demands on the project managers - I am simply nervous about this PR getting a conflict or something like it, which would be non-trivial to me, as a very inexperienced GitHub user.

PR #400, opened 2020-11-20
Restrict `Level`s to real quantities

### LINKED ISSUE #398: Complex `Level`
It is possible to create a complex `Level`:
```julia
julia> @dB (3+2im)V/1V
11.139433523068368 + 5.1073257213090795im dBV

julia> @dB (3+2im)V/(1+1im)V
8.129133566428555 - 1.714556047900126im dB ((1 + 1im) V)
```
Should this be restricted to real quantities? It is not possible to create a complex `Gain`, so this seems inconsistent (and not very useful, at least to me).

And if we want to keep complex `Level`s, the printing needs to be fixed (there need to be parentheses around the complex number).

sostock:
Fixes #398. This allows only real quantities in `Level`s (for the value as well as the reference quantity).

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -228,7 +228,7 @@ abstract type LogScaled{L<:LogInfo} <: Number end
 const RealQuantity = Union{Real, AbstractQuantity{<:Real}}
 
 """
-    struct Level{L, S, T<:Number} <: LogScaled{L}
+    struct Level{L, S, T<:Union{Real, AbstractQuantity{<:Real}}} <: LogScaled{L}
 A logarithmic scale-based level. Details about the logarithmic scale are encoded in
 `L <: LogInfo`. `S` is a reference quantity for the level, not a type. This type has one
 field, `val::T`, and the log of the ratio `val/S` is taken. This type differs from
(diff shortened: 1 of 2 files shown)


src/types.jl
```diff
@@ -225,24 +225,25 @@ used in promotion to put levels and gains onto a common log scale.
 """
 abstract type LogScaled{L<:LogInfo} <: Number end
 
+const RealQuantity = Union{Real, AbstractQuantity{<:Real}}
+
 """
     struct Level{L, S, T<:Number} <: LogScaled{L}
```
giordano, line 231: Update also the docstring?

src/types.jl
```diff
@@ -279,6 +280,8 @@ const RootPowerRatio{T} = IsRootPowerRatio{true,T}
 @inline unwrap(x::IsRootPowerRatio) = x.val
 @inline unwrap(x) = x
 
+const ReferenceQuantity = Union{RealQuantity, IsRootPowerRatio{S,<:RealQuantity} where S}
+
```
giordano, line 284: Maybe define before the first use?

sostock: For that, we would need to move the definition of `IsRootPowerRatio` as well. I have no problem with that, what do you think?

src/types.jl
sostock: Done!

src/types.jl
giordano: Ok, yeah, it's better to not move stuff around for no good reason

PR #403, opened 2020-12-07
Add latexify functions

### LINKED ISSUE #167: [suggestion] Latex format?
Is there a printing function or macro that is readable as a LaTeX string? E.g., something like
```julia
julia> α = 3u"m/s"
3 m s^-1

julia> latexify(α)
L"3\,mathrm{m}\,\mathrm{s}^{-1}"
```
where the important details are
- the `\,` spaces between value and units (and between units),
- the `\mathrm` commands to make sure the unit symbols are upright and not in "math-mode",
- and the exponents are inside brackets because LaTeX does not give precedence to the entire exponent — and would print only the first character (the `-`, minus) as an exponent (instead of the full `-1` exponent in the example above).

gustaphe:
Fixes #167 

Implement a latexrecipe (to interface with the Latexify package).
```julia
julia> latexify(612.2u"nm")
L"$612.2\;\mathrm{nm}$"
julia> latexify(u"m/s^2",unitformat=:siunitx)
L"\si{\meter\per\second\tothe{2}}"
```


giordano:
I'm not very happy about adding these extra dependencies which will pull in turn more dependencies

goretkin:
Is it possible to not express a dependency on `Latexify.jl` and instead use `Requires.@require`, like what was done (in reverse) here: https://github.com/korsbo/Latexify.jl/blob/36aa8a2a047011ef2f29e9960680de3163305d97/src/Latexify.jl#L51-L68

goretkin:
> I'm not very happy about adding these extra dependencies which will pull in turn more dependencies

I agree. Look

https://juliahub.com/ui/Packages/Unitful/5WYj5/1.5.0?t=1
```
Direct Dependencies (4)
Indirect Dependencies (1)
```


https://juliahub.com/ui/Packages/Latexify/AItXo/0.14.6?t=1
```
Direct Dependencies (8) 
Indirect Dependencies (24)
```

gustaphe:
Yeah, I get that. It seems different packages have different routines around that, and this package didn't already use Requires. But I can make a quick rewrite.

goretkin:
In my opinion, `latexprefixdict` should be defined regardless of whether `Latexify.jl` is loaded. The same goes for any other functionality that can be factored out to not require explicit dependencies. I'm not a maintainer of this package, however, so take that with a grain of salt.

sostock:
If Latexify already uses `Requires` to add support for different packages, why not add Unitful support to Latexify instead of the other way round?

gustaphe:
> If Latexify already uses `Requires` to add support for different packages, why not add Unitful support to Latexify instead of the other way round?

I'm by no means an expert, but philosophically it makes sense to me the way it's described in Latexify's docs, that the recipe to latexify a type belongs with that type (just like with Plots recipes). From what I gather, the ones that are being imported there are from before `@latexrecipe` was implemented.

[There is a PR there too](https://github.com/korsbo/Latexify.jl/pull/129#issuecomment-740032845), so to me it's no big difference.

rafaqz:
Plots.jl has RecipesBase.jl to get around this problem. But even then, handling of Unitful.jl plots recipes is in another package.

If Latexify already has Requires.jl as a dependency it makes more sense to add this PR over there.

gustaphe:
[Interface package](https://github.com/gustaphe/UnitfulLatexify.jl) is on merge countdown in Registry. Thank you for the feedback :)

PR #407, opened 2020-12-20
Implement traits correctly

sostock:
The old implementation works correctly for `Quantity{T}` when `T` is a number type defined in `Base`, but not when `T` is defined elsewhere and has a custom implementation of these traits.


giordano:
Any test relevant for this change?

sostock:
> Any test relevant for this change?

I’ll add one.

sostock:
I added tests, utilizing the `Num` type already defined in the test suite.

Issue #135, opened 2018-04-26
hbar has wrong units.


MasonProtter:
I was (perhaps erroneously) working under the assumption that `hbar` in Unitful was reduced Plank's constant but that appears to be false. Instead, I get
```
julia> dimension(1u"hbar")
𝐋^-1 𝐌 𝐓^-2
```
which is off by a power of length cubed. Is `1u"hbar"` something other than what I thought?

ajkeller34:
Hectobars... an unfortunate consequence of supporting most SI prefixes. You can type `\hbar <tab>` in the REPL to autocomplete `ħ`.

PR #414, opened 2021-01-12
doc: Note on hbar/ħ

musoke:
Add note clarifying the difference between hbar and ħ.  Hopefully fewer people like me will have to go searching for #135


giordano:
Good idea, thanks!

musoke:
Thanks for the package!

PR #416, opened 2021-01-21
WIP: Add DifferenceQuantities

Keno:
These quantities represent differences between units.
At the moment, this is restricted to affine units, since
that is where this makes the most sense. However, in theory
this kind of thing could be used where-ever the scale is
non-absolute (e.g. angle differences) - though perhaps that
is best done by a AffineUnit with no offset for a particular
application.

This is by no means done, but I figured I'd put it up for
a discussion. I'm planning to play with this a bit and see
if I like it or not.

Demo:
```
julia> 1°C + 1Δ°C
2 °C

julia> 1°C + 1°C
ERROR: AffineError: an invalid operation was attempted with affine quantities: 1 °C + 1 °C
Stacktrace:
 [1] +(x::Quantity{Int64, 𝚯, Unitful.FreeUnits{(K,), 𝚯, Unitful.Affine{-5463//20}}}, y::Quantity{Int64, 𝚯, Unitful.FreeUnits{(K,), 𝚯, Unitful.Affine{-5463//20}}})
   @ Unitful ~/.julia/dev/Unitful/src/quantities.jl:184
 [2] top-level scope
   @ REPL[5]:1

julia> 5°C - 2°C
3 Δ°C
```

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -283,6 +283,9 @@ baremodule DefaultSymbols
     Core.eval(DefaultSymbols, Expr(:import, Expr(:(.), :Unitful, :°C)))
     Core.eval(DefaultSymbols, Expr(:export, :°C))
 
+    Core.eval(DefaultSymbols, Expr(:import, Expr(:(.), :Unitful, :Δ°C)))
+    Core.eval(DefaultSymbols, Expr(:export, :Δ°C))
+
     Core.eval(DefaultSymbols, Expr(:import, Expr(:(.), :Unitful, :°)))
     Core.eval(DefaultSymbols, Expr(:export, :°))
 end
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -154,6 +154,32 @@ function +(x::AffineQuantity{S,D}, y::AbstractQuantity{T,D}) where {S,T,D}
 end
 +(x::AbstractQuantity, y::AffineQuantity) = +(y,x)
 
+function +(x::AbstractQuantity{S, D, FreeUnits{U, B, A}},
+           y::AbstractQuantity{S, D, FreeUnits{U, B, Difference{A}}}) where {S, D, U, B, A}
+    Quantity(x.val + y.val, unit(x))
+end
+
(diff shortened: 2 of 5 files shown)


Keno:
Or maybe, we just define Kelvin to be the base unit of temperature difference rather than temperature, change the addition methods to prefer the affine unit, and maybe even add a Kelvin unit that is just an affine unit with zero base point. What I ran into with this approach is that I wanted `1/Δ°C`, but that doesn't work with this.

c42f:
Agreed, Kelvin is the natural temperature difference unit for °C and we already have

```
julia> 5°C - 2°C
3 K
```

I'm not sure all things work as expected for K and °C interop but if not, fixing those would make most sense to me.

If you want things to work for general affine units we'd need something more general though.

Keno:
Yes, in that interpretation the following is bad though:
```
5.0°C + 2K
280.15 K
```
because that conflates Kelvin-as-temperature and Kelvin-as-temperature difference. I think they need to be two separate types.

Keno:
Right, my only point was then that I don't think `°C` should implicitly convert to `K` at all, e.g. in:
```
julia> +(1.0°C, 1K, 1.0°C)
549.3 K
```

c42f:
Yes I don't mean to say this all works smoothly right now as things are.

If we had `+(1.0°C, 1K)` returning `°C` I think we'd be in fairly good shape in terms of usability and consistency?

(Side note - in the past I've thought a lot about encoding affine vs vector types with the type system (for geometry, not units), but I came to the conclusion that it might not be super helpful in Julia. It brings clear value in statically typed languages where you get a compile error. But in Julia it doesn't bring as much in terms of error checking — you still need unit tests to cover those types of errors.)

sostock:
That’s basically what `ContextUnits` are for, maybe we should advertise them better? Or just define `Δ°C = ContextUnits(K, °C)`?
```julia
julia> Δ°C = Unitful.ContextUnits(u"K", u"°C")
K

julia> 2u"°C" + 5Δ°C
7//1 °C
```

PR #417, opened 2021-01-21
Remove bad at-eval

Keno:
I don't know what this was supposed to do, but what it accomplishes in reality is to break pre-compilation,
so better remove it ;).


giordano:
Thanks!

Issue #330, opened 2020-05-14
No entropy/information types


freemin7:
I noticed that Unitful has no 'Bit' type. While there is no way to represent a Bit in SI units it does indeed interact with physical quantities, such as kbit/s or landauer's limit Energy/Bit. There also exists a rich ecosystem of other scales to represent information such as nit or dit depending on the application.

This proposals opens the doors to more problems though. What about Bauds (symbols/s)?

Can i hear your thoughts on the issue?

m-wells:
Seems rather straight forward to implement.
```julia
julia> using Unitful
       Unitful.register(@__MODULE__);
       @unit bit "bit" Bit 1 true;

julia> 1u"kbit/s"
1 kbit s^-1

julia> 2u"J/bit"
2 J bit^-1

julia> typeof(u"bit")
Unitful.FreeUnits{(bit,),NoDims,nothing}
```

m-wells:
This doesn't seem to fit in the base Unitful package but could potentially (if there was someone motivated to do so) be apart of a new package. There are already a number of extensions to Unitful (see [Units-packages](https://github.com/PainterQubits/Unitful.jl#units-packages)).

freemin7:
Oh, i wasn't aware that it would be that easy.

PR #419, opened 2021-01-31
Implement bit and byte for information and data

### LINKED ISSUE #173: Feature Request: Units of information
[Units of information](https://en.wikipedia.org/wiki/Units_of_information)
It was mentioned in [discourse](https://discourse.julialang.org/t/unitful-for-file-size/16423).

BrandonConder:
Create information. Define Byte as basic unit for data storage. Include data rates. Modify barn (b -> bn) and Bel (B -> bel) to accommodate the new definitions.

Closes #173.
Mostly addresses #330.
Does not include e.g. methods for converting between bits and entropy, this is probably reasonable since most users will think of bits as data (e.g. storage capacity, transfer rates) rather than information. 

Personal preferences:
- I don't think this belongs in a separate units package because bit and byte are common and conflict with the existing definitions of barn and Bel.
- Less common units are not included: nibble, 1024 multiples (kibibytes, mebibytes, etc.). A user could easily implement these.
- Baud not defined because bits per symbol is not standardized.

Quirks:
- I had to manually implement the SI prefixes for Bytes to avoid deci-bytes (dB), which is a meaningless unit and a conflict with Decibel.
- ZettaByte and YottaByte were not included because the literal 8_000_000_000_000_000_000_000 undergoes type promotion to Float64 and causes a rounding error.
    - It might be possible to implement this using Int128(8_000_000_000_000_000_000_000)
    - If there were a way to use the Unitful prefix system excluding deciByte, Unitful handles these types greacefully (e.g. `1Zm/Em == 1000` returns `true`)
    - Behavior with the 32-bit Julia binary needs review, to ensure no units promote to Float32

I've included tests for the manual SI prefixes.

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -224,15 +224,14 @@ const Å = Å = angstrom
 # Instead define all units in common use
 # Overflow warning:
 #  Gigascale and above likely exceed Int32
-#  Exascale and above likely exceed Int64
-@unit      kB   "kB"     KiloByte          8_000b      false
-@unit      MB   "MB"     MegaByte          8_000_000b     false
-@unit      GB   "GB"     GigaByte          8_000_000_000b     false
-@unit      TB   "TB"     TeraByte          8_000_000_000_000b    false
-@unit      PB   "PB"     PetaByte          8_000_000_000_000_000b    false
-@unit      EB   "EB"     ExaByte           8_000_000_000_000_000_000b    false
-@unit      ZB   "ZB"     ZettaByte         8_000_000_000_000_000_000_000b    false
-@unit      YB   "YB"     YottaByte         8_000_000_000_000_000_000_000_000b    false
+#  Exascale likely exceeds Int64
+# Omitting ZB and YB: Julia interprets literal as Float64, which causes rounding errors
+@unit      kB   "kB"       KiloByte          8_000b      false
+@unit      MB   "MB"       MegaByte          8_000_000b     false
+@unit      GB   "GB"       GigaByte          8_000_000_000b     false
+@unit      TB   "TB"       TeraByte          8_000_000_000_000b    false
+@unit      PB   "PB"       PetaByte          8_000_000_000_000_000b    false
+@unit      EB   "EB"       ExaByte           8_000_000_000_000_000_000b    false
 # Data rate
 @unit      bps  "bps"      BitPerSecond      1b/s         true
 @unit      Bps  "Bps"      BytePerSecond     8b/s         true
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -9,6 +9,7 @@
 @dimension 𝚯 "𝚯" Temperature    # This one is \bfTheta
 @dimension 𝐉 "𝐉" Luminosity
 @dimension 𝐍 "𝐍" Amount
+@dimension 𝐛 "𝐛" Information
 const RelativeScaleTemperature = Quantity{T, 𝚯, <:AffineUnits} where T
 const AbsoluteScaleTemperature = Quantity{T, 𝚯, <:ScalarUnits} where T
 
@@ -60,6 +61,7 @@ const AbsoluteScaleTemperature = Quantity{T, 𝚯, <:ScalarUnits} where T
 @refunit  cd      "cd"     Candela   𝐉            true
 @refunit  g       "g"      Gram      𝐌           true
 @refunit  mol     "mol"    Mole      𝐍           true
+@refunit  b       "b"      Bit       𝐛           false
 
 # Angles and solid angles
 @unit sr      "sr"      Steradian   1                       true
@@ -113,7 +115,7 @@ end
 # The hectare is used more frequently than any other power-of-ten of an are.
 @unit a      "a"        Are         100m^2                  false
 const ha = Unitful.FreeUnits{(Unitful.Unit{:Are, 𝐋^2}(2, 1//1),), 𝐋^2}()
-@unit b      "b"        Barn        100fm^2                 true
+@unit bn     "bn"       Barn        100fm^2                 true
 
 # Volume
 # `l` is also an acceptable symbol for liters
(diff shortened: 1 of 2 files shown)


BrandonConder:
It looks like the Float32 problem is realized on x86. How to proceed? I can leave out SI prefixes and let users decide if they want to declare kB, MB, GB, etc.

Alternatively I could try to rewrite the @unit macro such that the input `tf` can accept a flag for "prefixes > 1 only", but that seems like a pretty big overhaul for this edge case.

src/pkgdefaults.jl
```diff
@@ -113,7 +115,7 @@ end
 # The hectare is used more frequently than any other power-of-ten of an are.
 @unit a      "a"        Are         100m^2                  false
 const ha = Unitful.FreeUnits{(Unitful.Unit{:Are, 𝐋^2}(2, 1//1),), 𝐋^2}()
-@unit b      "b"        Barn        100fm^2                 true
+@unit bn     "bn"       Barn        100fm^2                 true
```
giordano, line 118: This is breaking

src/pkgdefaults.jl
```diff
@@ -216,11 +218,29 @@ const Å = Å = angstrom
...
+# Data rate
+@unit      bps  "bps"      BitPerSecond      1b/s         true
+@unit      Bps  "Bps"      BytePerSecond     8b/s         true
+
 #########
 # Logarithmic scales and units
 
 @logscale dB    "dB"       Decibel      10      10      false
-@logscale B     "B"        Bel          10      1       false
+@logscale bel   "Bel"      Bel          10      1       false
```
giordano, line 243: And this one as well

BrandonConder:
Is backward-compatibility-breaking prohibited? In my field Bel is never used (only decibel), and I don't know how common or significant Barn is, so I assumed bit and Byte were more common definitions of those abbreviations.

giordano:
We can have a breaking change, but need to do a breaking release (i.e., go to v2.0), which would cause some friction in the ecosystem of packages relying on Unitful, for little benefit

sostock:
I would suggest to use `Byte` and `bit` instead of `B` and `b`, that way it isn’t breaking.

src/pkgdefaults.jl
```diff
@@ -9,6 +9,7 @@
 @dimension 𝚯 "𝚯" Temperature    # This one is \bfTheta
 @dimension 𝐉 "𝐉" Luminosity
 @dimension 𝐍 "𝐍" Amount
+@dimension 𝐛 "𝐛" Information
```
mcabbott, line 12: Should this be a dimension? I thought that `nat` is, like `rad`, a pure number -- it's just some `p*log(p)` where `p` is a probability. And bits, like degrees etc, are multiples of that.

src/pkgdefaults.jl
```diff
@@ -216,11 +218,29 @@ const Å = Å = angstrom
 # Pressure
 @unit psi       "psi"      PoundsPerSquareInch  1lbf/inch^2             false
 
+# Data
+@unit      B    "B"        Byte              8b          false
```
mcabbott, line 222: If the definitions were like this, would it still convert MB to GB as integers, without introducing the `log(2)`?
```suggestion
# Information
@unit    nat    "nat"      Nat                1            false
@unit    bit    "bit"      Bit          log(2)*u"nat"      false
@unit      B    "B"        Byte              8bit          false
```

src/pkgdefaults.jl
```diff
@@ -216,11 +218,29 @@ const Å = Å = angstrom
...
+# deciBytes can be a problem with deciBels, and we don't actually care about orders < 1
+# Instead define all units in common use
+# Overflow warning:
+#  Gigascale and above likely exceed Int32
+#  Exascale likely exceeds Int64
+# Omitting ZB and YB: Julia interprets literal as Float64, which causes rounding errors
+@unit      kB   "kB"       KiloByte          8_000b      false
+@unit      MB   "MB"       MegaByte          8_000_000b     false
+@unit      GB   "GB"       GigaByte          8_000_000_000b     false
```
mcabbott, line 231:
```suggestion
@unit      kB   "kB"       KiloByte          8_000b      false
@unit      kiB   "kiB"       KibiByte          8_192b      false
@unit      MB   "MB"       MegaByte          8_000_000b     false
@unit      MiB   "MiB"       MebiByte          8_388_608b     false
@unit      GB   "GiB"       GigaByte          8_000_000_000b     false
@unit      GiB   "GiB"       GibiByte          8_589_934_592b     false
```

rico-cl:
Regarding the question of adding another dimension for the desired units, I propose to go with dimensionless units, since
* Unitful seem to try and adhere to the SI which doesn't provide such a dimension 
* how to consistently handle units that are dimensionless in the framework of SI as of now is still an unsolved (or at least debated) problem in the respective research communities as far as I understand, so I wouldn't expect Unitful to come up with a solution itself
* the units in question can be used to measure different things: amount of data, amount of storage, informational content of data (as denoted by Shannon entropy) - so even if one would like to assign a dimension to them, Information might not even be the most meaningful choice

src/pkgdefaults.jl
adkabo: `"GiB"` appears twice in the suggested change, looks like a typo.

mcabbott:
I was going to say that these points all apply to angles, where radians are also pure numbers. But it turns out they have in fact been blessed by SI.

Nevertheless, it is not an accident that storage and information are measured in the same units. There are mistakes you can make which won't be caught by checking the units, but there are also such mistakes when measuring both height and distance in metres.

filchristou:
I am in need of such a unit in order to handle data rates in networking problems.
I think it's worth having a specific unit for this case because there is a big confusion in my field when someone talks about `Mbits/sec` or `MBytes/sec`. A package that handles this transparently will be great.

singularitti:
Hope it's not off-topic: there is a relevant package https://github.com/uriele/UnitfulData.jl

Issue #424, opened 2021-02-15
Add consistent CGS unit system


Gregstrq:
First of all, I would like to thank the creators of the package. It is very powerful: one can add all sorts of units by expressing them in terms of SI units. This way, some CGS units were already added in issue #395.

However, I should point out that varying units is not the only difference between the two unit systems. They also differ in how the base dimensions and derived dimensions are defined. (Of course, it is true only for electromagnetic quantities.) So, if one plugs CGS units into CGS equations, but CGS units are expressed in terms of SI units, the dimension of the answer is nonsensical.

So, in order to work with CGS units consistently, one needs to correctly define base and derived dimensions. The good news is, as far as I understand, one does not need to rewrite everything to incorporate such changes; one only needs to rewrite the file [pkgdefaults.jl](https://github.com/PainterQubits/Unitful.jl/blob/master/src/pkgdefaults.jl).

Since the form of equations involving electromagnetic quantities are different in SI and CGS systems, one never mixes both types of units in the calculation: the quantities are first converted to a single preferred system, and only then are plugged into equations.
The conversion part should be easily handled by the units definitions (we can define CGS units in terms of SI units, but we can also do the reverse thing). Because of this, there is, actually, no point to bother too much about the transformations of quantities between the two unit systems. A user would choose the unit system when the package is loaded, and then he/she would stick with it.

Gregstrq:
The way I see it, one just needs to add an alternative pkgdefaults file and provide an interface to choose the needed one.
What do you think about this idea?

I am ready to write an alternative pkgdefaults file myself, however, the interface part needs discussion.
It seems that it may be a good idea to look closely into issue #359. There, it is suggested to separate everything besides the [pkgdefaults.jl](https://github.com/PainterQubits/Unitful.jl/blob/master/src/pkgdefaults.jl) file into UnitfulBase.jl, and then have separate packages with pkgdefaults.

rafaqz:
This sounds reasonable to me. UnitfulBase.jl and Unitful.jl could both be in this repository, minimizing the maintenance overhead, sharing docs, etc. UnitfulBase.jl would be a subdirectory.

Gregstrq:
> This sounds reasonable to me. UnitfulBase.jl and Unitful.jl could both be in this repository, minimizing the maintenance overhead, sharing docs, etc. UnitfulBase.jl would be a subdirectory.

If UnitfulBase is a subdirectory, is it possible to import only it without importing the whole Unitful?

[...]

Issue #359, opened 2020-06-16
More modular structure


aplavin:
There are many units defined by `Unitful` itself, and as I see it gives no way to disable them. Some of the default units have very short generic names, even though these units themselves are mostly used in relatively specific areas. Some examples are `u"G"` (gravity constant), `u"R"` (molar gas constant), `u"P"` (poise), `u"B"` (bel) and so on. Personally I often wish I could make `u"G"` to mean Gauss instead.

I think it would be more convenient to have something like "UnitfulBase.jl" as a separate package that sets up the infrastructure and defines no units. Then "Unitful" itself would use that package and basically consist of a single file `pkgdefaults.jl`, and other packages would have complete freedom with defining units that suit any possible domain.
Does this look reasonable to you?

cstjean:
> I think it would be more convenient to have something like "UnitfulBase.jl" as a separate package that sets up the infrastructure but defines no units. Then "Unitful" itself would use that package and basically consist of a single file pkgdefaults.jl, and other packages would have complete freedom to define units that suit any possible domain.
Does this seem reasonable to you?

One problem with this is that even if you don't want `u"G"` to be defined, you may import some astronomy package that will `import Unitful`. No? The `u` syntax is unfortunately a kind of global namespace.

cstjean:
I think it should be possible to make `u` module-specific. `__module__` is also available inside of string macros, so one could envision something like

````julia
using Unitful
@use_cgs_units()    # for this module, have `u""` use CGS units instead.
``` 

You'd need a `module_specific_u_namespace = Dict()` in Unitful.

PR #428, opened 2021-03-06
WIP: split core functionality into UnitfulBase

Gregstrq:
This PR adresses issue #424 and issue #359.

- I have moved all the files except "pkgdefaults.jl" and "dates.jl" into a subdirectory package UnitfulBase.
- I also moved the definitions of radian, steradian and degrees into UnitfulBase. The reason for that was [the extension of the function `mod2pi`](https://github.com/PainterQubits/Unitful.jl/blob/e52c19243e264f77c13fcf23d612df1cfad42f0b/src/quantities.jl#L125) which uses degrees. I have decided to move degrees into UnitfulBase instead of keeping `mod2pi` extension in Unitful. Since radian and steradian are dimensionless and thematically connected to degrees, I have moved them as well.
- Unitful is now a registered extension of UnitfulBase.

- I have edited the tests so they work with the splitted UnitfulBase. I have run the tests on my computer: they run successfully with the exception of the single test for issue #272 (see [these lines](https://github.com/PainterQubits/Unitful.jl/blob/ca230d46155d9d893a62b8afec403bd146748ba9/test/runtests.jl#L1725-L1730)). The problem is that if @u_str is imported without at least importing Unitful, Unitful is not registered as an extension of UnitfulBase.

- Since UnitfulBase is not a registered package yet, the only way to add it to Unitful project environment was to `dev` it. The information about the path to UnitfulBase was stored only in Manifest.toml, which I did not commit. As a result, the tests might be broken in CI for now. If this PR will ever get merged, the problem will disappear once UnitfulBase is registered as a package.

I did not touch the docs yet. I would like to hear your opinion about the contents of this PR first.


Gregstrq:
I understand that the idea of the splitting of the core functionality is an arguable thing.

I created this PR as an experiment to poke around and to see what we get.
That said, I am eager to work on any feedback from the maintainers.
(May be, a good approach is to create an experimental branch and push this PR there instead of master.)

Issue #46, opened 2016-11-10
Unitful linear algebra


ChrisRackauckas:
Some linear algebra operations like `\` have a hard time with unitful numbers. Take the following case:

```julia
a = [1.0u"N" 2.0u"N"
      3.0u"N" 1.0u"N"]
b = [1.0u"N";3.0u"N"]
```

If you try `\`  it gives a dimension error. If you try `*` then it uses the generic `*` fallback and not BLAS.

I suggest trying something like, performing a dimensional analysis, strip the units, use the general `*` or `\`, and then re-apply units. 

To do this properly, you might need a `UnitfulArray` instead of an array of Unitful values. I am not sure if stripping the units on an array of Unitful numbers can be done without making a temporary.

giordano:
Indeed the algorithm used in `generic_lufact!` doesn't look very physical-units-friendly:

```julia
Akkinv = inv(A[k,k])
for i = k+1:m
    A[i,k] *= Akkinv
end
```

Other types for which `T` and `T*inv(T)` don't have the same type would fail at this point.  As a physicist, I expect (or hope) that a function performs dimensionally consistent computations.  It may be worth trying looking into the upstream factorization method.

briochemc:
Is there a plan to eventually resolve this issue or was this given up on?

cstjean:
Depending on your use case, https://github.com/PainterQubits/Unitful.jl/pull/191 might work? I don't know if I'll ever find the time to finish it, though.

briochemc:
Thanks for the suggestion! Unfortunately it does not really help me because my use case involves large sparse LU factorizations (`factorize`, `ldiv!`, etc.). I also outlined my own dirty workaround in https://github.com/PainterQubits/Unitful.jl/issues/150, but I'm not sure it is a good one at all (no one  commented on it). I realize discussion on this has been going for a while (e.g., https://github.com/JuliaLang/julia/pull/20484 and https://github.com/JuliaLang/julia/issues/7623), but I am unaware of any progress regarding the issue here, so I wanted to revive this issue a bit! 🙏

timholy:
The idea of stripping the units seems a bit limiting. For example this matrix has an inverse:
```julia
julia> using Unitful: mm, s, kg

julia> A = [2mm^2 1mm*s; 1mm*s 2s^2]
2×2 Array{Quantity{Int64,D,U} where U where D,2}:
 2 mm^2  1 mm s
 1 mm s   2 s^2
```
[...]

PR #437, opened 2021-04-13
Some linear algebra, via `reinterpret`

mcabbott:
This is a start on #46.

### DIFF
--- a/src/linearalgebra.jl
+++ b/src/linearalgebra.jl
@@ -1,0 +1,42 @@
+using LinearAlgebra
+
+# This function is re-defined during testing, to check we hit the fast path:
+linearalgebra_count() = nothing
+
+function LinearAlgebra.mul!(C::StridedVecOrMat{<:AbstractQuantity{T}}, 
+        A::StridedMatrix{<:AbstractQuantity{T}}, 
+        B::StridedVecOrMat{<:AbstractQuantity{T}},
+        alpha::Bool, beta::Bool) where {T<:Base.HWNumber}
+    # This is exactly how A * B creates C = similar(B, T, ...)
+    eltype(C) == Base.promote_op(LinearAlgebra.matprod, eltype(A), eltype(B)) || error("bad eltypes")
+    C0 = ustrip(C)
+    A0 = ustrip(A)
+    B0 = ustrip(B)
+    mul!(C0, A0, B0)
+    linearalgebra_count()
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -41,6 +41,16 @@ using Dates:
 
 const colon = Base.:(:)
 
+ambig = sort(detect_ambiguities(Unitful), by = a -> [string(a[1].name), string(a[2].module)])
+if length(ambig) > 0
+    println(stdout, "detect_ambiguities(Unitful) found $(length(ambig)) issues:")
```
mcabbott, line 46: I was concerned about creating ambiguities, so made it print them before starting. 

In the end, overloading `mul!` not `*` seems not to create problems. But there are (for me) 30 other ambiguities this detects (and prints out).

src/linearalgebra.jl
```diff
@@ -0,0 +1,42 @@
+using LinearAlgebra
+
+# This function is re-defined during testing, to check we hit the fast path:
+linearalgebra_count() = nothing
+
+function LinearAlgebra.mul!(C::StridedVecOrMat{<:AbstractQuantity{T}}, 
+        A::StridedMatrix{<:AbstractQuantity{T}}, 
+        B::StridedVecOrMat{<:AbstractQuantity{T}},
+        alpha::Bool, beta::Bool) where {T<:Base.HWNumber}
```
mcabbott, line 9: The case `α, β::Bool` is what `A * B` produces. Allowing other pure numbers ought to be fine. Allowing these to have units I haven't looked into.

src/linearalgebra.jl
```diff
@@ -0,0 +1,84 @@
...
+    Tu = typeof(one(eltype(C0)) * u)
+    return reinterpret(Tu, C0)
+end
+
+function inv(A::StridedMatrix{<:AbstractQuantity{T}}) where {T<:Base.HWNumber}
+    C0 = inv(ustrip(A))
+    _linearalgebra_count()
+    u = inv(unit(eltype(A)))
+    Tu = typeof(one(eltype(C0)) * u)
+    return reinterpret(Tu, C0)
```
mcabbott, line 71: Note that these methods will return a `ReinterpretArray{Quantity{...` not a `Matrix`. There's no obvious analogue the `mul!` overloaded for `A * B`.

PR #439, opened 2021-04-17
Test for ambiguities?

mcabbott:
This runs `detect_ambiguities`, but doesn't (yet) demand the result be empty. 

Here's the list it prints out: https://github.com/PainterQubits/Unitful.jl/pull/439/checks?check_run_id=2367869650#step:6:35

### DIFF
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -191,6 +191,7 @@ for (_x,_y) in [(:fma, :_fma), (:muladd, :_muladd)]
     # Catch some signatures pre-promotion
     @eval @inline ($_x)(x::Number, y::AbstractQuantity, z::AbstractQuantity) = ($_y)(x,y,z)
     @eval @inline ($_x)(x::AbstractQuantity, y::Number, z::AbstractQuantity) = ($_y)(x,y,z)
+    @eval @inline ($_x)(x::AbstractQuantity, y::AbstractQuantity, z::AbstractQuantity) = ($_y)(x,y,z)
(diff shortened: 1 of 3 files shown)


giordano:
Interesting.  I think it makes sense to have this test if it also enforces that there are no ambiguities, just printing them doesn't encourage _too_ much to fix them :slightly_smiling_face:

mcabbott:
I fixed a few, not sure I understand the package well enough to fix all of them, nor to know whether they can ever matter in practice. 

So maybe this is more an issue than a PR for now. It could perhaps test `length(detect_ambiguities(Unitful) <= 25` which would at least make it harder to introduce new ones.

src/quantities.jl
```diff
@@ -191,6 +191,7 @@ for (_x,_y) in [(:fma, :_fma), (:muladd, :_muladd)]
     # Catch some signatures pre-promotion
     @eval @inline ($_x)(x::Number, y::AbstractQuantity, z::AbstractQuantity) = ($_y)(x,y,z)
     @eval @inline ($_x)(x::AbstractQuantity, y::Number, z::AbstractQuantity) = ($_y)(x,y,z)
+    @eval @inline ($_x)(x::AbstractQuantity, y::AbstractQuantity, z::AbstractQuantity) = ($_y)(x,y,z)
```
sostock, line 194: Adding this method means that the one below it can be deleted (it’s the same but with unused type parameters).

PR #445, opened 2021-05-19
To certain functions, add the `fancy_exponents` kwarg, which allows the user to toggle the "fancy exponents" setting in a thread-safe way

DilumAluthge:
This pull request adds the `fancy_exponents::Bool` kwarg to several methods of the following functions:
- `show`
- `showval`
- `showrep`
- `superscript`

We use the fancy Unicode exponents if and only if `fancy_exponents` is `true`.

The `fancy_exponents::Bool` kwarg is optional. The default value is determined as follows:
1. If the user explicitly provides the `fancy_exponents` kwarg, we use that value.
2. Otherwise, if the user has set the `UNITFUL_FANCY_EXPONENTS` environment, and the lowercased value parses to a `Bool`, we use that value.
3. Otherwise, if the user has set the `UNITFUL_FANCY_EXPONENTS` environment variable, but the lowercased value does not parse to a `Bool`, we default to `false`.
3. Otherwise, we default to `true` on macOS and `false` otherwise.

Therefore, if the user does not provide the `fancy_exponents` kwarg, this PR preserves the existing behavior.

# Motivation

Currently, the only way to control the behavior of the fancy Unicode exponents is by the `UNITFUL_FANCY_EXPONENTS` environment variable. Unfortunately, as far as I understand, there is no thread-safe way to modify the environment while calling a Julia function. (The `withenv` function is not thread-safe. The `addenv` function is thread-safe, but as far as I can tell, it only lets you run a `Cmd`.)

The `fancy_exponents` kwarg allows the user to turn fancy Unicode exponents on/off in a thread-safe way.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Unitful"
 uuid = "1986cc42-f94f-5a68-af5c-568840ba703d"
-version = "1.7.0"
+version = "1.8.0"
 
 [deps]
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
--- a/src/display.jl
+++ b/src/display.jl
@@ -23,6 +23,15 @@ const prefixdict = Dict(
     24  => "Y"
 )
 
+const docstring_for_fancy_exponents_kwarg = "- `fancy_exponents::Bool`. By default, this is `true` on macOS and false otherwise."
+
+function get_fancy_exponents_env()
+    v = get(ENV, "UNITFUL_FANCY_EXPONENTS", Sys.isapple() ? "true" : "false")
+    t = tryparse(Bool, lowercase(v))
+    fancy_exponents = (t === nothing) ? false : t
+    return fancy_exponents
+end
+
 """
 `abbr(x)` provides abbreviations for units or dimensions. Since a method should
 always be defined for each unit and dimension type, absence of a method for a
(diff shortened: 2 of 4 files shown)


giordano:
I can't say I'm a fan of using environment variables for this. Why not using IO context instead?

DilumAluthge:
Sure, that would work. I just didn't want to break the existing behavior of the package.

DilumAluthge:
I.e. the package already uses the environment variable.

giordano:
How using IO context would be different from using an environment variable?

DilumAluthge:
Ah I think I see what you mean.

DilumAluthge:
Well just to clarify, we'd use IO contexts instead of the kwarg. The `UNITFUL_FANCY_EXPONENTS` environment still has to stay as a fallback.

giordano:
Yes, the extra argument is redundant since you can pass extra stuff already from the IO. I use IO context in Measurements.jl, see for example https://github.com/JuliaPhysics/Measurements.jl/blob/master/src/show.jl#L42 and the tests

DilumAluthge:
> Yes, the extra argument is redundant since you can pass extra stuff already from the IO. I use IO context in Measurements.jl, see for example https://github.com/JuliaPhysics/Measurements.jl/blob/master/src/show.jl#L42 and the tests

It's so much cleaner! I like it.

Closing this PR in favor of https://github.com/PainterQubits/Unitful.jl/pull/446, which uses the IO context property.

PR #446, opened 2021-05-20
Use the `:fancy_exponent` IO context property to override the behavior of fancy exponents

DilumAluthge:
Before this PR, this is how we determine the behavior of fancy exponents:
1. If the user has set the `UNITFUL_FANCY_EXPONENTS` environment variable, and the lowercased value parses to a `Bool`, we use that value.
2. Otherwise, if the user has set the `UNITFUL_FANCY_EXPONENTS` environment variable, but the lowercased value does not parse to a `Bool`, we default to false.
3. Otherwise, we default to true on macOS and false otherwise.

After this PR, this is how we determine the behavior of fancy exponents:
1. **If the IO has the `:fancy_exponent` IO context property, and the value of `:fancy_exponent` is a `Bool`, we use that value.**
2. If the user has set the `UNITFUL_FANCY_EXPONENTS` environment variable, and the lowercased value parses to a `Bool`, we use that value.
3. Otherwise, if the user has set the `UNITFUL_FANCY_EXPONENTS` environment variable, but the lowercased value does not parse to a `Bool`, we default to false.
4. Otherwise, we default to true on macOS and false otherwise.

Example usage:

```julia
julia> sprint(io -> show(IOContext(io, :fancy_exponent => true), u"m/s"))
"m s⁻¹"

julia> sprint(io -> show(IOContext(io, :fancy_exponent => false), u"m/s"))
"m s^-1"
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Unitful"
 uuid = "1986cc42-f94f-5a68-af5c-568840ba703d"
-version = "1.7.0"
+version = "1.8.0"
 
 [deps]
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
--- a/docs/src/display.md
+++ b/docs/src/display.md
@@ -3,7 +3,8 @@
 By default, exponents on units or dimensions are indicated using Unicode superscripts on
 macOS and without superscripts on other operating systems. You can set the environment
 variable `UNITFUL_FANCY_EXPONENTS` to either `true` or `false` to force using or not using
-the exponents.
+the exponents. You can also set the `:fancy_exponent` IO context property to either `true`
+or `false` to force using or not using the exponents.
 
 ```@docs
 Unitful.BracketStyle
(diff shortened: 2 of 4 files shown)


DilumAluthge:
@giordano If you want, I can completely remove the environment variable functionality, and bump the version number to `2.0.0`.

DilumAluthge:
Bump @giordano.

Can you take a look at this and let me know if you want to keep the environment variables stuff or just remove it entirely?

giordano:
I doubt anyone will be happy about making a breaking change just for removing an environment variable which is barely used by anyone.

DilumAluthge:
> I doubt anyone will be happy about making a breaking change just for removing an environment variable which is barely used by anyone.

Yeah that was my thinking.

DilumAluthge:
Does this PR look good?

DilumAluthge:
Bump @giordano

DilumAluthge:
Thanks! Now that we have this, perhaps we could deprecate the environment variable? #447

thchr:
An unhappy consequence of the macOS special-casing is that it is difficult to unit-test the output of things that print something from Unitful over multiple OSes. 

With Windows terminal being widespread these days, and its better unicode rendering, maybe this should also be true for Windows?

PR #447, opened 2021-05-21
Deprecate the `UNITFUL_FANCY_EXPONENTS` environment variable in favor of the `:fancy_exponent` IO context property

DilumAluthge:
Adding a deprecation warning can be done in a non-breaking release.

I didn't actually use `depwarn` for this, because those kinds of deprecation warnings are turned off by default. So instead, I used `@warn` with `maxlog=1`, which ensures that this warning will be shown by default.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Unitful"
 uuid = "1986cc42-f94f-5a68-af5c-568840ba703d"
-version = "1.8.0"
+version = "1.8.1"
 
 [deps]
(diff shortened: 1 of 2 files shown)


DilumAluthge:
It would probably be best to first merge #446. Then we can update the deprecation warning to say "please use the `:fancy_exponent` IO context property instead".

DilumAluthge:
Bump @giordano

DilumAluthge:
Bump @giordano

giordano:
I honestly don't have a strong opinion about deprecating this

PR #448, opened 2021-05-22
Introduce `ubasestrip` for easy consistency with SI base units

j-fu:
The purpose is to allow users to define unitless constants like `mm`,`nm` etc
and to allow consistent calculations with them, and to provide values
of physical constants in a consistent manner.

### DIFF
--- a/docs/src/manipulations.md
+++ b/docs/src/manipulations.md
@@ -33,6 +33,12 @@ Unitful.dimension
 Unitful.ustrip
 ```
 
+## Unit stripping from base units
+
+```@docs
+Unitful.ubasestrip
+```
+
 ## Unit multiplication
 
 ```@docs
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -27,7 +27,7 @@ import Random
 
 import ConstructionBase: constructorof
 
-export logunit, unit, absoluteunit, dimension, uconvert, ustrip, upreferred
+export logunit, unit, absoluteunit, dimension, uconvert, ustrip, ubasestrip, upreferred
 export @dimension, @derived_dimension, @refunit, @unit, @affineunit, @u_str
 export Quantity, DimensionlessQuantity, NoUnits, NoDims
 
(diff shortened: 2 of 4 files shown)


sostock:
Personally, I think `ustrip(upreferred(q))` is not very complicated and makes it explicit what is happening, so I don’t see a reason to add a new function for it. Of course, that only works for quantities, not for units and strings. But I don’t think we should treat units like quantities anyway. But that’s only my opinion.

If we decide to add this, I would suggest to rename the function to `upreferredstrip` to make it explicit that it converts to `upreferred` (the term “base unit” might not be so clearly defined for dimensions that are not part of SI). I also would remove the `String` method, since no other functions except for `uparse` operate on strings. I think it is better to require the user to explicitly `uparse` the string.

j-fu:
`upreferredstrip` would be ok for  me. Anyway I am not intending to  argue with the policy of the package.

I am working on finding a consistent pattern with Unitful for the use case in PDE codes where we sooner or later are involved with unitless packages (e.g. sparse direct solvers, mesh generators) . I would like to recommend this pattern as a policy for several packages we are developing.  Unitful then would just serve
as a consistent database....

Essentially this would amount to:

```
const nm=upreferredstrip(u"nm")
const mA=upreferredstrip(u"mA)
const k_B=upreferredstrip(BoltzmannConstant)
...
current=20mA
... do all  calculations in numerical values of preferred units
println("current = $(current/mA) mA")
```
The last line here calls for something like
```
upreferredconvert(u::Unitful.Unitlike,v::Number)=uconvert(u,v*upreferred(u))
```
as the "inverse" to `upreferredstrip` so we could write
```
println("current = $(upreferredconvert(u"nA",current))")
```
etc

j-fu:
I put the discussion (and another API idea) on the zulip: https://julialang.zulipchat.com/#narrow/stream/287930-Unitful  Let us decide about the PR  after we see where this goes...

j-fu:
Close this for the time being -  I'll try to grasp more of the unitful logic...

PR #451, opened 2021-05-29
Showing how to manually cancel units

KronosTheLate:
The doc-section on Unit cancellation currently only states that units with differing SI-prefixes don't get cancelled. In this PR, I have added documentation for ways to cancel the units manually, which I have found very useful.

### DIFF
--- a/docs/src/conversion.md
+++ b/docs/src/conversion.md
@@ -300,19 +300,19 @@ in unit cancellation. For instance, `mV/V` is not simplified, although `V/V` is.
 Also, `N*m/J` is not simplified: there is currently no logic to decide
 whether or not units on a dimensionless quantity seem "intentional" or not.
 It is however possible to cancel units manually. In the equivalent examples below,
-it is shown how to cancel the units of 1km/1m:
+it is shown how to cancel the units of 1km/2.5m:
 ```jldoctest
 julia> using Unitful, Unitful.DefaultSymbols
 
 # Converting the units to be equal with `uconvert` for automatic canceling:
-julia> uconvert(m, 1km) / 1m
-1000.0
+julia> uconvert(m, 1km) / 2.5m
+400.0
 
 # Using the unit as a function for conversion instead of `uconvert`
-julia> m(1km) / 1m
-1000.0
+julia> m(1km) / 2.5m
+400.0
 
-# Using the piping operator to convert to the NoUnit type
-julia>  1km/1m |> NoUnit
-1000.0
+# Using the piping operator to convert to the NoUnits type
+julia>  1km/2.5m |> NoUnits
+400.0
 ```
--- a/docs/src/conversion.md
+++ b/docs/src/conversion.md
@@ -300,10 +300,10 @@ in unit cancellation. For instance, `mV/V` is not simplified, although `V/V` is.
 Also, `N*m/J` is not simplified: there is currently no logic to decide
 whether or not units on a dimensionless quantity seem "intentional" or not.
 It is however possible to cancel units manually, by passing the dimentionless
-size to the `NoUnits` constructor. This takes care of different SI-prefixes:
+size to the [`NoUnits`](@ref) constructor. This takes into account different SI-prefixes:
 ```jldoctest
 julia> using Unitful
 
-julia>  1u"km"/4u"m" |> NoUnits
+julia>  julia> 1u"kN*m"/4u"J" |> NoUnits
 250.0
 ```
--- a/docs/src/conversion.md
+++ b/docs/src/conversion.md
@@ -312,7 +312,7 @@ julia> uconvert(m, 1km) / 1m
 julia> m(1km) / 1m
 1000.0
 
-# Using the piping operator to convert to meter/meter:
+# Using the piping operator to convert to m/m, which is automatically canceled:
 julia>  1km/1m |> m/m
 1000.0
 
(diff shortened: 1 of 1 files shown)


docs/src/conversion.md
```diff
@@ -299,3 +299,24 @@ For multiplication and division, note that powers-of-ten prefixes are significan
 in unit cancellation. For instance, `mV/V` is not simplified, although `V/V` is.
 Also, `N*m/J` is not simplified: there is currently no logic to decide
 whether or not units on a dimensionless quantity seem "intentional" or not.
+It is however possible to cancel units manually. In the hour examples below,
+it is shown how to cancel the units of 1km/1m:
+```jldoctest
+julia> using Unitful, Unitful.DefaultSymbols
+
+# Converting the units to be equal with `uconvert`:
+julia> uconvert(m, 1km) / 1m
```
cstjean, line 308: At that point, wouldn't you rather write `ustrip(1km, m)`?

docs/src/conversion.md
```diff
@@ -299,3 +299,24 @@ For multiplication and division, note that powers-of-ten prefixes are significan
...
+julia> m(1km) / 1m
+1000.0
+
+# Using the piping operator to convert to m/m, which is automatically canceled:
+julia>  1km/1m |> m/m
+1000.0
+
+# The units are arbitrary, since the quantity is dimensionless:
+julia>  1km/1m |> J/J
+1000.0
```
giordano, line 321: This looks rather contrived, why not using
```julia
julia> 1km / 1m |> NoUnits
1000.0
```

docs/src/conversion.md
KronosTheLate: Because the output would be
```julia
julia> ustrip(m, 1km)/1m
1000.0 m^-1
```
, whereas the intended output was to cancel the `km` with the `m`. I see that this could have been made more clear by dividing with e.g. `2cm`.

docs/src/conversion.md
KronosTheLate: It is indeed quite contrived, I simply thought that it was a cool special case, and also felt that it hightlighted that you don't have to convert to e.g. `(m*V/A/s)/(m*V/A/s)` if that is your unit and you want to cancel it - it showcases that the unit is arbitrary when both sizes are dimensionless.

I did however not know about the `NoUnits` type, and that it would do the correct conversion when passed a dimensionless quantity with differing SI prefix. I agree that this is a better and more clear example.

I have a personal preferance to keep this example simply to showcase the cool special case, and add `1km / 1m |> NoUnits`. But I am also fine with replacing the `J/J` with `NoUnits` if you feel it is better ^_^

giordano: Yeah, I don't think it's a good idea to use `J/J` (or `m/m` above): it makes code hard to understand and doesn't use the more idiomatic, simple and correct `NoUnits`.  Please do the same also above

cstjean:
As an aside, why do you care about unit cancelation? In my code, I either use `ustrip` when I have to pass a number to a function that can't deal with units, or I leave the units be...

KronosTheLate:
> As an aside, why do you care about unit cancelation? In my code, I either use `ustrip` when I have to pass a number to a function that can't deal with units, or I leave the units be...

For the case that made me create this PR, was printing - I foundit annoying to get `1km/m` when I expect 1000.

docs/src/conversion.md
cstjean: I meant just `ustrip(m, 1km)`, which is equivalent to `uconvert(m, 1km) |> NoUnit`.

KronosTheLate: Oh, right. Because the point of the `/1m` is not to remove the unit, it is part of the expression. I'll make this more clear in the example by changing it to `2.5m`

KronosTheLate:
I sense that I think this is more important than you who have reviewed it think. I also see that including both of the first two examples seems more like documenting the fact that you can use units as functions for conversion. Maybe the example should be changed to only the `NoUnits` one, to make this section more concise?

docs/src/conversion.md
```diff
@@ -299,3 +299,11 @@ For multiplication and division, note that powers-of-ten prefixes are significan
 in unit cancellation. For instance, `mV/V` is not simplified, although `V/V` is.
 Also, `N*m/J` is not simplified: there is currently no logic to decide
 whether or not units on a dimensionless quantity seem "intentional" or not.
+It is however possible to cancel units manually, by passing the dimentionless
+size to the `NoUnits` constructor. This takes care of different SI-prefixes:
```
sostock, line 303:
```suggestion
quantity to the `NoUnits` constructor. This takes care of different SI-prefixes:
```

docs/src/conversion.md
```diff
@@ -299,3 +299,11 @@ For multiplication and division, note that powers-of-ten prefixes are significan
 in unit cancellation. For instance, `mV/V` is not simplified, although `V/V` is.
 Also, `N*m/J` is not simplified: there is currently no logic to decide
 whether or not units on a dimensionless quantity seem "intentional" or not.
+It is however possible to cancel units manually, by passing the dimentionless
```
sostock, line 302:
```suggestion
It is however possible to cancel units manually, by passing the dimensionless
```

docs/src/conversion.md
KronosTheLate: Accepting the "dimensionless" correcting, the size -> quantity one got outdated. I make a new commit, making the change.

cstjean (approved):

PR #460, opened 2021-07-07
A tip for specifying precision of displayed quantities.

mateuszbaran:
I've spent about half an hour figuring it out so I think this tip may be useful for other people :slightly_smiling_face: .

### DIFF
--- a/docs/src/display.md
+++ b/docs/src/display.md
@@ -6,6 +6,20 @@ variable `UNITFUL_FANCY_EXPONENTS` to either `true` or `false` to force using or
 the exponents. You can also set the `:fancy_exponent` IO context property to either `true`
 or `false` to force using or not using the exponents.
 
+## Specifying precision
+
+If you need to set precision of displayed value of your unitful quantity, you can use the
+`round` function as follows:
+
+```julia
+julia> d = 1.7438748921932u"mm"
+1.7438748921932 mm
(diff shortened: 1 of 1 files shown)


sostock:
I think the wording here is misleading: `round` does not “set precision of displayed value”. The `round` function does not display anything by itself, it returns a rounded value. Describing this as _specifying precision for display_ sounds like it is some equivalent to `printf` formatting.

I think it would be good to add a section about rounding. In that section, we could explain why it is necessary to specify the output unit (or is it obvious enough?).

Note that you can also specify a unit instead of a type as first argument. We should document this as well:
```julia
julia> round(typeof(d), d; sigdigits=3)
1.74 mm

julia> round(u"mm", d; sigdigits=3)
1.74 mm

julia> round(u"m", d; sigdigits=3)
0.00174 m
```

mateuszbaran:
"set precision of displayed value" is the thing I wanted to do, so to me this is the key part that needs documenting. If you thing this should be done differently than using `round`, I'm open to suggestions.

mateuszbaran:
I had a quantity computed with way too many digits and when printed it to a string it gave me way too many digits, more than the precision with which the quantity was determined. So it looked silly and I wanted to trim it -- and it's definitely not obvious how it should be done.

Issue #461, opened 2021-07-10
Multiplication of gains is problematic


Keno:
Multiplication between `Gain`s is defined to be the same as addition:

```
julia> 2dB + 5dB
7 dB

julia> 2dB * 5dB
7 dB
```

I think this is somewhat problematic. Multiplication is supposed to be commutative/associative, which this is not:
```
julia> 2 * 5 * (dB * dB)
ERROR: ArgumentError: cannot multiply logarithmic units together.
Stacktrace:
 [1] *(#unused#::Unitful.MixedUnits{Gain{Unitful.LogInfo{:Decibel, 10, 10}, :?, T} where T<:Real, Unitful.FreeUnits{(), NoDims, nothing}}, #unused#::Unitful.MixedUnits{Gain{Unitful.LogInfo{:Decibel, 10, 10}, :?, T} where T<:Real, Unitful.FreeUnits{(), NoDims, nothing}})
   @ Unitful ~/.julia/packages/Unitful/0oqX5/src/logarithm.jl:102
 [2] top-level scope
   @ REPL[25]:1
```

I can see why this is convenient, but the `*` generic function really is supposed to behave like multiplication. If you're using a logarithmic scale, use the addition functions. @sostock you seem to have been the last person to touch this code extensively, thoughts?

sostock:
I also think this should be changed. I would be in favor of disallowing the multiplication. The question is whether we consider this a breaking change.

Issue #402, opened 2020-11-26
Equality between `Gain`s is not transitive


sostock:
When comparing a power gain `Gain{L, :p}` and a root-power gain `Gain{L, :rp}` with the same `L`, they get promoted to `Gain{L, :?}` without changing their numerical values. This breaks the transitivity of `==`:
```julia
julia> a = Gain{Unitful.Decibel, :rp}(20) # 20 dB (root-power) == factor 10
20 dB

julia> b = Gain{Unitful.Decibel, :p}(20) # 20 dB (power) == factor 100
20 dB

julia> c = Gain{Unitful.Bel, :p}(2) # 2 B (power) == factor 100
2 B

julia> a == b == c
true

julia> a == c
false
```
Comparison involving numbers is intransitive as well:
```julia
julia> a = Gain{Unitful.Decibel, :rp}(20) # 20 dB (root-power) == factor 10
20 dB

julia> b = Gain{Unitful.Decibel, :p}(10) # 10 dB (power) == factor 10
10 dB

julia> c = Gain{Unitful.Decibel, :p}(20) # 20 dB (power) == factor 100
20 dB

julia> a == 10 == b
true

julia> a == b
false

julia> 10 == a == c
true

julia> 10 == c
false
```
To obtain a consistent behavior for `==`, the promotion rules for `Gain` would have to be changed. It might be difficult to figure out a set of rules that is both practical and consistent, unless one gets rid of `Gain{L, :?}` altogether.

sostock:
One consequence of the intransitive `isequal` is that a correct `hash` implementation for `Gain` is impossible: We have

* `isequal(Gain{Decibel, :p}(20), Gain{Decibel,:rp}(20))`,
* `isequal(Gain{Decibel, :p}(20), 100)`, and
* `isequal(Gain{Decibel,:rp}(20), 10)`,

but `hash(10) != hash(100)`.

PR #463, opened 2021-07-10
Tweak Gain arithmetic 

Keno:
This tweaks gain arithmetic, by making both `Gain` and `Level`
either purely linear or purely logarithmic, semantically. `Level`
was already mostly linear, with a few minor exceptions that were
removed here. `Gain` was previously logarithmic for purposes of
`+`,`-` and linear for purposes of `*`,`/` (i.e. both sets of
operations were equivalent - #461). This makes it purely logarithmic,
i.e. `*` and `/` between Gains are disallowed. When arithmetic
is mixed, only one of `(+,-)`, `(*,/)` is disallowed. Which
one depends on whether the non-Gain unit involved is a Level
(in which case `(+,-)` is used) or a regular Quantity (in which
case `(*,/)` is used). Note that this design leaves all examples
that are currently in the docs on this topic unchanged (though
I did rewrite the docs to clarify the new semantics). I'll take
that as a hint that this is a reasonable design that is not too
breaking to people's expectations. I also think it's a lot more
consistent.

The primary advantage of doing this is that it'll allow us to
reclaim other arithmetic operations like `*`/`/` on `Gain`,
while being quite consistent. For example, I'm workign with
an SDR that supports gains between `0.0dB` and `1.0dB` in its
amplifier, in setting increments of `1/256dB` (linearlly spaced
in dB space, i.e. logarithmically spaced). I'd like to represent
this as `0.0dB:(1/256)dB:1.0dB`, but with the current definition
of arithmetic, the definitions of range are unusable, because
the arithmetic system is iternally non-consistent.

This does not yet add those new operations, becuase some things
(like `/(::Gain, ::Gain)`) would change meaning, so I'd like to
give people a couple weeks to notice that things changed, but
I'm hoping to do that in the future.

cc @sostock - let me know what you think

### DIFF
--- a/docs/src/logarithm.md
+++ b/docs/src/logarithm.md
@@ -82,13 +82,10 @@ julia> (@dB 10mW/1mW) + (@dB 10mW/2mW)
 20 mW
 ```
 
-Addition will be discussed more later.
-
-Note that logarithmic "units" can only multiply or be multiplied by pure numbers and linear
-units, not other logarithmic units or quantities. This is done to avoid issues with
-commutativity and associativity, e.g. `3*dB*m^-1 == (3dB)/m`, but `3*m^-1*dB == (3m^-1)*dB`
-does not make much sense. This is because `dB` acts more like a constructor than a proper
-unit.
+With a few exceptions, dimensionful logarithmic units, such as `dBm` behave just like the underlying
+linear unit for purposes of arithmetic (i.e. arithmetic operations commute with `linear`). However,
+note that they will still be displayed logarithmically. In contrast, arithmetic on dimensionless
+logarithmic units (i.e. gains/attenuations) such as `dB` behaves logarithmically. This will be explored in more detail below.
 
 The `@dB` and `@Np` macros will fail if either a dimensionless number or a ratio of
 dimensionless numbers is used. This is because the ratio could be of power quantities or of
@@ -134,11 +131,11 @@ logarithmic quantity is:
     Unitful.Gain
 ```
 
-One might expect that any gain / attenuation factor should be convertible to a pure number,
+One might expect that any gain / attenuation factor should be convertible to a scalar,
 that is, to `x == y/z` if you had `10*log10(x)` dB. However, it turns out that in dB, a ratio
 of powers is defined as `10*log10(y/z)`, but a ratio of voltages or other root-power
-quantities is defined as `20*log10(y/z)`. Clearly, converting back from decibels to a real
-number is ambiguous, and so we have not implemented automatic promotion to avoid incorrect
+quantities is defined as `20*log10(y/z)`. Clearly, converting back from decibels to a scalar
+is ambiguous, and so we have not implemented automatic promotion to avoid incorrect
 results. You can use [`Unitful.uconvertp`](@ref) to interpret a `Gain` as a ratio of power
 quantities (hence the `p` in `uconvertp`), or [`Unitful.uconvertrp`](@ref) to interpret as
 a ratio of root-power (field) quantities.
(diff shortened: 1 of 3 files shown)


docs/src/logarithm.md
```diff
@@ -134,11 +131,11 @@ logarithmic quantity is:
     Unitful.Gain
 ```
 
-One might expect that any gain / attenuation factor should be convertible to a pure number,
+One might expect that any gain / attenuation factor should be convertible to a scalar,
```
sostock, line 134: I agree that “pure number” is not a good term here, but I also don’t like “scalar”. To me, scalar means “not a vector”, and even though that is technically correct (dimensionful quantities form a vector space), I think it doesn’t make this more clear. I would write “real number”, but I see that you replaced that as well. Can you explain why you prefer “scalar” over “real number”?

If we agree on “scalar” here, it should also be used in the first sentence in the “Constructing logarithmic quantities” section.

docs/src/logarithm.md
```diff
@@ -167,34 +164,55 @@ Finally, for completeness we note that both `Level` and `Gain` are subtypes of `
     Unitful.LogScaled
 ```
 
-## Multiplication rules
+## Addition and Multiplication rules
```
sostock, line 167:
```suggestion
## Addition and multiplication rules
```

docs/src/logarithm.md
```diff
@@ -167,34 +164,55 @@ Finally, for completeness we note that both `Level` and `Gain` are subtypes of `
...
 
-Multiplying a dimensionless logarithmic quantity by a pure number acts as like it does for
-linear quantities:
+For dimensionaless logarithmic units, addition behaves as one might expect:
```
sostock, line 169:
```suggestion
For dimensionless logarithmic quantities, addition behaves as one might expect:
```

docs/src/logarithm.md
```diff
@@ -167,34 +164,55 @@ Finally, for completeness we note that both `Level` and `Gain` are subtypes of `
...
-julia> 3u"dB" * 2
-6 dB
+julia> 10u"dB" + 10u"dB"
+20 dB
+```
 
-julia> 2 * 0u"dB"
-0 dB
+I.e. the gains add. However, as hinted at above, dimensionful logarithmic units,
+behave as their corresponding linear unit:
```
sostock, line 177:
```suggestion
I.e. the gains add. However, as hinted at above, dimensionful logarithmic quantities
behave as their corresponding linear quantity:
```

docs/src/logarithm.md
```diff
@@ -82,13 +82,10 @@ julia> (@dB 10mW/1mW) + (@dB 10mW/2mW)
...
 
-Addition will be discussed more later.
-
-Note that logarithmic "units" can only multiply or be multiplied by pure numbers and linear
-units, not other logarithmic units or quantities. This is done to avoid issues with
-commutativity and associativity, e.g. `3*dB*m^-1 == (3dB)/m`, but `3*m^-1*dB == (3m^-1)*dB`
-does not make much sense. This is because `dB` acts more like a constructor than a proper
-unit.
+With a few exceptions, dimensionful logarithmic units, such as `dBm` behave just like the underlying
+linear unit for purposes of arithmetic (i.e. arithmetic operations commute with `linear`). However,
```
sostock, line 86: `linear` is not yet introduced at this point, so users will not know what this means.

docs/src/logarithm.md
```diff
@@ -167,34 +164,55 @@ Finally, for completeness we note that both `Level` and `Gain` are subtypes of `
...
 
-Justification by example: consider the example of the exponential attenuation of a signal on
-a lossy transmission line. If the attenuation goes like $10^{-kx}$, then the (power)
-attenuation in dB is $-10kx$. We see that the attenuation in dB is linear in length. For an
-attenuation constant of 3dB/m, we better calculate 6dB for a length of 2m.
+Note that this may seem strange from an arithmetic perspective, as written, but
+the arithmetic is entirely consistent. It can be helpful to think of the arithmetic
+as being performed on the linear units, with the logarithmic units simply being a
+display hint (although the quantity being stored is indeed the displayed logarithmic
+value).
```
sostock, line 194: This is not true. `Level` does actually store the linear value:
```julia
julia> @dB 100u"mW"/1u"mW"
20.0 dBm

julia> ans.val
100 mW
```

docs/src/logarithm.md
```diff
@@ -228,14 +246,29 @@ julia> 0u"dB/Hz"
...
 ```
 
-Mathematical operations are forwarded to the logarithmic part, so that for example,
-`100*((0dBm)/s) == (20dBm)/s`. We allow linear units to commute with logarithmic quantities
-for convenience, though the association is understood (e.g. `s^-1*(3dBm) == (3dBm)/s`).
+Since dimensionful logarithmic units still behave as their corresponding linear unit,
+working with dimensionful units is entirely consistnet.
 
-The behavior of multiplication is summarized in the following table, with entries marked by
+The behavior of addition and multiplication are summarized in the following table, with entries marked by
```
sostock, line 252:
```suggestion
The behavior of addition and multiplication is summarized in the following tables, with entries marked by
```

docs/src/logarithm.md
```diff
@@ -228,14 +246,29 @@ julia> 0u"dB/Hz"
 [0 dB] Hz^-1
 ```
 
-Mathematical operations are forwarded to the logarithmic part, so that for example,
-`100*((0dBm)/s) == (20dBm)/s`. We allow linear units to commute with logarithmic quantities
-for convenience, though the association is understood (e.g. `s^-1*(3dBm) == (3dBm)/s`).
+Since dimensionful logarithmic units still behave as their corresponding linear unit,
+working with dimensionful units is entirely consistnet.
```
sostock, line 250:
```suggestion
working with dimensionful units is entirely consistent.
```

docs/src/logarithm.md
```diff
@@ -228,14 +246,29 @@ julia> 0u"dB/Hz"
 [0 dB] Hz^-1
 ```
 
-Mathematical operations are forwarded to the logarithmic part, so that for example,
-`100*((0dBm)/s) == (20dBm)/s`. We allow linear units to commute with logarithmic quantities
-for convenience, though the association is understood (e.g. `s^-1*(3dBm) == (3dBm)/s`).
+Since dimensionful logarithmic units still behave as their corresponding linear unit,
```
sostock, line 249:
```suggestion
Since dimensionful logarithmic quantities still behave as their corresponding linear quantities,
```

docs/src/logarithm.md
```diff
@@ -276,88 +309,34 @@ julia> 1u"V" * 20u"dB"
...
-1.1 kg m^2 s^-3
-```
-i.e. `1.1 W`.
+julia> 10u"dBm" * 20u"dB"
+ERROR: ArgumentError: Multiplying a level by a Gain is disallowed. Use addition, or `linear` depending on context.
 
-Rules for addition are summarized in the following table, with entries marked by †
-indicating prohibited operations. This table is populated automatically whenever the docs
-are built.
+julia> linear(10u"dBm") * 20u"dB"
```
sostock, line 333:
```suggestion
julia> 10u"mW" * 20u"dB"
```
I think this would make it more clear.

docs/src/logarithm.md
```diff
@@ -276,88 +309,34 @@ julia> 1u"V" * 20u"dB"
...
-tab = fill("", length(head), length(head))
-for col = eachindex(head), row = 1:col
-    try
-        tab[row, col] = string(quantities[row] + quantities[col])
-    catch
-        tab[row, col] = "†"
-    end
-end
-mdtable(tab, latex=false, head=head, side=side)
+julia> linear(10u"dBm") + 20u"dB"
```
sostock, line 336:
```suggestion
julia> 10u"mW" + 20u"dB"
```

sostock (commented):
This first review is only about the docs, I haven’t looked at the code yet.

sostock:
We should definitely fix the arithmetic to be more consistent. However, I would consider this a breaking change, and if we change `Gain` arithmetic for a 2.0 release, I would prefer to fix other issues with the arithmetic as well, for example #402. Of course, this can happen in a separate PR, but maybe it is worth to work out a consistent arithmetic now and discuss what changes we want to make.

sostock:
Negating a `Level` throws an error, but subtracting a larger from a smaller level does not. I think this should be disallowed (it errors when printed):
```julia
julia> 2u"dBm" - 3u"dBm";

julia> show(ans)
ERROR: DomainError with -0.4103691225077659:
NaN result for non-NaN input.
```

PR #467, opened 2021-07-17
Round-trip printing

YingboMa:


sostock:
Thanks for your contribution! As indicated by the CI, you need to adapt the tests to the new behavior. It would also be good to add tests for printing with `MIME"text/plain"` and to test that round-tripping actually works.

Issue #310, opened 2020-02-25
`isapprox` for heterogeneous arrays


sostock:
`isapprox` behaves incorrectly for heterogeneous arrays of quantities:
```julia
julia> a = Quantity[1u"s", 1u"minute"]
2-element Array{Quantity,1}:
      1 s
 1 minute

julia> isapprox(a,a)
false
```
This is due to the following implementation:
https://github.com/PainterQubits/Unitful.jl/blob/f6866eaffbb3a9150921cdcdacc967e8ff4a5c03/src/quantities.jl#L247-L248

m-wells:
Maybe something like `all(isapprox.(x,y))`?

sostock:
> Maybe something like `all(isapprox.(x,y))`?

That’s probably too simple, since `isapprox` should use the specified `norm` for arrays. From the docstring:

>   `x` and `y` may also be arrays of numbers, in which case `norm` defaults to the usual `norm` function in LinearAlgebra, but may be changed by passing a `norm::Function`
  keyword argument. (For numbers, `norm` is the same thing as `abs`.) When `x` and `y` are arrays, if `norm(x-y)` is not finite (i.e. `±Inf` or `NaN`), the comparison falls back
  to checking whether all elements of `x` and `y` are approximately equal component-wise.

m-wells:
Perhaps but it could serve as a fallback for now instead of the current always `false` method.

sostock:
Another problem is that it is difficult to choose a default `rtol` if the array contains different units. Maybe `isapprox` should throw an error for heterogeneous arrays if the `rtol` keyword argument is not given.

PR #468, opened 2021-07-17
Fix #465 for isapprox with complex arrays

### LINKED ISSUE #465: `isapprox` for complex unitful arrays fails
MWE for v1.8.0:

```julia
using Unitful: mm
z = (1+im) * ones(2,3) * mm
isapprox(z, 2z)

ERROR: MethodError: no method matching isless(::ComplexF64, ::ComplexF64)
...
```

It works fine for a complex scalar `z = (1 + im) * mm`
This is perhaps slightly related to #310 but should be an easier case.

JeffFessler:
Fix #465
The substantive change was `real{T1}` on line 300.

The rest of it is just white space to make it easier for me to sort out why it worked for scalars and not arrays.
The scalar case on line 282 has `real(T)` in it, so now it is matching between scalar and array.

This seems like a bug fix so I put in v1.9.1, but if you prefer v1.10.0 because it now it is new support for complex arrays that's fine too.  Up to you...


src/quantities.jl
```diff
@@ -275,18 +275,31 @@ for (i,j) in zip((:<, :isless), (:_lt, :_isless))
...
-isapprox(x::AbstractQuantity{T,D,U}, y::AbstractQuantity{T,D,U}; atol=zero(Quantity{real(T),D,U}), kwargs...) where {T,D,U} =
-    isapprox(x.val, y.val; atol=uconvert(unit(y), atol).val, kwargs...)
+
+function isapprox(
+    x::AbstractQuantity{T,D,U},
+    y::AbstractQuantity{T,D,U};
+    atol = zero(Quantity{real(T),D,U}),
+    kwargs...,
+) where {T,D,U}
+    return isapprox(x.val, y.val; atol=uconvert(unit(y), atol).val, kwargs...)
```
sostock, line 285:
```suggestion
    return isapprox(x.val, y.val; atol=ustrip(unit(y), atol), kwargs...)
```

test/runtests.jl
```diff
@@ -741,6 +741,10 @@ end
         @test isapprox(1.0u"m", 1.1u"m"; rtol=0.2)
         @test !isapprox(1.0u"m", 1.1u"m"; rtol=0.05)
 
+        # Issue 465:
+        z = (1+im) * ones(2,3) * 1m
+        @test !isapprox(z, 2z)
```
giordano, line 746: Maybe also add a positive test like `isapprox(z, z)`?

src/quantities.jl
JeffFessler: I'll commit this suggestion, but I should mention that the `uconvert` version was already in there.

test/runtests.jl
JeffFessler: Done

JeffFessler:
bump - I think I have addressed all the suggestions last summer.  Then I forgot about it, but I just hit this same issue again and "rediscovered" this PR.
Hopefully the only remaining question is whether to call it v0.11.0 or v0.10.1 (I chose the latter).

test/runtests.jl
```diff
@@ -760,6 +760,11 @@ end
         @test isapprox(1.0u"m", 1.1u"m"; rtol=0.2)
         @test !isapprox(1.0u"m", 1.1u"m"; rtol=0.05)
 
+        # Issue 465:
+        z = (1+im) * ones(2,3) * 1m
```
sostock, line 764:
```suggestion
        z = fill((1+im)m, 2, 3)
```
I think this would be clearer than the multiplication chain.

sostock (approved):
LGTM. We should resolve some more issues with `isapprox`(#310, ~~support for the `nans` keyword~~), but that can happen in a future PR.

test/runtests.jl
JeffFessler: agreed!

PR #472, opened 2021-07-24
Support `zero` for array with heterogeneous units

lamorton:
When working with ModelingToolkit, we often would like to have heterogeneous units within an vector of variables. Presently there's no method to create a `zero` array for such a case, which is often needed when doing solves or automatic differentiation. 

This construction only works if an appropriate template array already exists (ie, cannot be done abstractly).  So there's no ambiguity about what the unit type is for each index.

### DIFF
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -376,6 +376,7 @@ zero(x::Type{<:AbstractQuantity{T}}) where {T} = throw(ArgumentError("zero($x) n
 zero(x::Type{<:AbstractQuantity{T,D}}) where {T,D} = zero(T) * upreferred(D)
 zero(x::Type{<:AbstractQuantity{T,D,U}}) where {T,D,U<:ScalarUnits} = zero(T)*U()
 zero(x::Type{<:AbstractQuantity{T,D,U}}) where {T,D,U<:AffineUnits} = zero(T)*absoluteunit(U())
+zero(x::AbstractArray{<:AbstractQuantity}) = map(zero,x)
 
 one(x::AbstractQuantity) = one(x.val)
 one(x::AffineQuantity) =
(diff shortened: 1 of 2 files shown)


sostock:
I think it would be better to add this to Julia itself (for all heterogeneous `AbstractArray`s). There’s an issue for that: https://github.com/JuliaLang/julia/issues/33099.

lamorton:
Makes sense. Lemme see if I can get a PR for Base, then.

sostock:
This PR changes the behavior for arrays with heterogeneous units but fixed dimension:

Before:
```julia
julia> zero(Quantity{Int,Unitful.𝐋}[1u"m", 1u"mm"])
2-element Vector{Quantity{Int64, 𝐋}}:
 0 m
 0 m
```
After:
```julia
julia> zero(Quantity{Int,Unitful.𝐋}[1u"m", 1u"mm"])
2-element Vector{Quantity{Int64, 𝐋}}:
  0 m
 0 mm
```
But I don’t think that’s too bad. Even the old behavior returned a vector with non-concrete eltype, (even though all elements are the same), so it will not worsen performance in this case.

src/quantities.jl
```diff
@@ -376,6 +376,7 @@ zero(x::Type{<:AbstractQuantity{T}}) where {T} = throw(ArgumentError("zero($x) n
 zero(x::Type{<:AbstractQuantity{T,D}}) where {T,D} = zero(T) * upreferred(D)
 zero(x::Type{<:AbstractQuantity{T,D,U}}) where {T,D,U<:ScalarUnits} = zero(T)*U()
 zero(x::Type{<:AbstractQuantity{T,D,U}}) where {T,D,U<:AffineUnits} = zero(T)*absoluteunit(U())
+zero(x::AbstractArray{<:AbstractQuantity}) = map(zero,x)
```
sostock, line 379:
```suggestion
zero(x::AbstractArray{T}) where {T<:Unitful.AbstractQuantity} = isconcretetype(T) ? fill!(similar(x), zero(T)) : map(zero, x)
```

This way the old method is used if the array is homogeneous, which is faster according to my benchmarks.

sostock:
This is replaced by #516, which supports arrays with unassigned elements as well.

PR #476, opened 2021-08-13
Unit construction macro now defines docs for units

### LINKED ISSUE #436: Add doc strings to units?
No documentation turns up here:
```
help?> Unitful.F
  No documentation found.

  Unitful.F is of type Unitful.FreeUnits{(F,), 𝐈² 𝐓⁴ 𝐋⁻² 𝐌⁻¹, nothing}.

  Summary
  ≡≡≡≡≡≡≡≡≡

  struct Unitful.FreeUnits{(F,), 𝐈² 𝐓⁴ 𝐋⁻² 𝐌⁻¹, nothing} <: Unitful.Units{(F,), 𝐈² 𝐓⁴ 𝐋⁻² 𝐌⁻¹, nothing}
  [...]
```

Would it be help to have an autogenerated doc string that says, e.g.,
```
Untiful.F defines the Farad as 1s^4*A^2/(kg*m^2) with dimensions 𝐈² 𝐓⁴ 𝐋⁻² 𝐌⁻¹
```
It seems this should be possible since pkgdefaults.jl defines
```
@unit F               "F"    Farad           1s^4*A^2/(kg*m^2)  true
```

lukebemish:
This is a fairly minor change that simply causes Docs.getdocs to be defined for the appropriate type when a unit is defined, allowing for basic documentation of the dimensions and conversion factor of every unit defined by the "unit" macro.

Close #436


sostock:
Thanks! I think the docstrings could be improved:
```
help?> Unitful.m
"Equal to 10^0 Meter, or 1 m, with dimensions 𝐋"

help?> Unitful.mm
"Equal to 10^-3 Meter, or 1//1000 m, with dimensions 𝐋"
```
* The docstrings should probably mention that the object is a unit (not a quantity), because to me, “equal to 1 m” sounds like it describes a quantity.
* In the `Unitful.m` docstring, I don’t think it’s helpful to mention that `m` is equal to `1m`.
* The quotes should not be included in the docstrings. Adding `getdoc` methods is not the right way to implement this, it’s better to add the docstrings in the expression returned by the macro.

Also, should this be optional? Maybe users want to write their own docstrings for units instead. We could add a new optional argument to the `@unit` macro that determines whether the automatic docstring is added.

I also think we should enable documenting the `@unit` macro by just adding `Base.@__doc__` in the right place:
```
julia> "docstring for MySecond"
       @unit mys "mys" MySecond 1u"s" true
ERROR: cannot document the following expression:

#= REPL[20]:2 =# @unit mys "mys" MySecond 1 * u"s" true

'@unit' not documentable. See 'Base.@__doc__' docs for details.
```
However, this would document only `mys`, not `mmys` (milli-MySecond). Maybe the docstrings for the prefixed units should be added automatically even if one supplies their own docstring?

lukebemish:
Alright; I agree with what you've said, though I'll have to do some research, as I haven't played with Docs too much. What do you think the docstring for the base unit should be (i.e., meters)? The reason I had implemented this using getdoc is that I couldn't find any easy way to come up with the appropriate doc string when the macro is run, but I'll look into better alternatives, as I agree that using getdoc is by far not the best option.

Allowing docstrings for base units should be fairly easy; I may go down that method instead. I could also probably implement a fairly simple addition for power-of-ten units; something like "A unit equal to 10^-3 meters" or the like.

lukebemish:
I just added a couple of commits that should provide documentation in a completely different way, relying on user-defined documentation instead and only automatically generating the SI power-of-ten derived units documentation. All in all, I think this is a much nicer way to go, as users can add as much custom documentation as they would like to to any units they define. I also started adding documentation for some of the units defined in pkgdefaults, though I figured I should hold off on defining documentation for all of them until we've got a common documentation format. What do you think?

lukebemish:
This has stagnated slightly, mostly because I am not sure if I have enough expertise to write a doc string for all the built-in units, and I am not yet sure what a standard doc string format for these units should be. I've given it a try for some simpler units, but I would love to see what somebody else thinks of this.

sostock:
I like that the new iteration of this PR now enables the user to write their own docstrings, even if it means that we have to document every unit in pkgdefaults.jl by hand. I haven’t looked at the code in detail, hopefully I have time for it soon.

Regarding the format of the docstrings, I think something like the following would be nice. In particular I think it would be good to spell out the names of the units (meter, joule etc.):

```
    Unitful.m

The meter, base SI unit of length.

Dimension: [`Unitful.𝐋`](@ref).
```

```
    Unitful.J

The joule, an SI unit of energy, equal to 1 kg*m^2*s^-2.

Dimension: `𝐌*𝐋^2*𝐓^-2`.
```

```
    Unitful.u

The unified atomic mass unit (also called dalton), a unit of mass, equal to 1.66053906660(50)*10^-27 kg.

Dimension: [`Unitful.𝐌`](@ref)
```

The automatically generated docstrings for prefixed units could look like this:
```
    Unitful.cm

A prefixed unit, equal to 10^-2 m.

Dimension: [`Unitful.𝐋`](@ref).

See also: [Unitful.m](@ref).
```

For gram and its prefixed versions, one could add a disclaimer:
```
    Unitful.g

The gram, an SI unit of mass, equal to 10^-3 kg. Note that `kg`, not `g`, is the base unit.

Dimension: [`Unitful.𝐌`](@ref).

See also: [`Unitful.kg`](@ref).
```

lukebemish:
Getting the dimension to show up in the generated docs for prefixed units is quite hard; I haven't yet found a method that works. In fact, I now realize that the way I was making a reference to the base unit was also flawed, but I'm fixing that now. If anyone has ideas for getting nice representations of dimension in prefixed units, that would be good to look at. Otherwise, I'll go with what I have now, which is to give the prefixed unit a reference to the base unit, and the base unit information on the dimension.

lukebemish:
I've added documentation in this format for all units defined in pkgdefaults. It could definitely use some review; additionally, I'd like to figure out a standard format for log scales and log scale units, and I haven't really used those much at all in Unitful.

lukebemish:
The format for equations/values I have converted everything to:

- Numbers, multiplication, and division are separated by spaces
- Large numbers are written in scientific notation as follows: `1.234,567,890 × 10^-5`
- Exponents and fractions are not separated by spaces
- For readability, fractions are written as `1/2`, not the julia format of `1//2`

What do you think?

sostock:
I like the use of `NoDims` instead of “dimensionless” and I also like the number format you have chosen. It is still somewhat inconsistently used with regard to spaces, e.g., the `rad` docstring uses `2π` and the `ħ` docstring uses `2 π` (with space). I would prefer the former. It also happens with multiplication of units, e.g., the `h` docstring uses `J*s` and the `C` docstring uses `A × s` (notice also the different multiplication sign).

lukebemish:
I don't currently add any documentation for log units, so I think that this documentation and the automatic generation of documentation for the macros, functions, and the like that log units also define can be left for another pull request. I believe that everything is (hopefully) actually using the same standardized format for the doc strings now; I have added automatically generated doc strings for the likes of `Unitful.Length`, `Unitful.LengthUnits`, and `Unitful.LengthFreeUnits`, though these could use some review.

lukebemish:
I attempted to build the documentation with this pull request, and realized that it has a LOT of warnings, courtesy of the hundreds of objects whose new doc strings are not in the documentation. If I were to put together a page of the documentation to hold all the information on built in units, how would it be best organized? I can use `@autodocs` to fetch all the unit/dimension/etc. documentation, but I'm not sure what the best ordering/format for this would be. What do you think?

lukebemish:
And should this be part of this pull request?

sostock:
I don’t think it is helpful to have all prefixed units listed in the documentation, maybe only the non-prefixed versions? In that case we would have to live with the warnings (I don’t think that’s too bad, it’s only one big warning, not several ones). I think we can leave the documentation for another PR, this one is already pretty big.

lukebemish:
What should we go with in the way of tests? I can't figure out any easy way to tell if something is documented, which is the direction we'd probably like to go in.

lukebemish:
Oh wait, docs can be cast to strings. I'll give this a try.

lukebemish:
The tests are failing because multiline-string indentation is weird in julia 1.0:

```julia
help?> Unitful.kJ
                       Unitful.kJ

  A prefixed unit, equal to 10^3 J.

  Dimension: 𝐋^2 𝐌 𝐓^-2

  See also: Unitful.J.
```

I'm not really sure how to fix this, other than making sure no multiline text blobs are indented at all. This is a bit of a pain, but may be necessary for julia-1.0 compatibility.

lukebemish:
I managed to fix this issue, but now I'm running into something else weird:

```
julia> string(@doc Unitful.Length)
"```\nUnitful.Length{T, U}\n```\n\nA supertype for quantities and levels of dimension [`Unitful.𝐋`](@ref) with a value                of type `T` and units `U`.\n\nSee also: [`Unitful.𝐋`](@ref), [`Unitful.Quantity`](@ref), [`Unitful.Level`](@ref).\n"
```

There's a random giant space in the middle. I think I know why; it's because that's where the indent is in the string. I think the only nice solution here is to just get rid of indents in multiline docstrings entirely.

sostock:
What did you try to fix it? I think it should be enough to just replace every `\n` in triple-quoted strings by actual linebreaks:
```julia
    name_doc = """
                   $__module__.$name{T, U}

               A supertype for quantities and levels of dimension [`$__module__.$s`](@ref) with a value
               of type `T` and units `U`.

               See also: [`$__module__.$s`](@ref), $name_links.
               """

```
```
julia> string(@doc Unitful.Length)
"```\nUnitful.Length{T, U}\n```\n\nA supertype for quantities and levels of dimension [`Unitful.𝐋`](@ref) with a value of type `T` and units `U`.\n\nSee also: [`Unitful.𝐋`](@ref), [`Unitful.Quantity`](@ref), [`Unitful.Level`](@ref).\n"
```

lukebemish:
I'll give that a try. That may be the issue.

PR #478, opened 2021-08-22
Attempt to allow preferunits to work with non-pure units

### LINKED ISSUE #457: incorrect preferred units conversion with atomic units
```Julia
using Unitful

using Unitful: preferunits
using UnitfulAtomic: me_au, e_au

preferunits(u"cm", u"ns", me_au, e_au / u"ns")

@show dimension(upreferred(u"V/cm"))
@show dimension(u"V/cm")

@assert dimension(upreferred(u"V/cm")) == dimension(u"V/cm")

```

results in

```Julia
dimension(upreferred(u"V/cm")) = 𝐋 𝐌 𝐈⁻¹ 𝐓⁻⁴
dimension(u"V/cm") = 𝐋 𝐌 𝐈⁻¹ 𝐓⁻³
ERROR: LoadError: AssertionError: dimension(upreferred(u"V/cm")) == dimension(u"V/cm")
[...]
```

with

```
[1986cc42] Unitful v1.8.0
[a7773ee8] UnitfulAtomic v1.0.0
```

lukebemish:
Should fix #457. Prior to this change, attempts to use something that looked like `preferunits(u"C/ms")` would result in strange behavior, without throwing any sort of errors. Now, that should work nicely. This means that, for example:

```
julia> Unitful.preferunits(u"ms",u"C/ms")

julia> upreferred(50u"A")
1//20 C ms^-1

julia> upreferred(50u"mA*ns")
1//20000000000 C

julia> upreferred(50u"mA/mC")
1//20 ms^-1
```

preferunits still requires units with pure dimensions of power=1.

I ran all of the tests and looked through the rest of the code for anywhere this could cause issues or incompatibilities, and I couldn't find anything, as upreferred(x::Dimensions) and preferunits seem to be the only things that touch the promotions Dict.

### DIFF
--- a/src/units.jl
+++ b/src/units.jl
@@ -302,6 +302,6 @@ factory defaults, this function will return a product of powers of base SI units
 (as [`Unitful.FreeUnits`](@ref)).
 """
 @generated function upreferred(x::Dimensions{D}) where {D}
-    u = *(FreeUnits{((Unitful.promotion[name(z)]^z.power for z in D)...,),()}())
+    u = prod(Unitful.promotion[name(z)]^z.power for z in D)
     :($u)
 end
--- a/src/units.jl
+++ b/src/units.jl
@@ -302,6 +302,6 @@ factory defaults, this function will return a product of powers of base SI units
 (as [`Unitful.FreeUnits`](@ref)).
 """
 @generated function upreferred(x::Dimensions{D}) where {D}
-    u = prod((Unitful.NoUnits, (Unitful.promotion[name(z)]^z.power for z in D)...))
+    u = prod((NoUnits, (promotion[name(z)]^z.power for z in D)...))
     :($u)
 end
--- a/src/units.jl
+++ b/src/units.jl
@@ -302,6 +302,6 @@ factory defaults, this function will return a product of powers of base SI units
 (as [`Unitful.FreeUnits`](@ref)).
 """
 @generated function upreferred(x::Dimensions{D}) where {D}
-    u = prod(Unitful.promotion[name(z)]^z.power for z in D)
+    u = prod([Unitful.NoUnits, (Unitful.promotion[name(z)]^z.power for z in D)...])
     :($u)
 end
--- a/src/units.jl
+++ b/src/units.jl
@@ -302,6 +302,6 @@ factory defaults, this function will return a product of powers of base SI units
 (as [`Unitful.FreeUnits`](@ref)).
 """
 @generated function upreferred(x::Dimensions{D}) where {D}
-    u = prod([Unitful.NoUnits, (Unitful.promotion[name(z)]^z.power for z in D)...])
+    u = prod((Unitful.NoUnits, (Unitful.promotion[name(z)]^z.power for z in D)...))
     :($u)
 end
--- a/src/user.jl
+++ b/src/user.jl
@@ -318,17 +318,19 @@ function preferunits(u0::Units, u::Units...)
     # unexpected (though still correct) results. There may be cases where people would want to do this,
     # so we won't stop them, just let them know they're doing it.
     ulist = (typeof(i[2]).parameters[1] for i in promotion)
-    ulist = (i^(1/i.power) for i in ulist)
-    check = Dict()
-    for i in ulist
-        k = string(dimension(i))
-        if haskey(check, k)
-            if i!=check[k]
-                @warn "Preferred units contain complex units based on units of the same dimension but different scales: "*string(i)*" and "*string(check[k])*
-                "\nThis may be intentional, but otherwise could lead to redundant units, such as ms/s or kg/g"
+    check = Dict{String, Unit}()
+    for a in ulist
+        ulistA = (i^(1/i.power) for i in a)
+        for i in ulistA
+            k = string(dimension(i))
+            if haskey(check, k)
+                if i!=check[k]
+                    @warn "Preferred units contain complex units based on units of the same dimension but different scales: "*string(i)*" and "*string(check[k])*
+                    "\nThis may be intentional, but otherwise could lead to redundant units, such as ms/s or kg/g"
+                end
+            else
+                check[k] = i
             end
-        else
-            check[k] = i
         end
     end
 
--- a/src/user.jl
+++ b/src/user.jl
@@ -317,14 +317,14 @@ function preferunits(u0::Units, u::Units...)
     # was defined in ns. We'll give a warning if this is the case so the user knows they may get
     # unexpected (though still correct) results. There may be cases where people would want to do this,
     # so we won't stop them, just let them know they're doing it.
-    ulist = [([typeof(i[2]).parameters[1] for i in promotion]...)...]
-    ulist = [i^(1/i.power) for i in ulist]
+    ulist = (typeof(i[2]).parameters[1] for i in promotion)
+    ulist = (i^(1/i.power) for i in ulist)
     check = Dict()
     for i in ulist
         k = string(dimension(i))
         if haskey(check, k)
             if i!=check[k]
-                @warn "Prefered units contain complex units based on units of the same dimension but different scales: "*string(i)*" and "*string(check[k])*
+                @warn "Preferred units contain complex units based on units of the same dimension but different scales: "*string(i)*" and "*string(check[k])*
                 "\nThis may be intentional, but otherwise could lead to redundant units, such as ms/s or kg/g"
             end
         else
--- a/src/user.jl
+++ b/src/user.jl
@@ -310,7 +310,7 @@ function preferunits(u0::Units, u::Units...)
             "For instance, it should not be used with units of dimension 𝐋^2.")
         end
         y = typeof(dim).parameters[1][1]
-        promotion[name(y)] = typeof(unit).parameters[1][1]
+        promotion[name(y)] = unit
     end
 
     nothing
--- a/src/user.jl
+++ b/src/user.jl
@@ -313,6 +313,25 @@ function preferunits(u0::Units, u::Units...)
         promotion[name(y)] = unit
     end
 
+    # Let's run a quick check for anything where, for instance, current was defined is C/ms and time
+    # was defined in ns. We'll give a warning if this is the case so the user knows they may get
+    # unexpected (though still correct) results. There may be cases where people would want to do this,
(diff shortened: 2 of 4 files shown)


src/user.jl
```diff
@@ -310,7 +310,26 @@ function preferunits(u0::Units, u::Units...)
...
         end
         y = typeof(dim).parameters[1][1]
-        promotion[name(y)] = typeof(unit).parameters[1][1]
+        promotion[name(y)] = unit
+    end
+
+    # Let's run a quick check for anything where, for instance, current was defined is C/ms and time
+    # was defined in ns. We'll give a warning if this is the case so the user knows they may get
+    # unexpected (though still correct) results. There may be cases where people would want to do this,
+    # so we won't stop them, just let them know they're doing it.
```
sostock, line 319: I would like to know what the other maintainers think about printing a warning here. On one hand, if it works correctly, why print a warning? It adds noise and slows down the code. On the other hand, I cannot think of a real usecase for preferred units based on different base units, so it is most likely a mistake and printing a warning is a good solution.

src/user.jl
```diff
@@ -310,7 +310,26 @@ function preferunits(u0::Units, u::Units...)
...
+    ulist = [([typeof(i[2]).parameters[1] for i in promotion]...)...]
```
giordano, line 320: I'm confused by this line:
```julia
[typeof(i[2]).parameters[1] for i in promotion]
```
is a vector generated by comprehension, then you splat it in a tuple, then you splat the tuple in a vector again?

Why not just
```suggestion
    ulist = (typeof(i[2]).parameters[1] for i in promotion)
```
which is a lazy generator instead?

src/user.jl
```diff
@@ -310,7 +310,26 @@ function preferunits(u0::Units, u::Units...)
...
+    ulist = [i^(1/i.power) for i in ulist]
```
giordano, line 321: Similar to the line above, you can use a lazy generator instead of allocating a vector:
```suggestion
    ulist = (i^(1/i.power) for i in ulist)
```

src/units.jl
```diff
@@ -302,6 +302,6 @@ factory defaults, this function will return a product of powers of base SI units
 (as [`Unitful.FreeUnits`](@ref)).
 """
 @generated function upreferred(x::Dimensions{D}) where {D}
-    u = *(FreeUnits{((Unitful.promotion[name(z)]^z.power for z in D)...,),()}())
+    u = prod([Unitful.NoUnits, (Unitful.promotion[name(z)]^z.power for z in D)...])
```
giordano, line 305: Also here you can avoid allocating an array:
```suggestion
    u = prod((Unitful.NoUnits, (Unitful.promotion[name(z)]^z.power for z in D)...))
```

src/user.jl
```diff
@@ -310,7 +310,26 @@ function preferunits(u0::Units, u::Units...)
...
+    check = Dict()
```
giordano, line 322: Can this dictionary by typed?

src/user.jl
giordano: I'm not sure `preferunits` is a performance-critical function, so slowing it down is fine, in my opinion.  I think removing a warning is non-breaking, so we could change the behaviour later in case we don't like it.

test/runtests.jl
```diff
@@ -320,6 +321,15 @@ Unitful.uconvert(U::Unitful.Units, q::QQQ) = uconvert(U, Quantity(q.val, cm))
 
 @testset "Promotion" begin
     @testset "> Unit preferences" begin
+        # Should warn on possible redundant units issue (ms and s)
+        @test_logs (:warn,) Unitful.preferunits(C/ms)
```
giordano, line 325: I'd check the message with something like
```suggestion
        @test_logs (:warn, r"^Preferred units contain complex units") Unitful.preferunits(C/ms)
```
to make sure we aren't catching something else by accident (even though it's very unlikely)

src/user.jl
```diff
@@ -310,7 +310,26 @@ function preferunits(u0::Units, u::Units...)
...
+    for i in ulist
+        k = string(dimension(i))
+        if haskey(check, k)
+            if i!=check[k]
+                @warn "Prefered units contain complex units based on units of the same dimension but different scales: "*string(i)*" and "*string(check[k])*
```
giordano, line 327:
```suggestion
                @warn "Preferred units contain complex units based on units of the same dimension but different scales: "*string(i)*" and "*string(check[k])*
```

giordano (changes requested):

giordano (approved):
Looks better to me now

src/user.jl
```diff
@@ -310,7 +310,28 @@ function preferunits(u0::Units, u::Units...)
...
+    # Let's run a quick check for anything where, for instance, current was defined is C/ms and time
+    # was defined in ns. We'll give a warning if this is the case so the user knows they may get
+    # unexpected (though still correct) results. There may be cases where people would want to do this,
+    # so we won't stop them, just let them know they're doing it.
+    ulist = (typeof(i[2]).parameters[1] for i in promotion)
+    check = Dict{String, Unit}()
+    for a in ulist
+        ulistA = (i^(1/i.power) for i in a)
+        for i in ulistA
+            k = string(dimension(i))
```
sostock, line 325: I wouldn’t use `string` for this, because it doesn’t have to be unique (it’s possible to define different dimensions that print the same, although I wouldn’t recommend it). The symbol associated with a dimension (like `:Length`) must be unique, it’s better to use that.
We could define a function like
```julia
function pure_dimension(x::Units) = pure_dimension(dimension(x))
function pure_dimension(::Dimensions{D}) where D
    (isone(length(D)) && isone(D[1].power)) || error("dimension is not pure")
    D[1]
end
```
and then use `name(pure_dimension(i))` instead of `string(dimension(i))`. I haven’t tested it though.

lukebemish: The reason I went with this is that the dimensions here don't have to be pure - this would also work if the user defined something in terms of, say, `mC` and `C`. I agree that the string representation is sup-par. I also think I figured it out, though; it works just fine if we just use `dimension(i)` as a dictionary key, since as far as I'm aware a `Unitful.Dimensions` object should be hashable.

src/units.jl
```diff
@@ -302,6 +302,6 @@ factory defaults, this function will return a product of powers of base SI units
 (as [`Unitful.FreeUnits`](@ref)).
 """
 @generated function upreferred(x::Dimensions{D}) where {D}
-    u = *(FreeUnits{((Unitful.promotion[name(z)]^z.power for z in D)...,),()}())
+    u = prod((Unitful.NoUnits, (Unitful.promotion[name(z)]^z.power for z in D)...))
```
sostock, line 305:
```suggestion
    u = prod((NoUnits, (promotion[name(z)]^z.power for z in D)...))
```
This is part of the `Unitful` module, so these module qualifiers aren’t necessary.

src/user.jl
```diff
@@ -310,7 +310,28 @@ function preferunits(u0::Units, u::Units...)
...
-        promotion[name(y)] = typeof(unit).parameters[1][1]
+        promotion[name(y)] = unit
+    end
+
+    # Let's run a quick check for anything where, for instance, current was defined is C/ms and time
+    # was defined in ns. We'll give a warning if this is the case so the user knows they may get
+    # unexpected (though still correct) results. There may be cases where people would want to do this,
+    # so we won't stop them, just let them know they're doing it.
+    ulist = (typeof(i[2]).parameters[1] for i in promotion)
+    check = Dict{Unitful.Dimensions, Unit}()
```
sostock, line 321:
```suggestion
    check = Dict{Dimensions, Unit}()
```

test/runtests.jl
```diff
@@ -320,6 +321,15 @@ Unitful.uconvert(U::Unitful.Units, q::QQQ) = uconvert(U, Quantity(q.val, cm))
...
 @testset "Promotion" begin
     @testset "> Unit preferences" begin
+        # Should warn on possible redundant units issue (ms and s)
+        @test_logs (:warn, r"^Preferred units contain complex units") Unitful.preferunits(C/ms)
+        # Test for wacky prefered units functionality
+        Unitful.preferunits(C/s)
+        @test @inferred(upreferred(u"V/m")) == u"kg*m*C^-1*s^-2"
+        @test dimension(upreferred(u"V/m")) == dimension(u"V/m")
+        # Reset prefered units to default
+        Unitful.preferunits(A)
```
sostock, line 331: Note that resetting this will not restore the behavior of `upreferred(u"V/m")`. Because `upreferred(::Dimensions)` is a generated method, it will only restore the original behavior for dimensions on which `upreferred` has not been called yet.

test/runtests.jl
```diff
@@ -320,6 +321,15 @@ Unitful.uconvert(U::Unitful.Units, q::QQQ) = uconvert(U, Quantity(q.val, cm))
 
 @testset "Promotion" begin
     @testset "> Unit preferences" begin
+        # Should warn on possible redundant units issue (ms and s)
+        @test_logs (:warn, r"^Preferred units contain complex units") Unitful.preferunits(C/ms)
+        # Test for wacky prefered units functionality
+        Unitful.preferunits(C/s)
+        @test @inferred(upreferred(u"V/m")) == u"kg*m*C^-1*s^-2"
+        @test dimension(upreferred(u"V/m")) == dimension(u"V/m")
```
sostock, line 329:
```suggestion
        @test @inferred(upreferred(V/m)) == kg*m*C^-1*s^-2
        @test dimension(upreferred(V/m)) == dimension(V/m)
```
All of these units are imported, so we don’t need the `u"…"` macro.

test/runtests.jl
lukebemish: Good to know. Is this likely to mess with anything elsewhere in the tests, now or in the future, or should we just leave it as is or remove the `Unitful.preferunits(A)`?

sostock: To avoid potential confusion in the future, we could move these tests to the end (i.e., move the testset to the end of the test suite and the new tests to the end of the testset), but I don’t think that’s necessary. Maybe the comment could be extended, e.g.
```
# Reset preferred units to default, except for units of dimension 𝐋*𝐌*𝐈^-1*𝐓^-3,
# because upreferred has already been called for that dimension
```

sostock (approved):

Issue #37, opened 2016-10-13
"Physicist" conversions


ajkeller34:
On master, you can now easily implement conversion between dimensions by multiplying by appropriate fundamental constants (e.g. energy = planck constant * frequency). Currently this is on a case-by-case basis, and is opt-in to avoid confusion:

```
julia> using Unitful

julia> import Unitful: uconvert, EnergyUnit, Frequency

julia> uconvert(e::EnergyUnit, f::Frequency) = uconvert(e, u"h"*f)
uconvert (generic function with 4 methods)

julia> 1u"GHz" |> u"eV"
4.135667662340165e-6 eV
```

It'd be cool if we could come up with a mechanism to implement this logic automatically given a set of fundamental constants, although this sort of thing would always need to be opt-in. Perhaps we could define a generalized `uconvert` called `dconvert` that attempts to convert quantities between dimensions using a set of fundamental constants when possible. The `h` vs `ħ` distinction would be super annoying.

MasonProtter:
Okay, so I've gotten a basic and likely suboptimal implementation of this working for general natural units. It should be able to convert any unit into powers of electron volts, and from electron volts back to any target units.

```
using Unitful
Unitful.promote_to_derived()

c  = 1u"c"
ħ  = 1.0545718e-34u"kg*m^2/s"
ħc = ħ*c
kB = 1.38064852e-23u"kg*m^2/s^2/K"
ϵ0 = 8.85418782e-12u"m^-3*kg^-1*s^4*A^2"

setDimPow(D::Dict, x::Unitful.Dimension{T}) where {T} = D[T] = x.power

function dimDict(x::Unitful.Dimensions{T}) where {T}
    D = Dict(:Length => 0,
             :Mass => 0,
             :Temperature => 0,
             :Time => 0,
             :Current => 0)
    for t in T
        setDimPow(D,t)
    end
    return D
end

dimDict(x::Unitful.Quantity{T}) where {T} = dimDict(dimension(x))

function toNaturalUnits(q)
    D = dimDict(q)
    (α,β,γ,δ,ϕ) = (D[:Length], D[:Mass], D[:Temperature], D[:Time], D[:Current])
    uconvert(u"eV"^(-α-δ+β+γ+ϕ),
             q*ħc^(-α+ϕ/2)*ħ^(-δ)*c^(2(β-ϕ/2))*kB^(γ)*(4π*ϵ0)^(-ϕ/2))
end

[...]

Issue #480, opened 2021-09-13
NaturallyUnitful


Mo8it:
Please merge [NaturallyUnitful](https://github.com/MasonProtter/NaturallyUnitful.jl) or at least mention it :)

sostock:
I don’t see a PR related to NaturallyUnitful, so I’m confused as to what should be merged. Of course we can mention it in the README. Someone just needs to make a PR that adds it to the list.

Mo8it:
The pull request is ready. I am waiting for the agreement of the original author of NaturallyUnitful.jl @MasonProtter :)

https://github.com/Mo8it/Unitful.jl/tree/feature/natural_units

Mo8it:
Ready: https://github.com/PainterQubits/Unitful.jl/pull/481

sostock:
Added to README in #483.

PR #481, opened 2021-09-14
Feature/natural units

Mo8it:
Many of Unitful.jl users are physicists and in many fields of physics, natural units are used.
There is an awesome package for conversions to and from natural units built on Unitful: [**NaturallyUnitful.jl**](https://github.com/MasonProtter/NaturallyUnitful.jl)

It is a good step to join forces. People not knowing about NaturallyUnitful.jl would benefit of the added two functions (`natural` and `unnatural`). No one needs to install an additional package.

I did include the functionality, tests and docs of NaturallyUnitful.jl
The original author @MasonProtter did agree. I hope that he would also maintain this part in Unitful after merging.

All tests did pass!
I just got this warning (not sure if it is related to the addition):

```
┌ Warning: Symbol m was found in multiple registered unit modules.
│ We will use the one from Main.ShadowUnits.
└ @ Unitful ~/Code/Unitful.jl/src/user.jl:589
```

I did also build with Documenter without errors.

You are free to adjust things.

Would solve https://github.com/PainterQubits/Unitful.jl/issues/37

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -3,11 +3,11 @@ using Documenter, Unitful, Dates
 DocMeta.setdocmeta!(Unitful, :DocTestSetup, :(using Unitful))
 
 makedocs(
-    sitename = "Unitful.jl",
-    format = Documenter.HTML(prettyurls = get(ENV, "CI", nothing) == "true"),
-    modules = [Unitful],
-    workdir = joinpath(@__DIR__, ".."),
-    pages = [
+    sitename="Unitful.jl",
+    format=Documenter.HTML(prettyurls=get(ENV, "CI", nothing) == "true"),
+    modules=[Unitful],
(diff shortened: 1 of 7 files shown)


sostock:
I don’t think this functionality needs to live in Unitful.jl. I think it would be better to just add a link to NaturallyUnitful.jl to the README.

Mo8it:
You could have said that on the issue before I invest time in it. But fine, I will add NaturallyUnitful.jl to the README.

sostock:
In the issue I said (https://github.com/PainterQubits/Unitful.jl/issues/480#issuecomment-918859126)
> Of course we can mention it in the README. Someone just needs to make a PR that adds it to the list.

Sorry if that wasn’t clear enough. I didn’t know that you intended to move all of NaturallyUnitful into Unitful.

PR #482, opened 2021-09-14
Update .gitignore

Mo8it:
Better structure from the [Julia .gitignore template of Github](https://github.com/github/gitignore/blob/master/Julia.gitignore).


.gitignore
```diff
@@ -1,8 +1,26 @@
...
 *.jl.mem
-deps/
+
+# System-specific files and directories generated by the BinaryProvider and BinDeps packages
+# They contain absolute paths specific to the host computer, and so should not be committed
+deps/deps.jl
+deps/build.log
+deps/downloads/
+deps/usr/
+deps/src/
```
sostock, line 14: This package doesn’t have any binary dependencies, so I don’t think these lines are necessary.

.gitignore
```diff
@@ -1,8 +1,26 @@
...
+
+# Build artifacts for creating documentation generated by the Documenter package
 docs/build/
-docs/Manifest.toml
```
sostock, line 7: Why remove this?

Mo8it: The line:
Manifest.toml

excludes all Manifest.toml files in every directory.
docs/Manifest.toml is therefore excluded as well.

.gitignore
Mo8it: We can remove them.

.gitignore
sostock: Ah, I did not know that.

PR #483, opened 2021-09-14
Update README.md

Mo8it:
Mentioned NaturallyUnitful.jl


sostock:
At first it seemed weird to me that UnitfulAtomic.jl is under “Units packages” and NaturallyUnitful.jl is under “Feature additions” even though both do similar things. But I guess it is correct since UnitfulAtomic.jl defines its own units and NaturallyUnitful.jl does not.

sostock (approved):

PR #484, opened 2021-09-16
Adapt tests for better inference in Julia 1.8

sostock:
For units and dimensions (but not for quantities), the result of exponentiation with rational numbers can be inferred on master, this adapts the tests.


sostock:
In Julia 1.7, we can use `@constprop :aggressive` to make all of them inferrable. I will upload a new PR to do that instead.

PR #500, opened 2021-11-24
make `norm` use `norm` rather than `abs`

oscardssmith:
Currently `norm(u"s"*1)` returns `1 s`. After it returns `1.0 s`. The reasoning for this change is to improve consistency with LinearAlgebra which makes the `norm` for `Integer` be the floating point equivalent.


oscardssmith:
@giordano can you take a look at this?

giordano:
I'm not even sure that method is needed at all

oscardssmith:
Thanks!

PR #501, opened 2021-11-25
support broadcasting on ranges

aplavin:


src/range.jl
```diff
@@ -103,6 +105,15 @@ broadcasted(::DefaultArrayStyle{1}, ::typeof(*), r::AbstractRange, x::AbstractQu
     broadcasted(DefaultArrayStyle{1}(), *, r, ustrip(x)) * unit(x)
 broadcasted(::DefaultArrayStyle{1}, ::typeof(*), x::AbstractQuantity, r::AbstractRange) =
     broadcasted(DefaultArrayStyle{1}(), *, ustrip(x), r) * unit(x)
+
+const BCAST_PROPAGATE_CALLS = Union{typeof(upreferred), typeof(ustrip), Units}
+broadcasted(::DefaultArrayStyle{1}, ::typeof(*), r::AbstractRange, x::Ref{<:Units}) = r * x[]
+broadcasted(::DefaultArrayStyle{1}, ::typeof(*), x::Ref{<:Units}, r::AbstractRange) = x[] * r
+broadcasted(::DefaultArrayStyle{1}, x::BCAST_PROPAGATE_CALLS, r::StepRangeLen) = StepRangeLen(x(r.ref), x(r.step), r.len, r.offset)
```
sostock, line 112: For typical `StepRangeLen`s, this will widen the eltype (because `typeof(r.ref)` is usually wider than `eltype(r)` when creating the range with `:` or `range`):
```julia
julia> r = (1f0:2f0:3f0)u"m"
(1.0:2.0:3.0) m

julia> eltype(r) # Quantity based on Float32
Quantity{Float32, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}

julia> ustrip.(r)
1.0:2.0:3.0

julia> eltype(ans) # Resulting range has eltype widened from Float32 to Float64
Float64
```

aplavin: Indeed, thanks for catching! See https://github.com/PainterQubits/Unitful.jl/pull/503

PR #504, opened 2021-12-01
ChainRules `rrule` Integration for Unitful

SBuercklin:
This intention of this PR is to implement the machinery within `Unitful.jl` to allow for autodiff over `Unitful.Quantity`s. Specifically, it should include the necessary `ChainRulesCore.jl` methods to provide some basic level of compatibility with `ChainRules`-based AD systems.

Before this PR, `Quantity`s would be reduced to `Tangent{Any}(val = ...)` which would break a lot of basic AD arithmetic. After this PR:
```julia
julia> Zygote.gradient((x,y) -> (x*W)/(y*μm)/ms, 3.0*W, 2.0*μm)
(0.5 W μm^-2 ms^-1, -0.75 W^2 μm^-3 ms^-1)

julia> Zygote.gradient((x,y) -> (x*ms + 9*y*ms)/μm, 2.0*W, 3.0*W)
(1.0 ms μm^-1, 9.0 ms μm^-1)
```

I've implemented an `rrule` for the `Quantity` constructor, `ProjectTo{Quantity}`, and arithmetic between `Number/Quantity` and `Units`, which are used to call the `Quantity` constructor. Generally speaking, projecting a `Quantity` to a `Number` involves projecting the **value** of the `Quantity` onto the number, and then propagating the units of the projecting `Quantity` onto the `Number`. This ensures the proper real/complex behavior is obeyed while maintaining correct units.

I wanted to get feedback before continuing to ensure that:

1. This is something other people have an interest in
2. It's being properly implemented and tested
3. Any necessary changes elsewhere can also be implemented

Testing is difficult as `ChainRulesTestUtils.jl` does not play nicely with `Unitful.jl` at the moment. I can manually test the `rrule`s and `ProjectTo`, but if there are any other testing approaches I'm open to ideas.

Regarding dependencies elsewhere, more full compatibility with `ChainRules.jl` needs [this PR](https://github.com/JuliaDiff/ChainRules.jl/pull/540) which relaxes the constraint over many of the rules from `Union{Real, Complex}` to just `Number`. This should give compatibility with `Quantity`s which subtype `Number`, but typically wrap `<:Real, <:Complex`.

Remaining work: 
* Fill out rules where it makes sense for operations between `Quantity`s and `Unit`s (right now only `*` and `/` are implemented)
* Implement `frule`s
* Testing


src/chainrules.jl
```diff
@@ -0,0 +1,34 @@
...
+    return unitful_x, uq_pullback
+end
+
+function ProjectTo(x::Quantity)
+    project_val = ProjectTo(x.val) # Project the literal number
+    return ProjectTo{typeof(x)}(; project_val = project_val)
+end
+
+function (projector::ProjectTo{<:Quantity})(x::Number)
+    new_val = projector.project_val(ustrip(x))
```
oxinabox, line 14: convention:
```suggestion
function (project::ProjectTo{<:Quantity})(x::Number)
    new_val = project.val(ustrip(x))
```

oxinabox (commented):
Seems fine to me

src/chainrules.jl
```diff
@@ -0,0 +1,34 @@
+function rrule(UT::Type{Quantity{T,D,U}}, x::Number) where {T,D,U}
+    unitful_x = Quantity{T,D,U}(x)
+    projector_x = ProjectTo(x)
+    uq_pullback(Δx) = (NoTangent(), projector_x(Δx) * oneunit(UT))
+    return unitful_x, uq_pullback
+end
+
+function ProjectTo(x::Quantity)
+    project_val = ProjectTo(x.val) # Project the literal number
+    return ProjectTo{typeof(x)}(; project_val = project_val)
```
mcabbott, line 10: This stores the complete type, and hence unit, of `x` in the type of the projector. But when this is applied, you use only `unit(dx)` and not this unit. That's mathematically correct, I think, since the gradient will typically have different units. But it also means that storing this is redundant. 

It could just be `ProjectTo{Quantity}`. But could it just be `ProjectTo(x.val)`?

Right now `ProjectTo{Float64}` will allow through any `dx` with units, without changing them. To get the present behaviour of this PR, could you just define methods for `(::ProjectTo{<:Number})(dx::Quantity)` which un-wrap, adjust the precision etc if necessary, and re-wrap? Or does this land you in dispatch ambiguity hell?

oxinabox: oh true, i missed that, my bad

src/chainrules.jl
```diff
@@ -0,0 +1,34 @@
...
+(project::ProjectTo{<:Real})(dx::Quantity) = project(ustrip(dx))*unit(dx)
+(project::ProjectTo{<:Complex})(dx::Quantity) = project(ustrip(dx))*unit(dx)
+
+function rrule(::typeof(*), x::Quantity, y::Units, z::Units...)
+    Ω = *(x, y, z...)
+    project_x = ProjectTo(x)
+    function times_pb(Δ)
+        δ = project_x(Δ)
+        units = (y, z...)
+        return (NoTangent(), *(δ, y, z...), ntuple(_ -> NoTangent(), length(units))...)
```
oxinabox, line 28: Why not
```suggestion
        return (NoTangent(), *(δ, y, z...), ntuple(_ -> NoTangent(), 1+length(z))...)
```

src/chainrules.jl
```diff
@@ -0,0 +1,34 @@
+function rrule(UT::Type{Quantity{T,D,U}}, x::Number) where {T,D,U}
+    unitful_x = Quantity{T,D,U}(x)
+    projector_x = ProjectTo(x)
+    uq_pullback(Δx) = (NoTangent(), projector_x(Δx) * oneunit(UT))
+    return unitful_x, uq_pullback
+end
```
mcabbott, line 6: When is this called? 

If't it's used when attaching units to an initially plain number, `x=1`  -> `unitful_x = 1m`, then the thinking is that if the loss is a unitless scalar, the gradient for `unitful_x ` will be `d loss / d unitful_x = 100/m`, and this will produce a gradient for `x` with no units (or units equivalent to 1)?

oxinabox:
I am trying to think  this one though.

Consider: `x = (5.0Meter/Second) * t)`
or written long:
```julia
function f(t)
    a = 5 Meter/Second
    x = a*t
    return x
end
```


so `dx/dt = 5.0 Meter/Second`
this is known

I initially assumed (incorrectly) that  the seed co-tangent has the same units as a difference of primals, which is same as the primal units in most (all?) cases
which would have ` t̄ = dt/dt = 1 Second`

Then we would get
`x̄ = a * t̄ `
`x̄ = (5.0 Meter/Second) * 1 Second`
`x̄ = (5.0 Meter)`
 Which is wrong since `x̄ = dx/dt`

So I guess the seed must have units of `Second/Second` or just `1`, makes sense since it is `dt/dt`

 ` t̄ = dt/dt = 1 Second/Second`
`x̄ = a * t̄ `
`x̄ = (5.0 Meter/Second) * 1 Second/Second`
`x̄ = (5.0 Meter/Second)`

Which was what was wanted.

src/chainrules.jl
mcabbott: Was going to suggest something like:
```
        nots = ntuple(Returns(NoTangent()), 1 + length(z))
        return (NoTangent(), *(ProjectTo(x)(Δ), y, z...), nots...)
```
since I think there is little to gain by making the pullback close over the ProjectTo instead of over `x`. But something to gain in readability by needing fewer symbols.

src/chainrules.jl
SBuercklin: `Zygote.sensitivity` returns the multiplicative identity which is usually `1.0`, even for `Unitful.Quantity`. The example worked out in @oxinabox's comment matches what I've done, so I think `Zygote` is correct here.

platawiec:
Are there any further steps required before this can get merged? Should there be a manual rrule implemented for the tests, maybe?

SBuercklin:
Added some tests for the `rrule`s. 

I left the `/` rules untested since the calls themselves dispatch to `*(::Number, ::Units)` which in turn dispatch to the `Quantity` constructor rule. We could write these dispatches out manually and test them, but this seems like something the AD backend should be inferring. 

`frule`s aren't implemented still, but it seems fine to put that off to a later issue if we want to get this merged. I removed the WIP prefix from the PR title, and this seems good to go if we're happy with just `rrule`s for now.

jgreener64:
Is there anything else that needs doing or can this be merged?

SBuercklin:
This only implements the `rrule`s right now, which I think should be sufficient for most AD usage. In my opinion it's fine to merge, but I'll update the PR title to emphasize that it only has `rrule`s, not `frules`

Relatedly, @oxinabox do you think we need the `frule`s right now?

SBuercklin:
Accidentally bumped the patch version before; should be good now

giordano:
This is overall ok with me, but the question is do we want to add another dependency?  This package has been traditionally rather conservative on taking on dependencies, however there is already a non-standard library and `ChainRulesCore` is widely used with [over 2k total dependents](https://juliahub.com/ui/Packages/ChainRulesCore/G6ax7?page=2), so the probability someone would have already this package in an environment is non-negligible.  Also, I verified loading time of `Unitful.jl` doesn't increase dramatically.

@ajkeller34 @sostock opinions?

amostof:
I am awfully new to Julia and SciML ecosystem. Would this addition make possible to run `sciml_train` from `DiffEqFlux` packages on `DifferentialEquations` models for parameter estimation?

oxinabox:
> Also, I verified loading time of `Unitful.jl` doesn't increase dramatically.

It should now be even lighter
https://github.com/JuliaDiff/ChainRulesCore.jl/pull/524

SBuercklin:
Is there anything left to do on this PR?

Issue #127, opened 2018-03-20
Addition of heterogeneous arrays


KristofferC:
```
b = [0.0, 0.0m]
b + b
b .+ b
```

Both of these fails. If `b` is declared as `Any` it works. Is would be nice if this worked.

Stacktrace:

```
TypeError: setindex!: in typeassert, expected Unitful.Quantity{Float64,Unitful.Dimensions{()},Unitful.FreeUnits{(),Unitful.Dimensions{()}}}, got Float64

Stacktrace:
 [1] setindex!(::Array{Unitful.Quantity{Float64,Unitful.Dimensions{()},Unitful.FreeUnits{(),Unitful.Dimensions{()}}},1}, ::Unitful.Quantity{Float64,Unitful.Dimensions{()},Unitful.FreeUnits{(),Unitful.Dimensions{()}}}, ::Int64) at ./array.jl:583
 [2] collect_to_with_first! at ./array.jl:494 [inlined]
 [3] collect(::Base.Generator{UnitRange{Int64},Base.Broadcast.##1#2{Tuple{Array{Unitful.Quantity{Float64,D,U} where U where D,1},Array{Unitful.Quantity{Float64,D,U} where U where D,1}},Tuple{Tuple{Bool},Tuple{Bool}},Tuple{Tuple{Int64},Tuple{Int64}},CartesianIndex{1}}}) at ./array.jl:476
 [4] broadcast_t(::Function, ::Type{Any}, ::Tuple{Base.OneTo{Int64}}, ::CartesianRange{CartesianIndex{1}}, ::Array{Unitful.Quantity{Float64,D,U} where U where D,1}, ::Array{Unitful.Quantity{Float64,D,U} where U where D,1}) at ./broadcast.jl:258
 [5] broadcast_c at ./broadcast.jl:321 [inlined]
 [6] broadcast at ./broadcast.jl:455 [inlined]
 [7] +(::Array{Unitful.Quantity{Float64,D,U} where U where D,1}, ::Array{Unitful.Quantity{Float64,D,U} where U where D,1}) at ./arraymath.jl:39
```

ajkeller34:
Nice catch. This is partially fixed, in that it now works but doesn't give the most specific eltype possible:

```
julia> b = [0.0, 0.0u"m"]
2-element Array{Quantity{Float64,D,U} where U where D,1}:
   0.0
 0.0 m

julia> b + b
2-element Array{Number,1}:
     0.0
 0.0 m  
```

[...]

PR #509, opened 2021-12-24
Fix some major invalidations to improve compile times

ChrisRackauckas:
Unitful.jl was flagged as a major invalidator of compilation downstream. Test:

```julia
# From: https://timholy.github.io/SnoopCompile.jl/stable/snoopr/
using SnoopCompile
invalidations = @snoopr begin
    using DifferentialEquations

    function lorenz(du,u,p,t)
     du[1] = 10.0(u[2]-u[1])
     du[2] = u[1]*(28.0-u[3]) - u[2]
     du[3] = u[1]*u[2] - (8/3)*u[3]
    end
    u0 = [1.0;0.0;0.0]
    tspan = (0.0,100.0)
    prob = ODEProblem(lorenz,u0,tspan)
    alg = Rodas5()
    tinf = solve(prob,alg)
end;

trees = SnoopCompile.invalidation_trees(invalidations);

@show length(SnoopCompile.uinvalidated(invalidations)) # show total invalidations

show(trees[end]) # show the most invalidated method
```

This method won the prize for the absolute most invalidations :tada:. But I think the bigger issue is that it simply doesn't follow Julia semantics. It fixes the types for issue https://github.com/PainterQubits/Unitful.jl/issues/127 in a way that gives a stricter type than Julia would do in the normal cases (which is why the invalidation occurs).

After this PR, heterogeneous arrays of numbers with Quantity in there act normally, and compile times are back to normal. Here's a showcase of it being normal:

```julia

using Unitful, Test
m = u"m"
cm = u"cm"

b = Union{Complex,Float64}[0 + 0im, 0.0]
@test b + b == b
@test b .+ b == b
@test eltype(b+b) === Number

b = Number[0 + 0im, 0.0]
@test b + b == b
@test b .+ b == b
@test eltype(b+b) === Number

b = [0.0, 0.0m]
@test b + b == b
@test b .+ b == b
@test eltype(b+b) === Number
```

### DIFF
--- a/src/promotion.jl
+++ b/src/promotion.jl
@@ -115,5 +115,3 @@ Base.promote_rule(::Type{Quantity{T}}, ::Type{Quantity{S,D,U}}) where {T,S,D,U} 
 
 Base.promote_rule(::Type{Quantity{S,D,U}}, ::Type{Quantity{T}}) where {T,S,D,U} =
     Quantity{promote_type(T,S)}
-
-Base.promote_typejoin(::Type{Quantity{S,D,U}}, ::Type{Quantity{T,D,U}}) where {T,S,D,U} = Quantity{T,S}
--- a/src/promotion.jl
+++ b/src/promotion.jl
@@ -116,4 +116,4 @@ Base.promote_rule(::Type{Quantity{T}}, ::Type{Quantity{S,D,U}}) where {T,S,D,U} 
 Base.promote_rule(::Type{Quantity{S,D,U}}, ::Type{Quantity{T}}) where {T,S,D,U} =
     Quantity{promote_type(T,S)}
 
-Base.promote_typejoin(::Type{T}, ::Type{Quantity{T,D,U}}) where {T,D,U} = Quantity{T}
+Base.promote_typejoin(::Type{Quantity{S,D,U}}, ::Type{Quantity{T,D,U}}) where {T,S,D,U} = Quantity{T,S}
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1263,7 +1263,7 @@ end
             b = [0.0, 0.0m]
             @test b + b == b
             @test b .+ b == b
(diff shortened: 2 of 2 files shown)


ChrisRackauckas:
It would be good if this could get a quick review because it effects a lot of downstream compile times @sostock

src/promotion.jl
```diff
@@ -116,4 +116,4 @@ Base.promote_rule(::Type{Quantity{T}}, ::Type{Quantity{S,D,U}}) where {T,S,D,U}
 Base.promote_rule(::Type{Quantity{S,D,U}}, ::Type{Quantity{T}}) where {T,S,D,U} =
     Quantity{promote_type(T,S)}
 
-Base.promote_typejoin(::Type{T}, ::Type{Quantity{T,D,U}}) where {T,D,U} = Quantity{T}
+Base.promote_typejoin(::Type{Quantity{S,D,U}}, ::Type{Quantity{T,D,U}}) where {T,S,D,U} = Quantity{T,S}
```
giordano, line 119: I don't think `Quantity{T,S}` is correct: the second parameter should be the dimension, not another numeric type.  However, tests as you modified them pass for me without this method at all

giordano (changes requested):

giordano:
I'd argue it's a bit weird that in
```julia
julia> b = [0.0, 0.0u"m"]
2-element Vector{Quantity{Float64}}:
   0.0
 0.0 m

julia> b + b
2-element Vector{Number}:
     0.0
 0.0 m
```
`b` and `b + b` have different eltypes (I think that the proper eltype in both cases should be `Number`), but this is probably outside of the scope of this PR.

src/promotion.jl
ChrisRackauckas: Yeah, the better thing might be to completely remove this overload.

sostock (approved):

giordano (approved):

PR #511, opened 2022-01-03
Remove unnecessary code related to encoding of μ

sostock:
Since Julia 0.6 (https://github.com/JuliaLang/julia/pull/19464), Julia treats µ (U+00B5) and μ (U+03BC) as the same character in identifiers. However, parts of this package are older than that behavior and there is some (now unnecessary) code, e.g., to define both `µ<unit>` and `μ<unit>`. This PR removes the unnecessary code, adds tests to make sure the behavior doesn’t regress with newer Julia versions, and uses U+03BC everywhere. (I chose U+03BC over U+00B5 because it was used more often. It is also the character that is produced by `\mu<tab>`.)

The PR also replaces ɛ (U+025B) by ε (U+03B5) where it occurs. These are also treated as identical (in identifiers) by Julia.


test/runtests.jl
```diff
@@ -1849,6 +1849,25 @@ end
     end
 end
 
+@testset "Encoding" begin
```
sostock, line 1852: I’m not convinced we actually _need_ to test this. We could just remove the tests.

test/runtests.jl
```diff
@@ -1849,6 +1849,25 @@ end
...
+    @test uparse("µm") === uparse("μm")
+    @test uparse("dBµV") === uparse("dBμV")
+    @test @doc(Unitful.µm) == @doc(Unitful.μm)
+    # Julia treats ɛ (U+025B) and ε (U+03B5) as the same
+    @test Unitful.ɛ0 === Unitful.ε0
+    @test u"ɛ0" === u"ε0"
+    @test uparse("ɛ0") === uparse("ε0")
+    @test @doc(Unitful.ɛ0) == @doc(Unitful.ε0)
+end
+
```
sostock, line 1870: I’m not convinced we _need_ these tests. We could just remove them.

giordano (approved):
Reading the diff of this PR is...interesting :mag: 

I think tests are fine, especially in the unlikely case anything breaks upstream in Julia

PR #513, opened 2022-01-12
Fix printing of `StepRangeLen` with complex elements

sostock:
`StepRangeLen` supports complex values, but printing them errors because the `show` method creates another range in a way that is not supported for complex values. This PR fixes that.

Before:
```
julia> range((1+2im)u"m", step=1u"m", length=5)
(Error showing value of type StepRangeLen{Quantity{Complex{Int64}, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}, Quantity{Complex{Int64}, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}, Quantity{Complex{Int64}, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}, Int64}:
ERROR: MethodError: no method matching isless(::Complex{Int64}, ::Complex{Int64})
[...]
```

After:
```
julia> range((1+2im)u"m", step=1u"m", length=5)
(1 + 2im:1 + 0im:5 + 2im) m
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Unitful"
 uuid = "1986cc42-f94f-5a68-af5c-568840ba703d"
-version = "1.10.2"
+version = "1.10.1"
 
 [deps]
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Unitful"
 uuid = "1986cc42-f94f-5a68-af5c-568840ba703d"
-version = "1.10.1"
+version = "1.10.2"
 
 [deps]
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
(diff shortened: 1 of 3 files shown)


sostock:
Note that this also changes the printing of `StepRangeLen`s with elements that are real but not floating-point types: After this PR the step is printed even when it is `1` (just like for non-Unitful `StepRangeLen`s):

Before:
```julia
julia> StepRangeLen(1,1,5)*u"m"       # non-floating-point StepRangeLen is printed without step
(1:5) m

julia> StepRangeLen(1.0,1.0,5)*u"m"   # floating-point StepRangeLen is printed with step
(1.0:1.0:5.0) m
```

After:
```julia
julia> StepRangeLen(1,1,5)*u"m"       # StepRangeLen is always printed with step
(1:1:5) m

julia> StepRangeLen(1.0,1.0,5)*u"m"
(1.0:1.0:5.0) m
```

We could avoid this by using the old method for real-valued `StepRangeLen`s, but I think it is not worth the additional complexity, since `StepRangeLen` is usually not used for non-floating-point values (`StepRange` is used instead).

giordano (approved):
:+1:

PR #514, opened 2022-01-13
Fix `range` implementation on Julia master and resolve method ambiguities

sostock:
This fixes our `range` implementation on Julia `master` (it relied on the internal function `Base._rangestyle` which was removed). It also resolves the method ambiguities for the `_range` function.

~~One thing that isn’t supported yet is `range(; start, stop)` (and `start:stop`) with quantities (the `step` should be 1, so it can only work with dimensionless quantities). However, `range(; start, length)` and `range(; stop, length)` work with dimensionless quantities (with the step defaulting to 1).~~ `range(start, stop)` now works for dimensionless quantities.

There is an inconsistency which existed before and isn’t fixed by this PR: In `range(start; step, length)`, the resulting unit does not depend on the unit of `step` (only `start`), but in `range(start; step, stop)`, it does depend on the units of `step` (and `start` and `stop`):
```julia
julia> range(1u"cm", step=1u"mm", stop=2u"cm") # `cm` and `mm` are promoted to `m`
(1//100:1//1000:1//50) m

julia> range(1u"cm", step=1u"mm", length=11)   # `cm` is used, unit of `step` does not matter
(1//1:1//10:2//1) cm
```

### DIFF
--- a/src/range.jl
+++ b/src/range.jl
@@ -103,7 +103,7 @@ end
 OrderStyle(::Type{<:AbstractQuantity{T}}) where T = OrderStyle(T)
 ArithmeticStyle(::Type{<:AbstractQuantity{T}}) where T = ArithmeticStyle(T)
 
-(colon(start::T, step::T, stop::T) where T <: Quantity{<:Real}) =
+colon(start::T, step::T, stop::T) where {T<:Quantity{<:Real}} =
     _colon(OrderStyle(T), ArithmeticStyle(T), start, step, stop)
 _colon(::Ordered, ::Any, start::T, step, stop::T) where {T} = StepRange(start, step, stop)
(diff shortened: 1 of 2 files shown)


sostock:
https://github.com/JuliaLang/julia/pull/43790 doesn’t help. It looks like we need to update our `range` implementation for Julia 1.8 (it currently relies on the internal function `Base._rangestyle` which was removed).

I think we can still merge this, since the package is broken on master anyway (with or without this PR). I will look into updating our `range` implementation.

sostock:
I now updated our `range` implementation.

One thing that isn’t supported yet is `range(; start, step)` (and `start:step`) with quantities (the step should be `1`, so it can only work with dimensionless quantities). However, `range(; start, length)` and `range(; stop, length)` work with dimensionless quantities (with the step defaulting to 1).

giordano (approved):

PR #516, opened 2022-02-01
Enable `zero` for arrays with non-concrete eltype

sostock:
Replaces #472.

This enables `zero(::AbstractArray{T})` for non-concrete `T<:AbstractQuantity` by calling `zero` on each element *if the element is defined* and calling `zero(T)` otherwise.

Support for `Union{T,Missing}` with *concrete* `T<:AbstractQuantity` is added as well (this is only necessary on Julia < 1.8, on master it works without specialized method).

Support for `Union{T,Missing}` with *non-concrete* `T<:AbstractQuantity` is not added (tests for it are added with `@test_broken`) because `zero(::AbstractArray{Union{T,Missing}}) where T<:AbstractQuantity` only matches concrete `T` and `zero(::AbstractArray{T}) where T<:Union{AbstractQuantity,Missing}` would be type piracy.

### DIFF
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -404,7 +404,7 @@ function zero(x::AbstractArray{T}) where T<:AbstractQuantity
         dest
     end
 end
-@static if VERSION < v"1.8.0-DEV"
+@static if VERSION < v"1.8.0-DEV.107"
     function zero(x::AbstractArray{Union{T,Missing}}) where T<:AbstractQuantity # only matches _concrete_ T ...
         @assert isconcretetype(T) # ... but check anyway
         z = zero(T)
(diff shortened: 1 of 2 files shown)


src/quantities.jl
```diff
@@ -388,6 +388,30 @@ zero(x::Type{<:AbstractQuantity{T,D}}) where {T,D} = zero(T) * upreferred(D)
...
+            if isassigned(x, i...)
+                dest[i] = zero(x[i])
+            else
+                dest[i] = zero(T)
+            end
+        end
+        dest
+    end
+end
+@static if VERSION < v"1.8.0-DEV"
```
giordano, line 407: Nitpicking, I believe `@static` is useless for top-level expressions: the false branch disappears anyway after precompilation.

giordano (approved):

src/quantities.jl
sostock: That makes sense. I would still leave it there for consistency, since other parts of the codebase use top-level `@static` as well (probably because they were written by me and I didn’t know better). We can remove them at some point. Does that sound good to you?

giordano: Yes, it doesn't hurt :+1:

PR #539, opened 2022-05-18
Add `modf`

rashidrafeek:
Have added support for `modf`.

Currently `modf` doesn't work:
```julia
julia> modf(2.5u"Å")
ERROR: MethodError: no method matching rem(::Quantity{Float64, 𝐋, Unitful.FreeUnits{(Å,), 𝐋, nothing}}, ::Float64)
Closest candidates are:
  rem(::Any, ::Any, ::RoundingMode{:ToZero}) at ~/Apps/julia/julia-1.7/share/julia/base/div.jl:80
  rem(::Any, ::Any, ::RoundingMode{:Down}) at ~/Apps/julia/julia-1.7/share/julia/base/div.jl:81
  rem(::Any, ::Any, ::RoundingMode{:Up}) at ~/Apps/julia/julia-1.7/share/julia/base/div.jl:82
  ...
Stacktrace:
 [1] modf(x::Quantity{Float64, 𝐋, Unitful.FreeUnits{(Å,), 𝐋, nothing}})
   @ Base.Math ./math.jl:892
 [2] top-level scope
   @ REPL[3]:1
```

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -631,6 +631,8 @@ end
         @test mod2pi(360°) === 0°           # 2pi is 360°
         @test mod2pi(0.5pi*u"m/dm") ≈ pi    # just testing the dimensionless fallback
         @test modf(2.5rad) === (0.5, 2.0)
+        @test modf(-250cm/m) === (-1//2, -2//1)
+        @test_throws MethodError modf(1m)
         @test @inferred(inv(s)) === s^-1
         @test inv(ContextUnits(m,km)) === ContextUnits(m^-1,km^-1)
         @test inv(FixedUnits(m)) === FixedUnits(m^-1)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -630,6 +630,7 @@ end
         @test mod(1hr+3minute+5s, 24s) == 17s
         @test mod2pi(360°) === 0°           # 2pi is 360°
         @test mod2pi(0.5pi*u"m/dm") ≈ pi    # just testing the dimensionless fallback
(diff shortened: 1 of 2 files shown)


wnoise:
This _should_ fail for dimensional units.  I would say adding it to the autogenerated unitless pass-through functions like log, sin, etc. on  [line 208 of quantities.jl](https://github.com/PainterQubits/Unitful.jl/blob/960e09ef518b0d3ff54747460b9c3872c08df0d7/src/quantities.jl#L208-L209) makes sense.

rashidrafeek:
You are right. I have changed it now. 

The original motivation for this came when I got a dimension error while passing dimensional input to `modf`. I realized subsequently my usage was wrong and I should have passed dimesion-less quantities to `modf`.

test/runtests.jl
```diff
@@ -630,6 +630,7 @@ end
         @test mod(1hr+3minute+5s, 24s) == 17s
         @test mod2pi(360°) === 0°           # 2pi is 360°
         @test mod2pi(0.5pi*u"m/dm") ≈ pi    # just testing the dimensionless fallback
+        @test modf(2.5rad) === (0.5, 2.0)
```
sostock, line 633: It would be good to add a test with a dimensionless unit that isn’t equal to 1, and another test that checks that an error is thrown for a dimensionful quantity:
```suggestion
        @test modf(2.5rad) === (0.5, 2.0)
        @test modf(-250cm/m) === (-1//2, -2//1)
        @test_throws MethodError modf(1m)
```

PR #551, opened 2022-08-30
Fix printing of StepRangeLen with affine units

tomyun:
This fixed a printing issue (#550) with `range` structures (`StepRange`, `StepRangeLen`) defined with `°C` as a primary unit and `K` as a step unit.

Before:
```
julia> StepRange(1u"°C", 1u"K", 3u"°C")
(1//1:-5443//20:3//1) °C

julia> StepRange(1u"°C", 1.0u"K", 3u"°C")
(1.0:-272.15:273.15) °C

julia> StepRangeLen{typeof(1u"°C"),typeof(1u"°C"),typeof(1u"K")}(1u"°C", 1u"K", 3, 1)
(1//1:-5443//20:-5433//10) °C

julia> StepRangeLen{typeof(1.0u"°C"),typeof(1.0u"°C"),typeof(1u"K")}(1.0u"°C", 1u"K", 3, 1)
(1.0:-272.15:-543.3) °C
```

After:
```
julia> StepRange(1u"°C", 1u"K", 3u"°C")
(1:1:3) °C

julia> StepRange(1u"°C", 1.0u"K", 3u"°C")
(1.0:1.0:3.0) °C

julia> StepRangeLen{typeof(1u"°C"),typeof(1u"°C"),typeof(1u"K")}(1u"°C", 1u"K", 3, 1)
(1:1:3) °C

julia> StepRangeLen{typeof(1.0u"°C"),typeof(1.0u"°C"),typeof(1u"K")}(1.0u"°C", 1u"K", 3, 1)
(1.0:1.0:3.0) °C
```

PR #553, opened 2022-09-04
Remove support for Julia <1.6

sostock:
The tests that are added in #551 reveal that some things don’t work correctly on Julia 1.0.

Julia 1.6 is the LTS release (since the release of Julia 1.7.0 on 2021-11-30), so support for older versions can be removed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         julia-version:
-          - "1.0"
+          - "1.6"
           - "1"
           - "nightly"
         os:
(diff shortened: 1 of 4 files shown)


giordano (approved):

giordano:
Should we make a release before merging this?  There are some changes since last version

sostock:
I am _slightly_ in favor of releasing a new version before merging this, because then people on Julia 1.0 still get #530 (which one could call a bugfix). The other changes since v1.11.0 are feature additions, so IMO it doesn’t make a difference whether they are released for Julia 1.0 or not. What do you think about this?

giordano:
Sounds good to me.  I just noticed that in #533 I put some `if isdefined` fences instead of `if VERSION`, those should all be true for Julia >= v1.6, so the fences can be safely removed

sostock:
> I just noticed that in #533 I put some `if isdefined` fences instead of `if VERSION`, those should all be true for Julia >= v1.6, so the fences can be safely removed in this PR

I just removed them.

sostock:
Superseded by #804

Issue #556, opened 2022-09-12
Unitful.q not under consts in file


KronosTheLate:
In the file defining constants, the elementary charge is not defined under "constants".
![image](https://user-images.githubusercontent.com/61620837/189643733-47cadf7f-327c-4eee-a868-be8fd00ed1ab.png)

<img width="384" alt="image" src="https://user-images.githubusercontent.com/61620837/189643718-90ad1f2f-26b5-4ab5-a1df-52b7f9ba680b.png">

This confused me. Any reason why?

Issue #545, opened 2022-07-20
Initialization problems with `preferunits` for physical constants on Julia 1.8


j-fu:
Hi, I struggle with a subtle issue concerning the definition of physical constants in unitful.jl after preferunits has been set.

I set up  a small package on  https://github.com/j-fu/UTest.jl which can be dev'd for testing.

In the root directory it contains `utest.jl` which can be called either with
```
julia --project=. --compile=min utest.jl`
```
or with
```
julia --project=. utest.jl`
```

The correct output is
```
Unitful.upreferred(1 * u"m") = 100 cm
ustrip(upreferred(u"ε0")) = 8.85418781762039e-18
ustrip(upreferred(Unitful.ε0)) = 8.85418781762039e-18
ustrip(upreferred(PhysicalConstants.CODATA2018.ε_0)) = 8.8541878128e-18
```
This is obtained on all older Julia versions, and on 1.8 (beta, rc3) with `--compile=min`

On  1.8 (beta, rc3) without `--compile=min` one gets
```
Unitful.upreferred(1 * u"m") = 100 cm
ustrip(upreferred(u"ε0")) = 8.85418781762039e-12
ustrip(upreferred(Unitful.ε0)) = 8.85418781762039e-12
ustrip(upreferred(PhysicalConstants.CODATA2018.ε_0)) = 8.8541878128e-18
```

One gets the correct result with optimized code if after modification, the `--compile=min` has been called first.  The problem occurs only for physical constants, not for units. PhysicalConstants.jl behaves in the right way, though.

I see three possibilities here:
- Problem with the architecture of Unitful
- Problem with code loading in Julia 1.8 when optimizing
- Evaluating upreferred into the global context in a package using Unitful is a bad idea, even if `preferunits` is called before loading the package. If this is the case, it needs to be documented. (in fact I struggled with this situation in more obvious situations, so I have an Idea what to write into a PR...)

j-fu:
May be also linked to https://github.com/JuliaLang/julia/pull/45861

j-fu:
Update: the problem persists in 1.8.0-rc4

sostock:
Now that I’ve looked at this, I’m surprised that it ever worked. The problem is here:
https://github.com/PainterQubits/Unitful.jl/blob/dc044eeed30405a4b87be569d5eb1f291e97e795/src/units.jl#L304-L307
This is a generated function that depends on non-constant global state (`Unitful.promotion`), which is not allowed.

[...]

PR #558, opened 2022-09-12
Move const `q` under `Constants`

KronosTheLate:
This PR moves the definition of the constant `q` under "Constants". See #556.

I also changed the comment on the line defining `q` from "`e` means 2.718..." to "avoid naming conflict with `e`", as I feel the original phrasing could be confused with the `e` in the line that means "times 10 to the power of..."

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -334,13 +334,6 @@ end
 @unit M      "M"        Molar       1mol/L                  true true
 
 # Energy
-"    Unitful.q
-\nA quantity equal to the elementary charge, the charge of a single electron,
-with a value of exactly 1.602,176,634 × 10^-19 C. The letter `q` is used instead of `e` to avoid
-confusion with Euler's number.
-\nDimension: 𝐈 𝐓.
-\nSee Also: [`Unitful.C`](@ref)."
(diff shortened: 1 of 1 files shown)


sostock:
You only moved the definition of `q`. The definition of `eV` needs to be moved as well, since it uses `q`.

KronosTheLate:
Nice catch @sostock. However, eV is not a constant, but a unit. So I will move it to _under_ the "constants" section, right?

j-fu:
Once you are working on the constants: are you aware of #545 ?

KronosTheLate:
not at all

sostock:
> However, eV is not a constant, but a unit. So I will move it to _under_ the "constants" section, right?

IMO, you can also move it into the constants section (right below `q`), since there are other units that are defined in that section as well (`c`, `u`).

KronosTheLate:
That is not an argument to move `eV` into the constants section, but rather `c` and `u` out of it IMO.

sostock:
That’s fine too, although I like that the units are defined immediately after the constants on which they are based.

KronosTheLate:
There is sone appeal, but I personally look up the variables for constants all the time, and love having them all in one place. So that has more weight to me, as I frequently make use of it.

PR #567, opened 2022-09-29
create Logo

lmiq:
I've noticed that the package does not have a logo, so I created one. Feel free to ignore :-) . 

The image is a pictorial representation of the standard kilogram: https://www.researchgate.net/profile/Beat-Jeckelmann/publication/266067006/figure/fig1/AS:669392329244682@1536606976525/The-Swiss-official-kilogram-prototype-No-38.png

The docs should look like this now:

image

I also tried to add the logo to the README.md file, but it didn't look quite nice, because of the alignment. Thus I removed it, I'm not sure if the changes in that file will appear in this PR.


sostock:
I like it! @giordano What do you think?

Maybe we could add a note to the frontpage of the docs, explaining what the logo shows (I wouldn’t have known without the explanation).

docs/src/index.md
```diff
@@ -118,3 +118,8 @@ A^2 s^4 kg^-1 m^-3
     temperature differences.
 
 See `test/runtests.jl` for more usage examples.
+
+## About the logo
+
+The logo is a pictorial representation of the [International Prototype of the Kilogram](https://en.wikipedia.org/wiki/International_Prototype_of_the_Kilogram), which was the standard definition of one kilogram from 1889 to 2019,
+when it was replaced by a definition based on Planck constant, the speed of light and hyperfine transition frequency of <sup>133</sup>Cs.
```
sostock, line 125:
```suggestion
when it was replaced by a definition based on the Planck constant, the speed of light, and the ground-state hyperfine transition frequency of ¹³³Cs.
```
* Instead of `the Planck constant`, `Planck's constant` would also be possible.
* Instead of `the ground-state hyperfine transition`, the less specific `a hyperfine transition` would also be possible.
* Documenter.jl doesn’t render the HTML tags (at least on my computer, `<sup>133</sup>` just appears verbatim in the output), so I replaced it with unicode. We could also use the unabbreviated form `caesium-133` instead.

sostock (changes requested):

sostock (approved):

sostock:
Thanks!

Issue #375, opened 2020-08-28
Should `uconvert` (inconsistently) change its argument's precision?


cnaak:
In the following example, `uconvert` returns an object of different, `Float64`, precision from its `Float32`-based argument:

```julia
julia> dia = cbrt(1.0f0u"l")
1.0f0 L^1/3

julia> typeof(dia)
Quantity{Float32,𝐋,Unitful.FreeUnits{(L^1/3,),𝐋,nothing}}

julia> typeof(uconvert(Unitful.m, dia))
Quantity{Float64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}}
```

I expect unit conversions not to change the underlying precision, even more when `Unitful.jl`'s design include a precision parameter in its types.

Moreover, as a consistency note, other tests showed that the above mentioned behavior is inconsistent as well, thus making it hard to predict the exact return type of `uconvert` calls:

```julia
julia> oth = 1.0f0u"dm"
1.0f0 dm

julia> uconvert(Unitful.m, oth)
0.1f0 m

julia> typeof(uconvert(Unitful.m, oth))
Quantity{Float32,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}}
```

Apologies if this is a duplicate, I scanned through the opened issues and none seemed to fit.

cnaak:
A workaround is to explicitly convert `uconvert`'s result:

```julia
julia> using Unitful

julia> dia = cbrt(1.0f0u"l")
1.0f0 L^1/3

julia> Float32(uconvert(Unitful.m, dia))
0.1f0 m

julia> typeof(ans)
Quantity{Float32,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}}
```

cnaak:
I've found another, possibly related, issue with `uconvert` with `BigFloat` arguments:

```julia
julia> THE = BigFloat("-50.1") * u"°"
-50.09999999999999999999999999999999999999999999999999999999999999999999999999978°

julia> uconvert(u"rad", THE)
-0.8744099552491591032660211979532505210954695940017700195312499999999999999999983 rad

julia> BigFloat("-50.1") * BigFloat(π) / 180
-0.8744099552491591180387690750127949694348788161594044535046929115256755664163204
```

[...]

PR #571, opened 2022-10-09
Fix `convfact` documentation

JeffFessler:
This is a bit related to #375, i.e., I noticed the docs here were wrong when I also hit that issue and was trying to diagnose the problem.

### DIFF
--- a/src/conversion.jl
+++ b/src/conversion.jl
@@ -1,6 +1,6 @@
 """
     convfact(s::Units, t::Units)
-Find the conversion factor from unit `t` to unit `s`, e.g. `convfact(m,cm) = 1//100`.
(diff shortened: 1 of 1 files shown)


src/conversion.jl
```diff
@@ -1,6 +1,6 @@
 """
     convfact(s::Units, t::Units)
-Find the conversion factor from unit `t` to unit `s`, e.g. `convfact(m,cm) = 0.01`.
+Find the conversion factor from unit `t` to unit `s`, e.g. `convfact(m,cm) = 1//100`.
```
sostock, line 3:
```suggestion
Find the conversion factor from unit `t` to unit `s`, e.g. `convfact(m,cm) == 1//100`.
```
I would use `==` here, otherwise it looks like a function definiton.

JeffFessler: Agreed - thanks!

PR #573, opened 2022-10-19
Update README.md

briochemc:
Add deprecation note for UnitfulRecipes.jl (and move it to the bottom of the list).


sostock:
Thanks!

PR #578, opened 2022-11-02
Remove UnitfulIntegration.jl from README

### LINKED ISSUE #376: remove UnitfulIntegration from the readme
quadgk seems to work fine with Unitful alone, but breaks when also using UnitfulIntegration. Guessing UnitfulIntegration is deprecated? delete it from the front page

sostock:
UnitfulIntegration.jl is also no longer needed, QuadGK can handle Unitful quantities since QuadGK v2.1.0.

I decided to remove the entry (instead of declaring it deprecated like UnitfulRecipes.jl) because it cannot even be installed with Unitful 1.x (it requires Unitful 0.x), so it isn’t relevant to the package in its current state.

Closes #376.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,6 @@ mathematical operations and collections that are found in Julia base.
 
 ### Feature additions
 
-- [UnitfulIntegration.jl](https://github.com/PainterQubits/UnitfulIntegration.jl): Enables use of Unitful quantities with [QuadGK.jl](https://github.com/JuliaMath/QuadGK.jl). PRs for other integration packages are welcome.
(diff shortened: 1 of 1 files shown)


sostock:
I just noticed that infinite integration limits don’t work, so the statement that QuadGK supports Unitful quantities is not correct (infinite integration limits were supported by UnitfulIntegration.jl):
```julia
julia> quadgk(x->exp(-x/(1.0m)), 0.0m, Inf*m)
ERROR: DomainError with Inf m:
integrand produced NaN m in the interval (0.0 m, Inf m)
Stacktrace:
[...]
```

PR #591, opened 2022-11-25
Update `Types` documentation

### LINKED ISSUE #590: Documentation for `Types` outdated?
[The documentation for `Types`](https://painterqubits.github.io/Unitful.jl/stable/types/) says

> An abstract type [Unitful.Units{N,D}](https://painterqubits.github.io/Unitful.jl/stable/types/#Unitful.Units) is defined, where N is always a tuple of Unit objects, and D is a [Unitful.Dimensions{N}](https://painterqubits.github.io/Unitful.jl/stable/types/#Unitful.Dimensions) object such as 𝐋, the object representing the length dimension.

and 

>  Currently implemented subtypes of Unitful.Units{N,D} include [Unitful.FreeUnits{N,D}](https://painterqubits.github.io/Unitful.jl/stable/types/#Unitful.FreeUnits), [Unitful.ContextUnits{N,D,P}](https://painterqubits.github.io/Unitful.jl/stable/types/#Unitful.ContextUnits), and [Unitful.FixedUnits{N,D}](https://painterqubits.github.io/Unitful.jl/stable/types/#Unitful.FixedUnits).

but, for example 
```julia
julia> typeof(u"m")
Unitful.FreeUnits{(m,), 𝐋, nothing}
```
where `FreeUnits` has 3 fields (see trailing `nothing`). The same is true for other types, e.g. `Units`. 

Is the documentation outdated or am I missing something? Thanks!

cmichelenstrofer:
Fixes #590

### DIFF
--- a/docs/src/conversion.md
+++ b/docs/src/conversion.md
@@ -190,7 +190,7 @@ Sometimes, a package may want to default to a particular behavior for promotion,
 presence of other packages that may require differing default behaviors. An example would be
 a CAD package for nanoscale device design: it makes more sense to promote to nanometers or
 microns than to meters. For this purpose we define `Unitful.ContextUnits{N,D,P,A}`. The `P` in
-this type signature should be some type `Unitful.FreeUnits{M,D,A}` (the dimensions must be the
+this type signature should be some type `Unitful.FreeUnits{M,D,B}` (the dimensions must be the
 same). We refer to this as the "context." `ContextUnits` may be easily instantiated by e.g.
 `ContextUnits(nm, μm)` for a `nm` unit that will promote to `μm`. Here's an example:
 
--- a/docs/src/conversion.md
+++ b/docs/src/conversion.md
@@ -189,8 +189,8 @@ in prior versions of Unitful.
 Sometimes, a package may want to default to a particular behavior for promotion, in the
 presence of other packages that may require differing default behaviors. An example would be
 a CAD package for nanoscale device design: it makes more sense to promote to nanometers or
-microns than to meters. For this purpose we define `Unitful.ContextUnits{N,D,P}`. The `P` in
-this type signature should be some type `Unitful.FreeUnits{M,D}` (the dimensions must be the
+microns than to meters. For this purpose we define `Unitful.ContextUnits{N,D,P,A}`. The `P` in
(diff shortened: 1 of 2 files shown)


sostock:
There are still some places where the last type parameter is missing in conversion.md:
https://github.com/PainterQubits/Unitful.jl/blob/7dc9dd7cc3e88706ffa1ee14d9c660164d90a3b3/docs/src/conversion.md?plain=1#L192-L193
https://github.com/PainterQubits/Unitful.jl/blob/7dc9dd7cc3e88706ffa1ee14d9c660164d90a3b3/docs/src/conversion.md?plain=1#L248

sostock:
I looked through the manual, the three occurrences I pointed out should be the only remaining ones.

cmichelenstrofer:
Cool. I fixed those three lines. Thanks!

docs/src/conversion.md
```diff
@@ -189,8 +189,8 @@ in prior versions of Unitful.
 Sometimes, a package may want to default to a particular behavior for promotion, in the
 presence of other packages that may require differing default behaviors. An example would be
 a CAD package for nanoscale device design: it makes more sense to promote to nanometers or
-microns than to meters. For this purpose we define `Unitful.ContextUnits{N,D,P}`. The `P` in
-this type signature should be some type `Unitful.FreeUnits{M,D}` (the dimensions must be the
+microns than to meters. For this purpose we define `Unitful.ContextUnits{N,D,P,A}`. The `P` in
+this type signature should be some type `Unitful.FreeUnits{M,D,A}` (the dimensions must be the
```
sostock, line 193:
```suggestion
this type signature should be some type `Unitful.FreeUnits{M,D,B}` (the dimensions must be the
```
The last parameter does not have to match, so I would change `A` to `B` to reflect that (just like `M` was used instead of `N`).

We could also write something like `Unitful.FreeUnits{N2,D,A2}`, but I’m not sure whether that’s better or worse for understanding.

sostock (changes requested):

docs/src/conversion.md
cmichelenstrofer: Ah! Makes sense 👍🏾

sostock (approved):

PR #598, opened 2022-12-16
Add `randn` method

JeffFessler:
I can add a simple test if maintainers are open to this addition.

### DIFF
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -512,6 +512,8 @@ end
 
 Base.rand(r::Random.AbstractRNG, ::Random.SamplerType{<:AbstractQuantity{T,D,U}}) where {T,D,U} =
     rand(r, T) * U()
+Base.randn(r::Random.AbstractRNG, ::Type{<:AbstractQuantity{T,D,U}}) where {T,D,U} =
+    randn(r, T) * U()
 Base.ones(Q::Type{<:AbstractQuantity}, dims::NTuple{N,Integer}) where {N} =
     fill!(Array{Q,N}(undef, map(Base.to_dim, dims)), oneunit(Q))
 Base.ones(Q::Type{<:AbstractQuantity}, dims::Tuple{}) = fill!(Array{Q}(undef), oneunit(Q))


sostock:
With this implementation, `randn(typeof(1.0mm/m))` wouldn’t sample from a distribution with standard deviation `1` but with standard deviation `1mm/m`, which is equal to `0.001`. This seems wrong to me and contradicts the documented behavior of `randn`.

IMO, `randn(Quantity{…})` should only work for dimensionless quantity types and should sample from a distribution with standard deviation equal to `1`, not some other number `1*some_unit` that isn’t equal to `1`. Everything differing from this behavior has the potential to introduce bugs that are hard to find.

A possible implementation could be the following (not tested):
```julia
Base.randn(r::Random.AbstractRNG, q::Type{<:DimensionlessQuantity{T}}) where T = convert(q, randn(r, T))
```

One argument in favor of this PR is that `rand` already works like this for quantities. However, I think the behavior is less harmful in the case of `rand`, since `rand` draws from different distributions anyway, so one always has to be careful (for integer types `T` it picks from `typemin(T):typemax(T)`, for floating-point types from `[0,1)`). Therefore, I think the `rand` behavior is fine (but should definitely be documented!), whereas the proposed `randn` behavior isn’t.

JeffFessler:
Thanks for the detailed response.  I had not thought about `1mm/m`.  (To be honest I don't understand why that ratio is a Unitful type rather than just a `Float64`, but I'm sure there is a good reason.) I agree that there can be different interpretations of what the variance "1" means for a "standard normal" when units are in the picture, and that could lead to confusion.  So I'll just close this request.

giordano:
>  I had not thought about `1mm/m`. (To be honest I don't understand why that ratio is a Unitful type rather than just a `Float64`, but I'm sure there is a good reason.)

It isn't super easy to automatically simplify division, there are many discussions about this here in this repository.  Besides, autosimplification isn't even always desired, and this defeats the purpose of an automatic operation.  For example the [Hubble parameter](https://en.wikipedia.org/wiki/Hubble%27s_law) is typically measured in `(km/s)/Mpc`, even if that's equivalent to ~`1/s` (`km` and `Mpc` are both lengths).

bjarthur:
@JeffFessler i'm not sure this will help you, but i use `randn!(ustrip(x))` to get a std dev of 1 in whatever the units of `x` happen to be.

JeffFessler:
Thanks for the suggestion.
I have gone with using `oneunit(x) * randn()` for now because it avoids requiring dependence on Unitful package since `oneunit` is in Base.  It even works for `1mm/m`.  Perhaps I should have based my original PR on something like that but maybe not worth it anyway for such a simple work-around.

PR #602, opened 2023-01-05
Use `Core.eval` and not a random unexported function

vtjnash:


cstjean:
Not that I want to quibble about it, as I'm sure there's a good reason for the change, but I'm a bit surprised that this is called a "random unexported function". In:

```julia
module Foo
f(x) = eval(x)
end
```
isn't that `eval` the same function as `Foo.eval`? I thought that each module has, implicitly, `eval(x) = Core.eval(current_module, x)` defined.

PR #606, opened 2023-01-18
add Aqua and resolve some ambiguities

sjkelly:
This resolves several ambiguities detected by Aqua.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,9 +13,10 @@ ConstructionBase = "1"
 julia = "1"
 
 [extras]
+Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["LinearAlgebra", "Test", "Random"]
+test = ["Aqua", "LinearAlgebra", "Test", "Random"]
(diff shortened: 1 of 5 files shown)


Project.toml
```diff
@@ -13,9 +13,10 @@ ConstructionBase = "1"
 julia = "1"
 
 [extras]
+Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
```
sostock, line 16: Aqua should also have version bounds in the `[compat]` section.

test/runtests.jl
```diff
@@ -2117,3 +2117,7 @@ finally
     rm(load_path, recursive=true)
     rm(load_cache_path, recursive=true)
 end
+
+using Aqua
+
+Aqua.test_all(Unitful, ambiguities=false, unbound_args=false)
```
sostock, line 2123:
```suggestion
Aqua.test_all(Unitful, ambiguities=false, unbound_args=false, piracy=VERSION≥v"1.8")
```
On older Julia versions, the `zero(x::AbstractArray{Union{Missing, T}}) where T<:Unitful.AbstractQuantity` method is flagged for piracy. Therefore, I would suggest to only test for piracy on Julia ≥1.8.

src/types.jl
```diff
@@ -256,6 +256,7 @@ For example, given a gain of `20dB`, we have `val===20`. This type differs from
 struct Gain{L, S, T<:Real} <: LogScaled{L}
     val::T
 end
+Unitful.Gain{L, S, T}(x::Unitful.Gain{L, S, T}) where {L,S,T<:Real} = x # for method ambiguity on 1.9
```
sostock, line 259:
```suggestion
Unitful.Gain{L, S, T}(x::Unitful.Gain{L, S, T}) where {L,S,T<:Real} = x # for ambiguity resolution
```
The ambiguity exists on all 1.x versions, not only 1.9.

sostock:
Thanks!

PR #622, opened 2023-02-25
support InverseFunctions

aplavin:
Useful in, for example:
```julia
using Accessors

x = (a=1, b=1.23m)
y = @set x.b |> ustrip(m, _) = 2.34
# y == (a=1, b=2.34m)


using FlexiMaps

lr = maprange(@optic(log(ustrip(m, _))), 10cm, 10m, length=5)
# lr ≈ [0.1, √0.1, 1, √10, 10]m
```

Only works on 1.9+, no changes on earlier julias.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,8 +17,8 @@ julia = "1"
 InverseFunctionsUnitfulExt = "InverseFunctions"
 
 [extras]
-InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
+InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"


test/runtests.jl
```diff
@@ -85,6 +85,11 @@ const colon = Base.:(:)
     @test ConstructionBase.constructorof(typeof(1.0m))(2) === 2m
 end
 
+VERSION >= v"1.9-" && @testset "inverse" begin
+    InverseFunctions.test_inverse(Base.Fix1(ustrip, u"m"), 2u"m")
+    InverseFunctions.test_inverse(Base.Fix1(ustrip, u"m"), 2u"mm")
```
sostock, line 90:
```suggestion
    InverseFunctions.test_inverse(Base.Fix1(ustrip, m), 2m)
    InverseFunctions.test_inverse(Base.Fix1(ustrip, m), 2mm)
```
The units `m` and `mm` are imported in the test suite, so this can be simplified.

sostock (changes requested):

sostock:
> Test failure seems to be an `Aqua.jl` bug: actually running `]resolve` doesn't change `Project.toml` content.

This is https://github.com/JuliaTesting/Aqua.jl/issues/105. It looks like Aqua wants the sections ordered alphabetically and the entries within the sections ordered alphabetically as well, which is not the order that Pkg uses.

ext/InverseFunctionsUnitfulExt.jl
```diff
@@ -0,0 +1,8 @@
+module InverseFunctionsUnitfulExt
+using Unitful
+import InverseFunctions: inverse
+
+# `true` plays the role of 1, but doesn't promote unnecessary
+inverse(f::Base.Fix1{typeof(ustrip)}) = Base.Fix1(*, true*f.x)
```
sostock, line 6:
```suggestion
inverse(f::Base.Fix1{typeof(ustrip), <:Unitful.Units}) = Base.Fix1(*, true*f.x)
```
I would restrict the type of the first `ustrip` argument, because there is currently a discussion about adding another two-argument `ustrip` method where the first argument isn’t a unit (`ustrip(reinterpret, x)`, cf. https://github.com/PainterQubits/Unitful.jl/issues/630#issuecomment-1482765400)

sostock (approved):

oschulz:
Regarding the limitation to v1.9 - why not making InverseFunctions a required depencency on <=v1.8? InverseFunctions is extremely lightweight - even on v1.9, using an extension probably won't save even half a millisecond of load time.

oschulz:
See #652

PR #634, opened 2023-03-23
Dark-mode version of the logo

sostock:
For the dark theme of the documentation, I made a version of the logo with white outlines. I’m not sure whether it’s actually an improvement. @lmiq What do you think?

old:
old
new:
new


lmiq:
I think it looks slightly better if one uses a light gray instead of white for the lines:

With color: `#bababa`

image

With white:

image

sostock:
Thanks! I changed it to your light-gray version.

PR #639, opened 2023-03-29
Update Unitful.uparse() 

michikawa07:
1 default value of unit_context argument
Nevertheless, expanded pkg of Unitful.jl (i.e., UnitfulAtomic.jl) execute Unitful.register(), Unitful.uparse did not recognize these pkgs automatically (due to the defaut value of unit_context arg is Unitful only).
This commit modified this issue.

2 allowed_funcs
To enable parsing some strings such as "(1:10)m", I added `:colon` and `:(:)`.

3 parsing macro call
To enable parsing some strings such as "u\"m/s\"" (unit string including @u_str macro), I added some code in uparse content

### DIFF
--- a/src/user.jl
+++ b/src/user.jl
@@ -657,12 +657,12 @@ julia> uparse("1.0*dB")
 1.0 dB
 ```
 """
-function uparse(str; unit_context=Unitful)
+function uparse(str; unit_context=vcat(Unitful, Unitful.unitmodules))
     ex = Meta.parse(str)
     eval(lookup_units(unit_context, ex))
 end
 
-const allowed_funcs = [:*, :/, :^, :sqrt, :√, :+, :-, ://]
+const allowed_funcs = [:*, :/, :^, :sqrt, :√, :+, :-, ://, :colon, :(:)]
 function lookup_units(unitmods, ex::Expr)
     if ex.head == :call
         ex.args[1] in allowed_funcs ||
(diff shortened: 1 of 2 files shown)


src/user.jl
```diff
@@ -657,12 +657,12 @@ julia> uparse("1.0*dB")
 1.0 dB
 ```
 """
-function uparse(str; unit_context=Unitful)
+function uparse(str; unit_context=vcat(Unitful, Unitful.unitmodules))
```
jusack, line 660: +1 for this change, I just spent way too long trying to figure out why the imported unit was not parsing, the error message of lookup_units also was not super helpful

sostock:
> 1 default value of unit_context argument

IMO, this is a good change.

> 2 allowed_funcs To enable parsing some strings such as "(1:10)m", I added `:colon` and `:(:)`.

We could think about this, but I would rather keep the number of allowed functions small. If `(1:3)m` is allowed, why not `[1,2,3]m`? What’s so special about `:`?

> 3 parsing macro call To enable parsing some strings such as "u"m/s"" (unit string including @u_str macro), I added some code in uparse content

I don’t see how this is useful. Where would a “nested” unit string appear?

michikawa07:
Thank you for your feedback.

 >We could think about this, but I would rather keep the number of allowed functions small. If (1:3)m is allowed, why not [1,2,3]m? What’s so special about :?
Hmmm, that may be true. If `uparse` allows ":", it should also allow "[]", and then there are no limits.

> I don’t see how this is useful. Where would a “nested” unit string appear?
Yes I want use the nested unit sting. When saving a number with units, it may be converted to a macro and then written out as a string.

Issue #455, opened 2021-06-15
Comparisons with Inf fail


JeffFessler:
This comparison fails:
```julia
1u"m" < Inf
ERROR: DimensionError: 1.0 m and Inf are not dimensionally compatible.
```
where I think most users would expect to be `true` because `±Inf` should not need to have units.

Now `x = 1u"m"; x > 0` also fails but here we have the acceptable alternative `x > zero(x)`.
But there (apparently) is no analog like `infinity(x)` so we are stuck with `Inf`.

These are the key lines:
https://github.com/PainterQubits/Unitful.jl/blob/1e78e6b036144d70039b536de747dc1e83355997/src/quantities.jl#L265
https://github.com/PainterQubits/Unitful.jl/blob/1e78e6b036144d70039b536de747dc1e83355997/src/quantities.jl#L274

That last line could check for `±Inf` instead of always throwing an error.  The performance hit would be small because that line only gets used in the error condition.  Would a PR be welcome?

I search all issue for `Inf` and found nothing and found just 2 issues that concern `isless`: #399 #30

giordano:
> because ±Inf should not need to have units.

Why?

JeffFessler:
Well, why should it have units?  And if it should, then why is there no `infinity(x)` like there is a `zero(x)`?

The above is a MWE.  Here's the real back story.

```julia
using Unitful
using Plots

import Dates; d = oneunit(Dates.Day)
x = d * (1:5)
scatter(x, 2x) # works fine

x = 1u"m" * (1:5)
scatter(x, 2x) # fails due to Inf comparison
```

If a plot vs dates works, so should a plot vs meters...
Of course I could take this specific issue up over at Plots, but I think Unitful would be more user friendly if comparisons with `0` and `Inf` just worked (even without units), in the same way that `Int16(0) < 1.0f0` just works.  It might help things just work in other places that make such comparisons, with no obvious downside over an error message.

giordano:
> Well, why should it have units?

Because it's a number just like all others?  I don't see any reasons why it should be special.

> then why is there no infinity(x) like there is a zero(x)?

Because no one took the time to do that?  But `zero` and `one` are important as they are the neutral element of basic arithmetic operations, there is no such argument for `infinity`

JeffFessler:
> Because it's a number just like all others? I don't see any reasons why it should be special.
> 
> >

I disagree that it is "just like all the others" but I'll take this as a no on the PR.
[...]

PR #640, opened 2023-03-31
support comparison with zero

aplavin:
This is convenient and unambiguous when constrained for scalar units


cstjean:
By the same token, you could support addition by 0 and equality with 0. But unfortunately,

```julia
julia> 0u"m" == 0
false
```
so that'd be a breaking change.

cstjean:
> But there's no restriction on <, and being able to do some_unitful_function() > 0 would be an improvement.

I agree that it's convenient. I just worry about general consistency. You'd have that `0m < 0, 0m > 0, 0m == 0` all false. `0 <= 0m` would also be false, which is particularly uncomfortable.

cstjean:
Yeah, but then you'd have that x less-than-or-equal y is not the same as `x<y || x==y`

sostock:
Some previous discussion: https://github.com/PainterQubits/Unitful.jl/issues/455

Another point to consider: it feels inconsistent to me that `0 < 1m` holds, but `0s < 1m` doesn’t.

But it would be nice to have a more convenient method of comparing with zero.

cstjean:
Or `ustrip(x) > 0`.

Rereading the issue now, I wondered if it would it have impact on inference / code-gen? We'd go from `<(::typeof(1.0u"m"), ::Int)` that always-throws to maybe-throws. Also, you'd have to special-case affine-units, since `-5°C > 0` is ambiguous.

All in all, while it's a very nice idea, feels like the complications outweigh the benefit/convenience.

sostock:
Julia 1.13 will have `ispositive` and `isnegative` functions, we just need to add methods for quantities.

Issue #644, opened 2023-04-14
cannot assign to ustriped AbstractArrays


bjarthur:
it sure would be nice if Unitful played better with CuArrays.  for example, currently, one cannot `ustrip` a `CuArray` and then assign to it:

```
julia> using Unitful, CUDA

julia> c = CuArray(zeros(3) * 1u"s")
3-element CuArray{Quantity{Float64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}}, 1, CUDA.Mem.DeviceBuffer}:
 0.0 s
 0.0 s
 0.0 s

julia> ustrip(c)[1]=1   # innocuous warning about scalar indexing removed for succinctness
1

julia> c
3-element CuArray{Quantity{Float64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}}, 1, CUDA.Mem.DeviceBuffer}:
 0.0 s
 0.0 s
 0.0 s
```

however, if i simply add a new method identical to [this one](https://github.com/PainterQubits/Unitful.jl/blob/9e329e8e7200b3e42e0a6676d2948c66d35061b4/src/utils.jl#L78) but with `CuArray`, then it works fine:

```
julia> import Unitful: ustrip

julia> @inline ustrip(A::CuArray{Q}) where {Q <: Quantity} = reinterpret(Unitful.numtype(Q), A)
ustrip (generic function with 16 methods)

julia> ustrip(c)[1]=1
1

julia> c
3-element CuArray{Quantity{Float64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}}, 1, CUDA.Mem.DeviceBuffer}:
 1.0 s
 0.0 s
 0.0 s
```

to be all inclusive, i think the right way to go here is just use `AbstractArray`.  then it'll work too for MtlArray, ROCArray, etc.

_Originally posted by @bjarthur in https://github.com/PainterQubits/Unitful.jl/issues/630#issuecomment-1485550726_

Issue #630, opened 2023-03-21
Inconsistency with `ustrip(array)` and mixed units


sostock:
When calling `ustrip` on a quantity with mixed units like `dBV/m`, both logarithmic and linear units are stripped. When calling `ustrip` on an *array* of these quantities, only the linear part is stripped:
```julia
julia> x = 5u"dBV/m"
[5.0 dBV] m^-1

julia> ustrip(x)
5.0

julia> ustrip([x])
1-element reinterpret(Level{Unitful.LogInfo{:Decibel, 10, 10}, 1 V, Quantity{Float64, 𝐋^2 𝐌 𝐈^-1 𝐓^-3, Unitful.FreeUnits{(V,), 𝐋^2 𝐌 𝐈^-1 𝐓^-3, nothing}}}, ::Vector{Quantity{Level{Unitful.LogInfo{:Decibel, 10, 10}, 1 V, Quantity{Float64, 𝐋^2 𝐌 𝐈^-1 𝐓^-3, Unitful.FreeUnits{(V,), 𝐋^2 𝐌 𝐈^-1 𝐓^-3, nothing}}}, 𝐋 𝐌 𝐈^-1 𝐓^-3, Unitful.FreeUnits{(m^-1,), 𝐋^-1, nothing}}}):
 5.0 dBV
```

This is because calling `ustrip` on an `Array{<:Quantity}` actually `reinterpret`s the array. It is not possible to `reinterpret` the value `5.0dBV/m` to the value `5.0`, because the `Level` type stores the linear quantity (in this case `1.7782794100389228 V`, not the number `5.0`).

bjarthur:
`ustrip` is nominally deprecated, for reasons i don't understand.  try your example using broadcasting instead:

```
julia> x = 5u"dBV/m"
[5.0 dBV] m^-1

julia> ustrip.(x)
5.0

julia> ustrip.([x])
1-element Vector{Float64}:
 5.0
```

the downside of broadcasting is that `ustrip` then allocates:

```
julia> aa=rand(1000)*1u"ms";

julia> @time ustrip(aa);
  0.000010 seconds (1 allocation: 32 bytes)

julia> @time ustrip.(aa);
  0.000022 seconds (3 allocations: 7.969 KiB)
```

bjarthur:
[...]

Issue #627, opened 2023-03-15
ustrip of a view cannot be assigned to :(


bjarthur:
i need to send a unitful array into `LinearAlgebra.BLAS.symv!` as the last argument (i.e. the one assigned to).  so i tried `ustrip` and that works if i use it on the entire array.  but the tricky bit is that i just want to input a portion of it, and to be fast and efficient with memory, would like to use a view.  i'm finding that a ustrip of a view cannot be assigned to.  what follows is a minimal example without symv.  is there a way to make this work without allocation?  thx.

```
julia> v = zeros(3) * 1u"kg"
3-element Vector{Quantity{Float64, 𝐌, Unitful.FreeUnits{(kg,), 𝐌, nothing}}}:
 0.0 kg
 0.0 kg
 0.0 kg

julia> vv = @view v[1:2]
2-element view(::Vector{Quantity{Float64, 𝐌, Unitful.FreeUnits{(kg,), 𝐌, nothing}}}, 1:2) with eltype Quantity{Float64, 𝐌, Unitful.FreeUnits{(kg,), 𝐌, nothing}}:
 0.0 kg
 0.0 kg

julia> ustrip(vv)[1] = 1    # seems to work, but...
1

julia> vv     # nope, it actually doesn't
2-element view(::Vector{Quantity{Float64, 𝐌, Unitful.FreeUnits{(kg,), 𝐌, nothing}}}, 1:2) with eltype Quantity{Float64, 𝐌, Unitful.FreeUnits{(kg,), 𝐌, nothing}}:
 0.0 kg
 0.0 kg

julia> v     # not updated here either
3-element Vector{Quantity{Float64, 𝐌, Unitful.FreeUnits{(kg,), 𝐌, nothing}}}:
 0.0 kg
 0.0 kg
 0.0 kg

julia> ustrip(v)[1] = 1    # but without the view....
1

julia> v     # it works just fine
3-element Vector{Quantity{Float64, 𝐌, Unitful.FreeUnits{(kg,), 𝐌, nothing}}}:
 1.0 kg
 0.0 kg
 0.0 kg
```

bjarthur:
doh, okay, to answer my own question just 20 min later, you just reverse the composition and take a view of a ustrip.  so, continuing from the example above:

```
julia> uv = ustrip(v)
3-element reinterpret(Float64, ::Vector{Quantity{Float64, 𝐌, Unitful.FreeUnits{(kg,), 𝐌, nothing}}}):
[...]

PR #645, opened 2023-04-14
make assignment work on ustrip'ed AbstractArrays

bjarthur:
fixes https://github.com/PainterQubits/Unitful.jl/issues/644

note that i'm only superficially familiar with Unitful.jl, so am not sure if this breaks anything or is consistent with the roadmap.  but the tests pass locally.


sostock:
This will also apply to ranges, which makes it different from the current behavior:

Before:
```julia
julia> r = (1:5)u"m"
(1:5) m

julia> ustrip(r)
1:1:5
```

After:
```julia
julia> r = (1:5)u"m"
(1:5) m

julia> ustrip(r)
5-element reinterpret(Int64, ::StepRange{Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}, Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}):
 1
 2
 3
 4
 5
```

We could add a method `ustrip(r::AbstractRange) = ustrip.(r)` to keep the old behavior for ranges, but there are probably more types of `AbstractArray`s for which reinterpreting isn’t useful, like `SArray` from the `StaticArrays` package. Or does it not make a difference in that case (performance-wise)?

Reinterpreting is useful in two ways: to enable mutation of the original array, and to avoid allocations. For ranges and `SArray`s, both of these do not apply, because they are immutable and stack-allocated (unless you use mutable number types, which you could technically do).

We could implement `ustrip(::typeof(reinterpret), ::AbstractArray{<:Quantity})` (see https://github.com/PainterQubits/Unitful.jl/issues/630#issuecomment-1482765400) instead, that way we don’t add to the confusion of when reinterpreting happens and when it doesn’t.

bjarthur:
how about redefining `ustrip(::Array...` to dispatch on `DenseArray` instead of,  as this PR currently stands, `AbstractArray`?  doesn't make sense to reinterpret anything other than a dense array anyway, and doing so would preclude StaticArrays and Ranges, the two types you bring up, yet still cover GPUArrays, which is what motivates me.

```
julia> typeof(SVector(1,2,3)) <: DenseArray
false

julia> typeof(CuVector{Float64}([1,2,3])) <: DenseArray
true

julia> typeof(1:5) <: DenseArray
false
```

sostock:
> how about redefining `ustrip(::Array...` to dispatch on `DenseArray` instead

I like that. Maybe we could even use `StridedArray` instead of `DenseArray`, which is more general and still makes sense. BLAS functions usually support `StridedArray`s.

bjarthur:
i tried StridedArrays and it works for my use case.  have rebased the change into this PR.  tests pass locally.

bjarthur:
anything else i need to do here?  the CI is "awaiting approval".

sostock:
I just noticed that this is technically a breaking change. At least I consider it as one. A user that calls `ustrip(view(…))` in their code may now unintentionally modify the original array, because `ustrip` used to create a new array before this PR.

This could be avoided if we implement this as `ustrip(reinterpret, x)`. But if we do that, I would want it to error in cases where a wrong result would be returned (#630), which will require some more code.

bjarthur:
`ustrip` used to create a new array before this PR because of a deprecated method.

perhaps it's time to bump the version to 2.0?  what other new features are being held up because it would be a breaking change that we could implement if we did so?

certainly i'd like to see that deprecated method removed.  that should've been done in v0.5.

sostock:
I would like to release v2.0, but there are more breaking changes that I think should be included (and some of them require some further discussion), so it will take some time.

I still think we should use `ustrip(reinterpret, x)` for this functionality. It seems error-prone to me that `ustrip(x)` would return an object that shares its memory with `x`. Maybe I am biased because I did not know that this functionality existed until you opened #627, but I think many users don’t know of this feature, so it has the potential to lead to many errors. I think avoiding errors is more important than convenience.

Functions like `view` or `reshape` exist precisely for the purpose of reusing memory and allowing mutation of the original array. But `ustrip` is mainly a function that is applied to single values, not arrays. Therefore, I think it is not obvious that `ustrip(x)` might reuse memory.

Is there a reason why you think we shouldn’t add `ustrip(reinterpret, x)`? If you just think it is too long, maybe we could call it `ustrip!(x)` instead?

bjarthur:
should probably add the v2.0 label to this PR or the [issue](https://github.com/PainterQubits/Unitful.jl/issues/644) it fixes.  i see there are two others with that label.

i don't find it confusing that `ustrip` on an array reinterprets.  that is clearly documented and has existed for a long time.  moreover, it is consistent with how arrays are passed in by pointer to functions.  and it is probably what one would want to do normally anyway.

what is confusing is that `ustrip`ing a unit-less array or a strided one with units allocates.

that we disagree is an indication that more opinions should be solicited.  who could we ping that is a heavy user of Unitful.jl?

PR #648, opened 2023-05-07
Divide units instead of basefactors in convfact()

### LINKED ISSUE #647: Converting large powers of units can result in `NaN`, `Inf`, or 0
```julia
julia> using Unitful

julia> uconvert(u"kb"^10, 1.0 * u"kb"^11 * u"b"^-1)
999.9999999999999 kb^10

julia> uconvert(u"kb"^11, 1.0 * u"kb"^12 * u"b"^-1)
0.0 kb^11

julia> uconvert(u"kb"^12, 1.0 * u"kb"^13 * u"b"^-1)
NaN kb^12

julia> uconvert(u"kOe"^161, 1.0 * u"kOe"^162 * u"Oe"^-1)
1000.0 kOe^161

julia> uconvert(u"kOe"^162, 1.0 * u"kOe"^163 * u"Oe"^-1)
Inf kOe^162

julia> uconvert(u"kOe"^163, 1.0 * u"kOe"^164 * u"Oe"^-1)
NaN kOe^163
```

Apologies for the contrived examples, but this is only an issue for units with an “inexact” conversion part `!= 1.0`, and becomes more serious the more different it is from 1. Originally, I ran into this in the context of astrophysical units (see JuliaAstro/UnitfulAstro.jl#33), but I wanted to give a more minimal example with “built-in” units.

The issue is that `inex^p` in `basefactor()`:

https://github.com/PainterQubits/Unitful.jl/blob/bd09747409496a10ec7fab9d719f0841d8f7d5c5/src/units.jl#L229-L247

can result in `Inf` (`inex > 1`) or `0.0` (`inex < 1`) for large `p`, even though all the large factors end up canceling in the conversion. Ideally, `uconvert()` would be smarter and avoid dividing these canceling huge/tiny factors, but at a bare minimum, this should give an error instead of silently returning something wrong!

Socob:
This fixes #647.

In addition to the change in `convfact()` which allows the examples in #647 to work, I also added an error condition in case of floating point overflow or underflow in `basefactor()`. The latter is not actually needed for these cases anymore, but would still show up when the *real* conversion factor is actually very large/small, such as
```julia
uconvert(u"kb^12", 1u"b^12")
# before:       NaN kb^12
# with this PR: throws ArgumentError
```
Since the current result of `NaN` is just wrong, I’d consider this a bugfix (so the change in behavior is OK), and either way this is an obscure edge case which nobody else seems to have run into so far.

### DIFF
--- a/src/conversion.jl
+++ b/src/conversion.jl
@@ -1,6 +1,6 @@
 """
     convfact(s::Units, t::Units)
-Find the conversion factor from unit `t` to unit `s`, e.g., `convfact(m,cm) == 1//100`.
+Find the conversion factor from unit `t` to unit `s`, e.g., `convfact(m, cm) == 1//100`.
 """
 @generated function convfact(s::Units, t::Units)
     sunits = s.parameters[1]
@@ -12,14 +12,12 @@ Find the conversion factor from unit `t` to unit `s`, e.g., `convfact(m,cm) == 1
     sdim != tdim && throw(DimensionError(s(),t()))
 
     # first convert to base SI units.
-    # fact1 is what would need to be multiplied to get to base SI units
-    # fact2 is what would be multiplied to get from the result to base SI units
 
     inex1, ex1 = basefactor(t())
     inex2, ex2 = basefactor(s())
 
     a = inex1 / inex2
-    ex = ex1 // ex2     # do overflow checking?
+    ex = ex1 // ex2
 
     tens1 = mapreduce(tensfactor, +, tunits; init=0)
     tens2 = mapreduce(tensfactor, +, sunits; init=0)
--- a/src/conversion.jl
+++ b/src/conversion.jl
@@ -12,6 +12,7 @@ Find the conversion factor from unit `t` to unit `s`, e.g., `convfact(m, cm) == 
     conv_units = absoluteunit(FreeUnits(t())) / absoluteunit(FreeUnits(s()))
     inex, ex = basefactor(conv_units)
     pow = tensfactor(conv_units)
+    inex_orig = inex
 
     fpow = 10.0^pow
     if fpow > typemax(Int) || 1/fpow > typemax(Int)
@@ -28,8 +29,15 @@ Find the conversion factor from unit `t` to unit `s`, e.g., `convfact(m, cm) == 
     if ex isa Rational && denominator(ex) == 1
         ex = numerator(ex)
     end
-    y = (inex ≈ 1.0 ? 1 : inex) * ex
-    :($y)
+
+    result = (inex ≈ 1.0 ? 1 : inex) * ex
+    if fp_overflow_underflow(inex_orig, result)
+        throw(ArgumentError(
+            "Floating point overflow/underflow, probably due to large \
+            exponents and/or SI prefixes in units"
+        ))
+    end
+    return :($result)
 end
 
 """
--- a/src/conversion.jl
+++ b/src/conversion.jl
@@ -7,30 +7,23 @@ Find the conversion factor from unit `t` to unit `s`, e.g., `convfact(m, cm) == 
     tunits = t.parameters[1]
 
     # Check if conversion is possible in principle
-    sdim = dimension(s())
-    tdim = dimension(t())
-    sdim != tdim && throw(DimensionError(s(),t()))
+    dimension(s()) != dimension(t()) && throw(DimensionError(s(), t()))
 
-    # first convert to base SI units.
-
-    inex1, ex1 = basefactor(t())
-    inex2, ex2 = basefactor(s())
-
-    a = inex1 / inex2
-    ex = ex1 // ex2
-
-    tens1 = mapreduce(tensfactor, +, tunits; init=0)
-    tens2 = mapreduce(tensfactor, +, sunits; init=0)
-
-    pow = tens1-tens2
+    # first convert to base SI units
+    # use absoluteunit because division is invalid for AffineUnits;
+    # convert to FreeUnits first because absolute ContextUnits might still
+    # promote to AffineUnits
+    conv_units = absoluteunit(FreeUnits(t())) / absoluteunit(FreeUnits(s()))
+    inex_a, ex = basefactor(conv_units)
+    pow = tensfactor(conv_units)
 
     fpow = 10.0^pow
-    if fpow > typemax(Int) || 1/(fpow) > typemax(Int)
-        a *= fpow
+    if fpow > typemax(Int) || 1/fpow > typemax(Int)
+        inex_a *= fpow
     else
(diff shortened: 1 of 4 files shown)


src/conversion.jl
```diff
@@ -1,38 +1,26 @@
...
+    # promote to AffineUnits
+    conv_units = absoluteunit(FreeUnits(t())) / absoluteunit(FreeUnits(s()))
+    inex_a, ex = basefactor(conv_units)
+    pow = tensfactor(conv_units)
 
     fpow = 10.0^pow
-    if fpow > typemax(Int) || 1/(fpow) > typemax(Int)
-        a *= fpow
+    if fpow > typemax(Int) || 1/fpow > typemax(Int)
+        inex_a *= fpow
```
sostock, line 19: Is there a reason for renaming this variable from `a` to `inex_a`? I agree that `a` is not a very expressive name, but I also think we shouldn’t just change things without a good reason.

 If we want to rename it, I would just call it `inex`, not `inex_a`.

src/units.jl
```diff
@@ -232,19 +232,23 @@ function basefactor(inex, ex, eq, tens, p)
...
-            x = dpow(eq,p)
-            return ((inex * ex * 10.0^tens)^p, isinteger(x) ? Int(x) : x)
+            x = dpow(eq, p)
+            result = ((inex * ex * 10.0^tens)^p, isinteger(x) ? Int(x) : x)
         else
-            return ((inex * ex * 10.0^tens * eq)^p, 1)
+            result = ((inex * ex * 10.0^tens * eq)^p, 1)
         end
     end
+    if isfinite(inex) && !isfinite(first(result)) || !iszero(inex) && iszero(first(result))
```
sostock, line 248: Why does this only throw an error if `inex` is finite? Wouldn’t it be better to throw an error if `inex` is infinite, too?

sostock:
I finally got around to reviewing this, looks good to me.

> ```julia
> uconvert(u"kb^12", 1u"b^12")
> # before:       NaN kb^12
> # with this PR: throws ArgumentError
> ```

Do you think it makes sense to add a test for this?

> Since the current result of NaN is just wrong, I’d consider this a bugfix (so the change in behavior is OK), and either way this is an obscure edge case which nobody else seems to have run into so far.

I also think this is okay to change in a patch release (I consider it a bugfix as well).

src/conversion.jl
Socob: That was just to clarify the meaning of the variable, and `inex` was already taken (at the end of the function). I can get rid of the additional variable name if that’s preferred.

src/units.jl
Socob: The idea was to catch the case where the `basefactor()` function itself is the cause of the non-finite value *for valid inputs*, i. e. to detect if the computations in `basefactor()` end up producing something non-finite even if the passed value of `inex` is finite. I don’t know if there could be any valid use case for non-finite `inex`, but if this shouldn’t be allowed, then any checks for this should be part of initial argument validation at the beginning of the function. The error message about floating point overflow/underflow also wouldn’t make sense if `inex` is already non-finite.

Socob:
@sostock

> Do you think it makes sense to add a test for this?

I suppose it would make sense if there’s some way of testing the fact that it’s not returning `NaN` without explicitly requiring the error to be thrown (since that’s not exactly desired behavior, but rather a limitation of the implementation).

I replied to the comments above and will make the change for the variable name and add the test soon.

src/units.jl
sostock: Ah right, I didn’t see that `inex` was an argument to this method and comes directly from the `basefactors` dict.

src/units.jl
```diff
@@ -232,19 +232,23 @@ function basefactor(inex, ex, eq, tens, p)
...
+        throw(ArgumentError("Floating point overflow/underflow, probably due to large exponent ($p)"))
```
sostock, line 249: Should we have a similar test in `convfact`? Because right now it can still overflow:

```julia
julia> Unitful.convfact(u"ab^11", u"Tb^11")
Inf
```

Socob: Right, that’s in some sense separate from my original issue and the main change in this PR, but is of course good to have and easy to add! I did that in the latest commit (see below).

test/runtests.jl
```diff
@@ -202,17 +202,24 @@ end
...
+            let test_finite_or_throws(extype, ex) =
+                    @eval try
+                        # if the first line throws, go to @test_throws in
+                        # catch clause if not: make sure that result is finite
+                        $ex
+                        @test isfinite($ex)
+                        @test !iszero($ex)
+                    catch
+                        @test_throws $extype $ex
+                    end
```
sostock, line 219: I would like it more if this was a macro instead of using `eval`:
```julia
macro test_finite_or_throws(extype, ex)
    quote
        try
            # if the first line throws, go to @test_throws in
            # catch clause if not: make sure that result is finite
            result = $ex
            @test isfinite(result) && !iszero(result)
        catch
            @test_throws $extype $ex
        end
    end
end
```
The macro would need to be defined outside of the `@testset`, but I don’t think that’s too bad. I would just define it at the top of the `runtests.jl` file.

sostock (commented):
I think this is good to merge, I just have one more comment.

test/runtests.jl
Socob: This is actually how I did it first, only to realize that the macro can’t be defined locally, and I wasn’t sure if that’s a concern. I’ll change it!

Socob:
@sostock You’re definitely right! I hadn’t checked that particular method. I’ve added a check now!

src/conversion.jl
```diff
@@ -41,9 +29,15 @@ Find the conversion factor from unit `t` to unit `s`, e.g., `convfact(m,cm) == 1
...
     end
-    a ≈ 1.0 ? (inex = 1) : (inex = a)
-    y = inex * ex
-    :($y)
+
+    result = (inex ≈ 1.0 ? 1 : inex) * ex
+    if fp_overflow_underflow(inex_orig, result)
+        throw(ArgumentError(
+            "Floating point overflow/underflow, probably due to large" *
+            "exponents and/or SI prefixes in units"
```
sostock, line 37: There is no space between `large` and `exponents`. Either add a space or just put the string on a single line.

Socob: Damn, sorry about that! It’s fixed now.

giordano:
I think minor releases are generally cheap to do once you are in 1.x series, so when in doubt I prefer to bump the minor version number, but this sounds like it's potentially a bugfix (hence patch bump)?

sostock:
I also think this can be considered a bugfix, I will make a patch release.

@Socob Thanks for your contribution!

PR #652, opened 2023-05-26
Improve support for InverseFunctions

oschulz:
Makes `ustrip` support for `inverse` available on Julia <1.9 as well and adds documentation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,7 +16,7 @@ InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 InverseFunctionsUnitfulExt = "InverseFunctions"
 
 [compat]
-Aqua = "0.6"
+Aqua = "0.6.3"
 ConstructionBase = "1"
 InverseFunctions = "0.1"
 julia = "1"
--- a/Project.toml
+++ b/Project.toml
@@ -5,16 +5,21 @@ version = "1.14.0"
 [deps]
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+
+[weakdeps]
+InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
+
+[extensions]
+InverseFunctionsUnitfulExt = "InverseFunctions"
 
 [compat]
 Aqua = "0.6"
 ConstructionBase = "1"
+InverseFunctions = "0.1"
 julia = "1"
-
-[extensions]
-InverseFunctionsUnitfulExt = "InverseFunctions"
 
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
@@ -25,6 +30,3 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
 test = ["Aqua", "InverseFunctions", "LinearAlgebra", "Test", "Random"]
-
-[weakdeps]
-InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -24,6 +24,13 @@ true
 julia> ustrip(Float64, u"m", 2u"mm") == 0.002
 true
 ```
+
+`ustrip`` supports `InverseFunctions.inverse`:
+
+```jldoctest
+julia> inverse(Base.Fix1(ustrip, u"m"))(5)
+5 m
+```
 """
 @inline ustrip(u::Units, x) = ustrip(uconvert(u, x))
 @inline ustrip(T::Type, u::Units, x) = convert(T, ustrip(u, x))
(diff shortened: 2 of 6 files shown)


oschulz:
CC @aplavin

oschulz:
@sostock can you enable CI?

oschulz:
CI failure seems to be due to Aqua Project.toml formatting test on Julia v1.0. I disabled that test for older Julia versions, @sostock can start CI again?

oschulz:
Good to merge?

Project.toml
```diff
@@ -5,17 +5,22 @@ version = "1.14.0"
...
+InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
+[weakdeps]
+InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
+
+[extensions]
+InverseFunctionsUnitfulExt = "InverseFunctions"
+
```
sostock, line 17: If InverseFunctions is added as a regular dependency (so it will always be installed), I don’t see the point of using the weakdeps functionality at all. Or am I missing something?

src/utils.jl
```diff
@@ -74,6 +74,13 @@ julia> a[1] = 3u"m"; b
  3
  2
 ```
+
+`ustrip`` supports `InverseFunctions.inverse`:
+
+```jldoctest
+julia> inverse(Base.Fix1(ustrip, u"m"))(5)
+5 m
+```
```
sostock, line 83: This should be added to a different `ustrip` docstring (the one for the two-argument method, which starts at line 9 of the same file).

Project.toml
oschulz: I think it often makes sense to use a weak dep for Julia >= v1.9 but a standard dep instead of Requires on Julia < v1.9. The cost of using Requires can be very high, so a standard dep on pre-1.9 will (if the dependency is not very very heavy) often be the better solution.

In this specific case, we could just use a standard dep on all Julia versions, since InverseFunctions has a load time of under a millisecond. But maybe it's just good practice to use weak deps where possible in the future?

src/utils.jl
oschulz: Thanks @sostock , I moved it.

src/utils.jl
```diff
@@ -24,6 +24,13 @@ true
 julia> ustrip(Float64, u"m", 2u"mm") == 0.002
 true
 ```
+
+`ustrip`` supports `InverseFunctions.inverse`:
```
sostock, line 28:
```suggestion
`ustrip` supports `InverseFunctions.inverse`:
```

test/runtests.jl
```diff
@@ -2178,4 +2178,10 @@ end
...
 using Aqua
 
-Aqua.test_all(Unitful, ambiguities=VERSION≥v"1.1", unbound_args=false, piracy=VERSION≥v"1.8")
+Aqua.test_all(Unitful, ambiguities=VERSION≥v"1.1", unbound_args=false, piracy=VERSION≥v"1.8", stale_deps=false, project_toml_formatting = false)
+# Workaround for https://github.com/JuliaTesting/Aqua.jl/issues/107:
+Aqua.test_stale_deps(Unitful, ignore=[:InverseFunctions])
+# Project.toml seems to get formatted differently on older Julia versions:
+if VERSION ≥ v"1.6"
+    Aqua.test_project_toml_formatting(Unitful)
+end
```
sostock, line 2187:
```suggestion
Aqua.test_all(Unitful, ambiguities=VERSION≥v"1.1", unbound_args=false, piracy=VERSION≥v"1.8", stale_deps=false, project_toml_formatting=VERSION≥v"1.6")
# Workaround for https://github.com/JuliaTesting/Aqua.jl/issues/107:
Aqua.test_stale_deps(Unitful, ignore=[:InverseFunctions])
```

This is consistent with how the `ambiguities` and `piracy` keyword arguments are used.

oschulz: Oh, that's nicer, thanks!

Project.toml
oschulz: @sostock , @giordano how would you prefer to handle it? As it is now, or InverseFunctions as a required depencency across all Julia versions?

oschulz: Then I suggest we merge it like it is?

sostock (approved):

oschulz:
Merge? :-)

test/runtests.jl
```diff
@@ -2178,4 +2178,6 @@ end
 
 using Aqua
 
-Aqua.test_all(Unitful, ambiguities=VERSION≥v"1.1", unbound_args=false, piracy=VERSION≥v"1.8")
+Aqua.test_all(Unitful, ambiguities=VERSION≥v"1.1", unbound_args=false, piracy=VERSION≥v"1.8", stale_deps=false, project_toml_formatting=VERSION≥v"1.6")
+# Workaround for https://github.com/JuliaTesting/Aqua.jl/issues/107:
+Aqua.test_stale_deps(Unitful, ignore=[:InverseFunctions])
```
sostock, line 2183:
```suggestion
Aqua.test_all(Unitful, ambiguities=VERSION≥v"1.1", unbound_args=false, piracy=VERSION≥v"1.8", project_toml_formatting=VERSION≥v"1.6")
```
https://github.com/JuliaTesting/Aqua.jl/issues/107 seems to be fixed in Aqua v0.6.3, so maybe we can simplify this and restrict the compat entry for Aqua.

Project.toml
```diff
@@ -5,17 +5,22 @@ version = "1.14.0"
...
 [compat]
 Aqua = "0.6"
```
sostock, line 19:
```suggestion
Aqua = "0.6.3"
```
see my other comment

oschulz:
Thanks @sostock .

Issue #649, opened 2023-05-10
uparse breaks precompilation


amilsted:
It seems I cannot precompile code that makes use of `uparse`. In Julia 1.9 I get the error
```julia
nested task error: Evaluation into the closed module `Unitful` breaks incremental compilation because the side effects will not be permanent. This is likely due to some other module mutating `Unitful` with `eval` during precompilation - don't do this.
    Stacktrace:
      [1] eval
        @ ./boot.jl:370 [inlined]
      [2] eval
        @ ~/.julia/packages/Unitful/KGZ8U/src/Unitful.jl:1 [inlined]
      [3] uparse(str::String; unit_context::Module)
        @ Unitful ~/.julia/packages/Unitful/KGZ8U/src/user.jl:662
      [4] uparse
        @ ~/.julia/packages/Unitful/KGZ8U/src/user.jl:660 [inlined]
```
while trying to `precompile` a package that has code that uses `uparse` in its `@compile_workload` block (using `PrecompilationTools`). The code just calls `uparse` on a string (no module argument).

amilsted:
Suggestion (from @ChrisRackauckas on slack, albeit without the Unitful context) to use Symbolics.jl (I guess this?: https://docs.sciml.ai/Symbolics/stable/manual/parsing/) or JuliaInterpreter.

timholy:
A MWE would be interesting to see.

ChrisRackauckas:
I don't remember responding on Slack to anything related to unitful here, just on parsing a Julia expression 😅 . `:(2 * x)` can be ran by using JuliaInterpreter, or if you have a runtime version of that function you can parse it with Symbolics.jl. I am not sure if that's related to `uparse` or know what `uparse` is doing.

amilsted:
Sorry @ChrisRackauckas, I know I didn't mention Unitful on slack!

The `uparse` implementation calls `eval()` on an expression like `100m/s^2`, where `m` and `s` are usually fields of the `Unitful` module. There is also support for supplying a user module containing alternative definitions of unit symbols.

@timholy here's an MWE. Try to precompile the following package:
```julia
module MyTestPkg

using PrecompileTools
using Unitful: uparse

@compile_workload begin
    uparse("100s")
end

end
```
You will get:
```julia
[...]

PR #657, opened 2023-06-19
Avoid eval in uparse in simple cases

amilsted:
An partial solution to https://github.com/PainterQubits/Unitful.jl/issues/649


amilsted:
Any chance of merging this?

PR #664, opened 2023-07-11
Allow for prefixes of atm

briochemc:
It is common in climate science to talk about partial pressures in micro atmospheres (μatm). This PR simply allows for prefixes of `atm`.


sostock:
Are other prefixes common? Like `natm`, `matm`, `katm`?

I’m not against merging this with the less common prefixes included, but we could also think about only defining `μatm`, like we have for `a` and `ha`.

briochemc:
> Are other prefixes common? Like natm, matm, katm?

I don't think so but I am not sure...

sostock:
Apparently, `matm` is used by some people:

https://www.chegg.com/homework-help/questions-and-answers/authors-use-milliatmospheres-centimeter-matm-cm-rather-equivalent-dobson-unit-express-unit-q13498369

https://www.convertunits.com/info/milliatmosphere+%5btechnical%5d

@giordano Are you ok with adding all prefixes of `atm`?

giordano:
I hadn't heard them before, but there's apparently some precedents, so I'd say yes?

Issue #665, opened 2023-07-20
Latexify as soft dependency


gustaphe:
I wanted to lift the idea of merging UnitfulLatexify as an extension, now that those are possible. This will remove some user level friction, `using Unitful, Latexify` will just work.

Why here and not Latexify? It's an arbitrary choice, but how a unit is latexified is a property of the unit. For instance, UnitfulLatexify contains a table of exceptions, which seems more likely to update due to a version change in Unitful than one in Latexify. 

I can make a PR if you agree this is a good idea.

https://github.com/gustaphe/UnitfulLatexify.jl

giordano:
Sounds good

PR #668, opened 2023-07-31
Add Latexify extension

gustaphe:
#665 

This is not quite done yet. Because of how extensions work I had to break two things from UnitfulLatexify. 

1. Instead of i.e. `latexslashlabel(s, u)` it's `latexify(s, u; labelformat=:slash)`. This is a better interface anyway, I've been meaning to change it since the start but didn't want to break it. As far as I can tell I can't export those functions from the extension in a meaningful way, so maybe this is just the time. I don't think those functions are very common anyway.
2. This one is the one I primarily want input on: `u"One"`. UnitfulLatexify had `u"one"` as a non-fragile version of `NoUnits`, so you could have numbers that are recognizably unitful but with units of `1`. AFAICT, I cannot currently register units in an extension (I tried a couple of things, but nothing quite stuck), so I needed to register this unit in Unitful. But because Unitful imports `Base.one` I needed to change the spelling. Perhaps the right way is to fix registering units in an extension (more people might need that), but I don't know where to start.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -12,15 +12,18 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 [weakdeps]
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
 InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
+Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
 
 [extensions]
 ConstructionBaseUnitfulExt = "ConstructionBase"
 InverseFunctionsUnitfulExt = "InverseFunctions"
+LatexifyUnitfulExt = "Latexify"
 
 [compat]
 Aqua = "0.6.3"
 ConstructionBase = "1"
 InverseFunctions = "0.1"
+Latexify = "0.15.1, 0.16"
 julia = "1"
 
 [extras]
@@ -32,4 +35,4 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Aqua", "ConstructionBase", "InverseFunctions", "LinearAlgebra", "Test", "Random"]
+test = ["Aqua", "ConstructionBase", "InverseFunctions", "LinearAlgebra", "Test", "Random", "Latexify"]
--- a/ext/LatexifyUnitfulExt.jl
+++ b/ext/LatexifyUnitfulExt.jl
@@ -1,0 +1,450 @@
+#=========================================#
+# Extension for Unitful.jl + Latexify.jl, #
+# based on UnitfulLatexify.jl by          #
+# David Gustavsson (@gustaphe)            #
+#=========================================#
+module LatexifyUnitfulExt
+
+using Unitful:
+    Unitful,
+    Unit,
+    Units,
+    AbstractQuantity,
+    AffineUnits,
+    AffineQuantity,
+    power,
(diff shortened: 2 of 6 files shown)


sostock:
> This is not quite done yet. Because of how extensions work I had to break two things from UnitfulLatexify.
> 1. Instead of i.e. `latexslashunitlabel(s, u)` it's `latexify(s, u; labelformat=:slash)`. This is a better interface anyway, I've been meaning to change it since the start but didn't want to break it. As far as I can tell I can't export those functions from the extension in a meaningful way, so maybe this is just the time. I don't think those functions are very common anyway.

Are these functions (`latexslashunitlabel` etc.) actually meant to be called by the user (they don’t appear in the documentation), or are they just helper functions that are used by `plot(...; unitformat=latexslashunitlabel)`? I am asking because in the latter case, I wouldn’t even consider this breaking (IMO, if `latexslashlabel` is considered an internal function, it can be removed at any time). Anyway, I like this change.

> 2. This one is the one I primarily want input on: `u"One"`. UnitfulLatexify had `u"one"` as a non-fragile version of `NoUnits`, so you could have numbers that are recognizably unitful but with units of `1`. AFAICT, I cannot currently register units in an extension (I tried a couple of things, but nothing quite stuck), so I needed to register this unit in Unitful. But because Unitful imports `Base.one` I needed to change the spelling. Perhaps the right way is to fix registering units in an extension (more people might need that), but I don't know where to start.

Here are my thoughts, but keep in mind that I’m not really familiar with Latexify and UnitfulLatexify:

To me, it looks like the reason for having the `u"One"` unit is to enable using siunitx’s `\num{…}` for plain numbers. I think using the abilities of `\num{…}` is a useful feature even for users who do not work with units at all, so I think it should just be part of Latexify itself [^1]. Latexify already has the `fmt` keyword argument that takes arguments of type `AbstractNumberFormatter`, so we could propose to add a type `SiunitxFormatter <: AbstractNumberFormatter` to Latexify that simply prints numbers with `\num{…}`. I think that would remove the need for the `u"One"` unit, or is there some other use for it?

Another question to consider: When this is merged and released, and users of UnitfulLatexify upgrade Unitful without making changes to their code, it will load both UnitfulLatexify and this extension (I think). Will that lead to problems? I haven’t tried it yet, but I think we should make sure that upgrading Unitful does not break the code of existing UnitfulLatexify users.

[^1]: The fact that it doesn’t really have anything to do with units leads me to the same conclusion.

gustaphe:
Thank you for the comments.

> are they just helper functions that are used by `plot(...; unitformat=latexslashunitlabel)`?

This, but that *is* a user facing interface. I would not be surprised if few to no people ever used them though. Users going from UnitfulLatexify to LatexifyUnitfulExtension will have to either change `plot(...; unitformat=latexslashunitlabel)` to `plot(...; unitformat=(s,u) -> latexify(s, u; labelformat=:slash))` or better yet use `set_default(labelformat=:slash); plot(...;, unitformat=latexify)`. It's a good change though.

> add a type SiunitxFormatter <: AbstractNumberFormatter

That's not a half bad idea. It could even supersede the entire `unitformat` argument (in `Latexify`, not in `Plots` ...). I've been blind to that option, will need to spend a bit of time thinking about it but it's probably a good idea.

> Another question to consider

Yes, this is going to take a bit of thinking. When we merged `UnitfulRecipes` into `Plots` we introduced a warning -- I'll look into how that was done. There was a bit of compatibility hickups at first, but hopefully we learned the lesson.

Ickaser:
As a user, I have actually been using the `latexsquareunitlabel` version; I would not mind a change in syntax as discussed here. (And would love to see this handled as a proper extension!)

PR #675, opened 2023-08-09
Use fallback implementation for `min`, `max`

Socob:
`min` and `max` are currently explicitly defined for `AbstractQuantity` arguments and have their own separate re-implementation of `uconvert` for some reason. However, it turns out that this is unnecessary since the standard fallback implementations of `min` and `max` already work perfectly fine for `AbstractQuantity`, because comparison is also defined. Thus, this simply removes these extra method definitions, which, as a bonus, also provides all of the error checking added in #648 to `min` and `max` as well.


sostock:
Great! This also fixes another bug, because the deleted methods used `<` instead of `isless`:

Before (wrong):
```julia
julia> min(NaN*u"m", 1.0u"m")
1.0 m

julia> min(1.0u"m", NaN*u"m")
NaN m
```

After (correct):
```julia
julia> min(NaN*u"m", 1.0u"m")
1.0 m

julia> min(1.0u"m", NaN*u"m")
1.0 m
```

It would be good to add tests for that as well.

Socob:
@sostock Ah yes, I didn’t even noticed that. Tests added!

sostock (approved):

PR #691, opened 2023-11-04
Add new prefixes

Eben60:
2022 four prefixes were added.  Two for forming multiples:  ronna and quetta ("R", "Q")  Two forming submultiples:  ronto and quecto ("r", "q"). 

https://www.nist.gov/pml/owm/metric-si-prefixes

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -1,5 +1,7 @@
 # Convenient dictionary for mapping powers of ten to an SI prefix.
 const prefixdict = Dict(
+    -30 => "q",
+    -27 => "r",
     -24 => "y",
     -21 => "z",
     -18 => "a",
@@ -20,7 +22,9 @@ const prefixdict = Dict(
     15  => "P",
     18  => "E",
     21  => "Z",
-    24  => "Y"
+    24  => "Y",
+    27  => "R",
+    30  => "Q"    
 )
 
 """


sostock:
Thanks for your contribution!

I think it would be good to add the new prefixes, but there is one aspect to consider: `Unitful.rm` (rontometre) will collide with `Base.rm` (function for deleting a file). In principle, this is not a problem (most people will probably use `u"rm"` instead of importing `Unitful.rm`), but there is at least a precedent for avoiding these kind of name collisions: the name `Unitful.minute` was chosen instead of `Unitful.min` to avoid confusion with `Base.min`.

Nevertheless, I think we should add these prefixes (they are part of SI after all).

We could consider not exporting the new units from `Unitful.DefaultSymbols` (this is actually what would happen if we merge this PR as is[^1]), so that users of `Unitful.DefaultSymbols` don’t run into this collision [^2]. On the other hand, exporting only _some_ prefixed units and not the others seem inconsistent (and if you just import every symbol from a package you have to expect collisions if that package adds new symbols). Also, `Base.rm` is probably not as widely used as `Base.min`.

@giordano What do you think about this?

[^1]: To export the new units from `Unitful.DefaultSymbols`, the new prefixes would have to be added here as well: https://github.com/PainterQubits/Unitful.jl/blob/bc5dfb2e87b176274c29c4caec1fd79006c16e17/src/pkgdefaults.jl#L698-L699

[^2]: There are [several packages](https://juliahub.com/ui/Search?q=DefaultSymbols&type=code) that use `Unitful.DefaultSymbols`.

sostock:
The test failures on master are unrelated (they are fixed by #689, which was merged after this PR was posted).

Eben60:
`rm` is a commonly used function, and as for rontometer - I'm not sure if any distance having physical meaning can be measured in rontometers. 

So personally I'd suggest not exporting the new prefixes from `Unitful.DefaultSymbols` (any of them - for consistency) and mention that in the documentation.

PR #692, opened 2023-11-06
Allow two argument version of `nextfloat` and `prevfloat`

cmichelenstrofer:
The second argument is an integer `n` and the function gives the `n`th next/previous float. Here is the function definition for `Floats` in Julia: https://github.com/JuliaLang/julia/blob/fae6b7859b493160a290e79cf00ca1549bc9b6c7/base/float.jl#L857C1-L857C1

### DIFF
--- a/src/quantities.jl
+++ b/src/quantities.jl
@@ -436,8 +436,10 @@ Returns the sign bit of the underlying numeric value of `x`.
 """
 signbit(x::AbstractQuantity) = signbit(x.val)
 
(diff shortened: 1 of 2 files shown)


sostock:
Can you add some tests?

cmichelenstrofer:
> Can you add some tests?

Done. Does this look good?

sostock:
Yes, looks good to me. Thanks!

PR #698, opened 2023-11-20
Define helpers for unitful interfaces.

RomeoV:
This commit adds two simple functions `similar_dims` (and `similar_units`) that return a `Quantity` type with the dimensions (and units) constrained to those of the parameters

I found myself trying to write simulation code with strongly typed interfaces, i.e. including information about the units. Initially I wrote my interfaces like so:

```julia
const Meters = typeof(1.0m);
circumference_of_circle(r::Meters) = pi*r^2
```

However, when trying to autodiff through this code, I run into a problem, because `Meters` has the numerical type `Float64` baked in, and autodiff evaluates on a type `Quantity{Dual{Float64}}` (roughly).

We can instead define `Meters` like so:

```julia
const Meters{T<:Real} = Quantity{T, dimension(1.0m), unit(1.0m)}
circumference_of_circle(r::Meters{T}) where {T} = pi*r^2
circumference_of_circle(r::Quantity{T, dimension(1.0m), unit(1.0m)}) where {T} = pi*r^2
```

but I thought a better approach would be to provide some syntactic sugar to this "unit constraint".

With this PR, we can write

```julia
circumference_of_circle(r::similar_dims(u"m")) where {T} = pi*r^2
circumference_of_circle(r::similar_units(u"m")) where {T} = pi*r^2
```

The difference is that the first one only constrains the dimension, and the latter constrains both dimension and unit (i.e. doesn't allow e.g. `km`).

I'm happy to receive any feedback on the idea and the naming. Other names could be e.g. `quantity_with_dims` (but too long for my taste), or `dims_as` etc., but `similar` is already Julia lingo and feels appropriate in this context.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -243,6 +243,49 @@ struct DimensionError <: Exception
     y
 end
 
+"""
+    similar_dims(q::Quantity)
+    similar_dims(u::Units)
+Returns a type of [`Unitful.Quantity`](@ref) with the dimensions contrained to the
+dimension of `q` or `u`.
+Useful to build unitful interfaces that don't contrain the numeric type of the specific unit.
+
+Examples:
+
+```jldoctest
+julia> circumference_of_square(side::similar_dims(u"m")) = 4*side;
+julia> circumference_of_square((1//2)m)  # works
+2//1 m
+julia> circumference_of_square((1//2)km)  # also works
+2//1 km
+```
+
+See also [`Unitful.similar_units`](@ref).
+"""
+similar_dims(q::Quantity)  = Quantity{T, dimension(q),         U} where {T<:Real, U<:Unitlike}
+similar_dims(u::Units)     = Quantity{T, dimension(u),         U} where {T<:Real, U<:Unitlike}
+
+"""
+    similar_units(q::Quantity)
+    similar_units(u::Units)
+Returns a type of [`Unitful.Quantity`](@ref) with the dimensions and units contrained to the
+dimension and units of `q` or `u`.
+Useful to build unitful interfaces that don't contrain the numeric type.
+
+Examples:
+
+```jldoctest
+julia> circumference_of_square(side::similar_units(u"m")) = 4*side;
+julia> circumference_of_square((1//2)m)  # works
+2//1 m
+julia> # circumference_of_square((1//2)km)  # doesn't work, constrained to exactly meters
+```
+
+See also [`Unitful.similar_dims`](@ref).
+"""
+similar_units(q::Quantity) = Quantity{T, dimension(q),   unit(q)} where {T<:Real}
+similar_units(u::Units)    = Quantity{T, dimension(u), typeof(u)} where {T<:Real}
+
 Base.showerror(io::IO, e::DimensionError) =
     print(io, "DimensionError: $(e.x) and $(e.y) are not dimensionally compatible.");
 
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -27,6 +27,7 @@ import LinearAlgebra: istril, istriu, norm
 import Random
 
 export logunit, unit, absoluteunit, dimension, uconvert, ustrip, upreferred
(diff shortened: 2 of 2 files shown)


test/runtests.jl
```diff
@@ -2071,6 +2071,7 @@ end
 Base.:+(a::Num, b::Num) = Num(a.x + b.x)
 Base.:-(a::Num, b::Num) = Num(a.x - b.x)
 Base.:*(a::Num, b::Num) = Num(a.x * b.x)
+Base.:<(a::Num, b::Num) = a.x < b.x
```
sostock, line 2074: If the tests are simplified as below, this line shouldn’t be necessary.

test/runtests.jl
```diff
@@ -2079,6 +2080,18 @@ Base.OrderStyle(::Type{Num}) = Base.Unordered()
...
+area_of_square(side::WithUnits(u"m")) = side^2
+
+@testset "Unitful interfaces" begin
+    @test area_of_circle(Num(1.0)u"m") ≈ pi*m^2
+    @test area_of_circle(Num(1.0)u"km") ≈ pi*km^2
+    @test_throws MethodError area_of_circle(Num(1.0)u"s")
+
+    @test area_of_square(Num(0.5)u"m") ≈ 0.25*m^2
+    @test_throws MethodError area_of_square(Num(0.5)u"km")
+    @test_throws MethodError area_of_square(Num(0.5)u"s")
```
sostock, line 2093: These tests can be simplified. The functions `area_of_circle` and `area_of_square` are not necessary, you can just check that the type matches without calling the functions:

```suggestion
@testset "WithDims, WithUnits" begin
    @test isa(Num(1.0)m, WithDims(m))
    @test isa(Num(1.0)km, WithDims(m))
    @test !isa(Num(1.0)s, WithDims(m))

    @test isa(Num(0.5)m, WithUnits(m))
    @test !isa(Num(0.5)km), WithUnits(m))
    @test !isa(Num(0.5)s, WithUnits(m))
```

(I also replaced `u"m"` by `m`. Since the test suite imports these units, the `u"…"` macro is not needed.)

RomeoV: Thanks, great idea. Fixed in 15aa3eb8447cba0f8c59401d67bff3efca1d4de3.

src/utils.jl
```diff
@@ -243,6 +243,49 @@ struct DimensionError <: Exception
...
+"""
+WithDims(q::Quantity)  = Quantity{T, dimension(q),         U} where {T<:Real, U<:Unitlike}
+WithDims(u::Units)     = Quantity{T, dimension(u),         U} where {T<:Real, U<:Unitlike}
+
+"""
+    WithUnits(q::Quantity)
+    WithUnits(u::Units)
+Returns a type of [`Unitful.Quantity`](@ref) with the dimensions and units contrained to the
+dimension and units of `q` or `u`.
+Useful to build unitful interfaces that don't contrain the numeric type.
```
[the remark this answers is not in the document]
RomeoV: Fixed in c5b82180671f6d09608e4bb1b9eb8239b8281774

RomeoV:
> `WithDims` doesn’t include `Level`s (i.e., logarithmic quantities)

That is true, I can get to that still. There seems to be `Gain` and `Level`, which are related to `MixedUnits` as far as I can see. Should be easy to support if we want to move forward with this.

Eben60:
> Regardless, this seems to be a rare case.

Using complex numbers for electrical units like voltage, current, or impedance is a _very_ common case in electrical engineering.

> It seems to me that electrical impedance is represented as a complex number only for mathematical convenience

Sure. It is "only" a way of concise representation of complex relations. But then, most of the mathematics is "only" a way of concise representation of complex relations.

> The seven SI base units are all real-valued. 

Complex numbers can be equally used to represent oscillation processes of any nature, be it mechanical or electromagnetic. Both length (displacement) and current can be "complex-valued" in this sense.

PR #699, opened 2023-11-21
Add the units: parts per hundred thousand (pcm), million (ppm), billion (ppb), trillion (ppt), and quadrillion (ppq)

### LINKED ISSUE #684: Add parts per million (ppm) unit
Unitful.jl already has parts per hundred (percent), per thousand (permille), per ten thousand (pertenthousand). Can we also have the most commonly used parts per million?

eliascarv:
Reference: [Parts-per notation - Wikipedia](https://en.wikipedia.org/wiki/Parts-per_notation)

closes #684

### DIFF
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -231,6 +231,10 @@ substrate per s.
 \nPermyriad, a unit meaning parts per ten thousand. Printed as \"‱\".
 \nDimension: [`Unitful.NoDims`](@ref)."
 @unit pertenthousand  "‱"    Pertenthousand  1//10000           false
+"    Unitful.pcm
+\nPercentmille, a unit meaning part per hundred thousand.
+\nDimension: [`Unitful.NoDims`](@ref)."
+@unit pcm             "pcm"  Percentmille    1//100000          false
 "    Unitful.ppm
 \nPermillion, a unit meaning parts per million.
 \nDimension: [`Unitful.NoDims`](@ref)."
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -232,7 +232,7 @@ substrate per s.
 \nDimension: [`Unitful.NoDims`](@ref)."
 @unit pertenthousand  "‱"    Pertenthousand  1//10000           false
 "    Unitful.pcm
-\nPercentmille, a unit meaning part per hundred thousand.
+\nPercentmille, a unit meaning parts per hundred thousand.
 \nDimension: [`Unitful.NoDims`](@ref)."
 @unit pcm             "pcm"  Percentmille    1//100000          false
 "    Unitful.ppm
--- a/src/pkgdefaults.jl
+++ b/src/pkgdefaults.jl
@@ -236,15 +236,15 @@ substrate per s.
 \nDimension: [`Unitful.NoDims`](@ref)."
 @unit ppm             "ppm"  Permillion      1//1000000         false
 "    Unitful.ppb
-\nPerbillion, a unit meaning parts per billion.
+\nPerbillion, a unit meaning parts per billion (in the short-scale sense), i.e., 10^-9.
 \nDimension: [`Unitful.NoDims`](@ref)."
 @unit ppb             "ppb"  Perbillion      1//1000000000      false
 "    Unitful.ppt
-\nPertrillion, a unit meaning parts per trillion.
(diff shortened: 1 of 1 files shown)


juliohm:
Appreciate if someone can review and merge this. It is a very useful unit for chemical analysis.

cstjean:
We define our own ppm/ppb, but it's its own dimension, which feels better than having 3ppm == 0.000003.

src/pkgdefaults.jl
```diff
@@ -231,6 +231,22 @@ substrate per s.
 \nPermyriad, a unit meaning parts per ten thousand. Printed as \"‱\".
 \nDimension: [`Unitful.NoDims`](@ref)."
 @unit pertenthousand  "‱"    Pertenthousand  1//10000           false
+"    Unitful.ppm
+\nPermillion, a unit meaning parts per million.
+\nDimension: [`Unitful.NoDims`](@ref)."
+@unit ppm             "ppm"  Permillion      1//1000000         false
+"    Unitful.ppb
+\nPerbillion, a unit meaning parts per billion.
```
sostock, line 239:
```suggestion
\nPerbillion, a unit meaning parts per billion (in the short-scale sense), i.e., 10^-9.
```
Since billion, trillion, and quadrillion are ambiguous (https://en.wikipedia.org/wiki/Long_and_short_scales), it would be good to mention the actual values.

src/pkgdefaults.jl
```diff
@@ -231,6 +231,22 @@ substrate per s.
...
+\nDimension: [`Unitful.NoDims`](@ref)."
+@unit ppb             "ppb"  Perbillion      1//1000000000      false
+"    Unitful.ppt
+\nPertrillion, a unit meaning parts per trillion.
```
sostock, line 243:
```suggestion
\nPertrillion, a unit meaning parts per trillion (in the short-scale sense), i.e., 10^-12.
```

src/pkgdefaults.jl
```diff
@@ -231,6 +231,22 @@ substrate per s.
...
+\nDimension: [`Unitful.NoDims`](@ref)."
+@unit ppt             "ppt"  Pertrillion     1//1000000000000   false
+"    Unitful.ppq
+\nPerquadrillion, a unit meaning parts per quadrillion.
```
sostock, line 247:
```suggestion
\nPerquadrillion, a unit meaning parts per quadrillion (in the short-scale sense), i.e., 10^-15.
```

eliascarv:
Sorry @cstjean, I didn't understand your comment. Could you explain it?

cstjean:
We have

```julia
@dimension PPer     "PartsPer"          PartsPer
@refunit ppb        "ppb"               PartsPerBillion     PPer         false
@unit ppm           "ppm"               PPM                 1000ppb      false
```
(perhaps that's not super clean, but eh)

I'm not sure what's more appropriate for Unitful, but for us, in terms of "units exist to handle conversion code and prevent errors", making it its own dimension is the right call, even if it means having to add a `/ppm` division here and there.

Are there circumstances where you'd want 1ppm to be equal to 0.000001? For us, it's no, but I can definitely see that it could be true in some calculations, and in that case, this PR makes more sense than our code. Dividing by `1000000ppm` sounds a bit awkward for that conversion.

sostock:
`Percent`, `Permille`, and `Permyriad` are dimensionless, so adding a dimension only for `ppm`,  `ppb`, `ppb`, and `ppq` would be highly inconsistent. And changing `Percent`, `Permille`, and `Permyriad` to use that new dimension would be breaking.

SI also defines `1% == 0.01`, and it doesn’t seem controversial to me.

sostock:
> Reference: [Parts-per notation - Wikipedia](https://en.wikipedia.org/wiki/Parts-per_notation)

The Wikipedia page also mentions `pcm` (per cent mille). Does it make sense to add that as well? Then we would have all units listed on the page.

eliascarv:
@sostock, added.

juliohm:
Thank you all for the input. We are looking forward to seeing it merged and released. We already have some downstream applications that will benefit from these new units.

eliascarv:
@sostock, the PR is ready. Can you check if it's ready to merge?

sostock (approved):

eliascarv:
@sostock, could we have a patch release with this PR?

sostock:
No, this is a feature, so it cannot go into a patch release. I will release it as v1.19.0 shortly.

PR #703, opened 2023-12-10
Update docs on Logarithmic scales

Eben60:
I suppose Logarithmic scales, introduced at v0.4.0, are no more "new to Unitful and experimental" - a small edit to reflect this.

### DIFF
--- a/docs/src/logarithm.md
+++ b/docs/src/logarithm.md
@@ -5,9 +5,11 @@ end
 ```
 # Logarithmic scales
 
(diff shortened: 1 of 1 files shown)


sostock:
Logarithmic values are not _new_, but I would still consider them _experimental_, simply because they break some of the basic assumptions about equality and hashing in Julia (see #402). Maybe this issue (and probably some others) should be mentioned in the documentation?

Eben60:
@sostock thank you, I've changed the text correspondingly

PR #704, opened 2023-12-24
Make docs chapter on default units and physical constants

Eben60:
Adding documentation page on default units and physical constants and a script for creation/updating such a page. The script is to be run on the _local_ computer computer on each update of `Unitful` - it will be called from `docs/make.jl`

Here is the new chapter as [HTML rendered page](https://htmlpreview.github.io/?https://github.com/Eben60/Unitful.jl/blob/doc_units/docs/build/defaultunits.html).

(The script code is written without a deep understanding of the `Unitful` type system and is probably not exactly beautiful. However it is not intended for the end user anyway, and the result of the script is easily controlled.)


Eben60:
bump

docs/src/defaultunits.md
giordano, line 1: If this file is automatically generated, why are you checking it in?

Eben60: I am not sure I understand the question. What should be the procedure in this case and what can I do towards it?

giordano: Delete the file and git ignore it?

Eben60: The workflow as I intended it is as following: The file `defaultunits.md` will be generated on package updates on _local_ build of the documentation by the developer, who may want to look onto what changed. Then the file to be checked in just as every other documentation chapter. The `docs/make.jl` script includes a check if it run locally and will _not_ re-build the `defaultunits.md` file on the CI server. 

`docs/make.jl` on the CI server will check the `Unitful` package version, and will error, if the chapter has not been generated locally for this version

If this workflow is OK, then the file shouldn't be git-ignored. If you think it is better to have it generated on the CI server, it is necessary to adapt the`docs/make.jl` script.

sostock: > `docs/make.jl` on the CI server will check the `Unitful` package version, and will error, if the chapter has not been generated locally for this version

This means we would always have to bump the version at the same time we add a new unit/constant, right?

Eben60: I assume adding new units/constants is a new feature, meaning the (minor) version must bumped. Or what is the question?

docs/make_def-units_docs.jl
```diff
@@ -0,0 +1,322 @@
+module MakeDefUnitsDocs
+
+using Unitful, OrderedCollections
```
sostock, line 3: If `OrderedCollections` is used, it should be added to docs/Project.toml (OrderedCollections is installed as a dependency of `Latexify`, but if some future version of `Latexify` removes the dependency, this will error).

Eben60: (That's my first attempt at a substantial commit in a community project, I'm not well versed in git/github workflows, please be indulgent with me :blush: )

`OrderedCollections` now added to `docs/Project.toml `

docs/make_def-units_docs.jl
```diff
@@ -0,0 +1,322 @@
...
+    isconst(Unitful, n) && 
+    !(getproperty(Unitful, n) isa Union{Type, Unitful.Units, Unitful.Dimensions, Module, Function}) &&
+    isdocumented(n) ]
+    sort!(ph_consts)
+    return ph_consts
+end
+
+function isnodims(u) 
+    u isa Unitful.FreeUnits || return false
+    return getproperty(typeof(u), :parameters)[2] == NoDims
```
sostock, line 188:
```suggestion
    return dimension(u) == NoDims
```

docs/make_def-units_docs.jl
```diff
@@ -0,0 +1,322 @@
...
+function make_chapter(wr = true; verbose = false)
+    uids = uful_ids()
+
+    (;basicdims, compounddims) = uids |> getphysdims |> physdims_categories
+
+    basic_units =  unitsdict(basicdims, uids)
+    compound_units = unitsdict(compounddims, uids)
+    nodims_units = nodimsunits(uids) 
+    sections = OrderedDict(["Basic dimensions" => basic_units, 
+        "Compound dimensions" => compound_units])
```
sostock, line 306:
```suggestion
    sections = OrderedDict(["Base dimensions" => basic_units, 
        "Derived dimensions" => compound_units])
```

docs/src/defaultunits.md
```diff
@@ -0,0 +1,965 @@
...
+
+The candela, the SI base unit of luminous intensity. 
+
+#### Lumen 
+
+```
+Unitful.lm
+```
+
+The lumen, an SI unit of luminous flux, defined as 1 cd × sr.
```
sostock, line 116: This has the same dimension as `cd` (because solid angles are dimensionless), but I don’t think it should be in this section, because luminous intensity and luminous flux are different quantities. It’s probably easiest to just do this by hand (move it into a new subsection “Luminous flux” in the “Derived dimensions” (currently called “Compound dimensions”) section).

docs/src/defaultunits.md
```diff
@@ -0,0 +1,965 @@
...
+A quantity representing the Stefan-Boltzmann constant, defined as π^2 × k^4 / (60 × ħ^3 × c^2). 
+
+#### ϵ0 
+
+```
+Unitful.ε0
+Unitful.ϵ0
+```
+
+A quantity representing the vacuum permittivity constant, defined as 1 / (μ0 × c^2).
```
sostock, line 940: This is included twice (see line 900).

docs/make_def-units_docs.jl
```diff
@@ -0,0 +1,322 @@
+module MakeDefUnitsDocs
+
+using Unitful, OrderedCollections
+
+mdfile = "docs/src/defaultunits.md"
+mdheader = "docs/src/assets/defaultunits-header.md"
+# mdfooter = "docs/src/assets/defaultunits-footer.md"
```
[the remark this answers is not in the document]
Eben60: then uncomment it

docs/src/assets/defaultunits-logunits.md
```diff
@@ -0,0 +1,21 @@
+## Logarithmic units 
+
+| Unit           | Name                            |
+|----------------|---------------------------------|
+| `dB`       |        Decibel |
+| `B`        |         Bel |
+| `Np`       |        Neper |
+| `cNp`      |       Centineper |
+
+### "Dimensionful" logarithmic units
```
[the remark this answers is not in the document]
Eben60: Then maybe “Log units related to reference levels” to keep the title a bit shorter (and probably related, not relative - but I am never sure with my English)

Eben60:
@sostock thank you a lot for the review. I have added a couple of comments, but actually I have little experience in github collaboration, and I am afraid if I start now trying to fix each point, I have every chance to screw it here and there and it would be just more work for you. Could you probably just change everything as you consider it proper?

sostock:
@Eben60 Yes, I can apply these changes myself. No problem!

docs/make_def-units_docs.jl
```diff
@@ -0,0 +1,318 @@
...
+Generates the text of the `Pre-defined units and constants` documentation section 
+and writes it into the file if `wr==true`
+"""
+function make_chapter(wr = true; verbose = false)
+    uids = uful_ids()
+
+    (;basicdims, compounddims) = uids |> getphysdims |> physdims_categories
+
+    basic_units =  unitsdict(basicdims, uids)
+    compound_units = unitsdict(compounddims, uids)
```
sostock, line 299: If I understand correctly, this only collects units of dimensions that were declared with `@derived_dimension`. If we add units whose dimension isn’t “named”, they won’t show up here.

docs/src/assets/defaultunits-header.md
```diff
@@ -0,0 +1,16 @@
+# Units and сonstants pre-defined in the `Unitful` package
```
sostock, line 1: I would shorten this to `Pre-defined units and constants`. I think mentioning the name of the package isn’t necessary here. It should be obvious, as this is part of the documentation of the Unitful package.

Eben60: OK

PR #715, opened 2024-04-20
Fix `ustrip` broadcasting when range step has different unit than eltype

### LINKED ISSUE #712: Wrong result when broadcasting `ustrip` over ranges whose `eltype` and `step` have different units
```julia
julia> r = StepRange{typeof(1m),typeof(1cm)}(1m,1cm,2m)
(1//1:1//100:2//1) m

julia> length(r)
101

julia> ustrip.(r)
1:1:2

julia> length(ans)
2
```

sostock:
Fixes #712.

This PR also changes the return type when broadcasting over a `StepRange` and the conversion factor is a floating-point number: in this case, a `StepRangeLen` is returned (instead of a `StepRange`) because it is the better choice for floating-point numbers.

### DIFF
--- a/src/range.jl
+++ b/src/range.jl
@@ -154,13 +154,22 @@ broadcasted(::DefaultArrayStyle{1}, ::typeof(*), r::AbstractRange, x::AbstractQu
 broadcasted(::DefaultArrayStyle{1}, ::typeof(*), x::AbstractQuantity, r::AbstractRange) =
     broadcasted(DefaultArrayStyle{1}(), *, ustrip(x), r) * unit(x)
 
-const BCAST_PROPAGATE_CALLS = Union{typeof(upreferred), typeof(ustrip), Units}
+const BCAST_PROPAGATE_CALLS = Union{typeof(upreferred), Units}
 broadcasted(::DefaultArrayStyle{1}, ::typeof(*), r::AbstractRange, x::Ref{<:Units}) = r * x[]
 broadcasted(::DefaultArrayStyle{1}, ::typeof(*), x::Ref{<:Units}, r::AbstractRange) = x[] * r
 broadcasted(::DefaultArrayStyle{1}, x::BCAST_PROPAGATE_CALLS, r::StepRangeLen) = StepRangeLen{typeof(x(zero(eltype(r))))}(x(r.ref), x(r.step), r.len, r.offset)
-broadcasted(::DefaultArrayStyle{1}, x::BCAST_PROPAGATE_CALLS, r::StepRange) = StepRange(x(r.start), x(r.step), x(r.stop))
+broadcasted(::DefaultArrayStyle{1}, x::BCAST_PROPAGATE_CALLS, r::StepRange) = x(r.start):x(r.step):x(r.stop)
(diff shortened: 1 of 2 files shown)


src/range.jl
```diff
@@ -154,14 +154,23 @@ broadcasted(::DefaultArrayStyle{1}, ::typeof(*), r::AbstractRange, x::AbstractQu
 broadcasted(::DefaultArrayStyle{1}, ::typeof(*), x::AbstractQuantity, r::AbstractRange) =
     broadcasted(DefaultArrayStyle{1}(), *, ustrip(x), r) * unit(x)
 
-const BCAST_PROPAGATE_CALLS = Union{typeof(upreferred), typeof(ustrip), Units}
+const BCAST_PROPAGATE_CALLS = Union{typeof(upreferred), Units}
 broadcasted(::DefaultArrayStyle{1}, ::typeof(*), r::AbstractRange, x::Ref{<:Units}) = r * x[]
 broadcasted(::DefaultArrayStyle{1}, ::typeof(*), x::Ref{<:Units}, r::AbstractRange) = x[] * r
 broadcasted(::DefaultArrayStyle{1}, x::BCAST_PROPAGATE_CALLS, r::StepRangeLen) = StepRangeLen{typeof(x(zero(eltype(r))))}(x(r.ref), x(r.step), r.len, r.offset)
-broadcasted(::DefaultArrayStyle{1}, x::BCAST_PROPAGATE_CALLS, r::StepRange) = StepRange(x(r.start), x(r.step), x(r.stop))
+broadcasted(::DefaultArrayStyle{1}, x::BCAST_PROPAGATE_CALLS, r::StepRange) = x(r.start):x(r.step):x(r.stop)
```
sostock, line 161: This is bad for performance:
```
julia> using BenchmarkTools

julia> using Unitful: mJ, eV

julia> r = 1eV:1eV:5eV
(1:5) eV

julia> @btime mJ.($r);
  25.844 ns (0 allocations: 0 bytes) # v1.19.0 (returns a StepRange)
  55.767 ns (0 allocations: 0 bytes) # this PR (returns a StepRangeLen)
```
We could add methods that call the `StepRangeLen` constructor directly instead of going through `start:step:stop`:
```
julia> simple_mJ(r::StepRange) = StepRangeLen{typeof(mJ(r.start))}(Base.TwicePrecision(mJ(r.start)), Base.TwicePrecision(mJ(r.step)), length(r));

julia> @btime _mJ($r)
  14.013 ns (0 allocations: 0 bytes)

julia> mJ.(r) === simple_mJ(r)
true
```
This is a simple implementation (maybe the accuracy can be improved in certain cases by setting the appropriate number of trailing zero bits of `step.hi`), but, at least in the following case, it is actually more accurate than going through `start:step:stop`:
```
julia> r2 = 1eV:1eV:500_000eV
(1:500000) eV

julia> last(mJ.(r2))
8.01086714823366e-11 mJ

julia> last(simple_mJ(r2))
8.01088317e-11 mJ
```

sostock: I have now rewritten the function. It is slightly slower than the old version that returned a `StepRange`, but should be more accurate than the simple version in the previous comment:
```
julia> @btime mJ.($r)
  25.844 ns (0 allocations: 0 bytes) # v1.19.0 (returns a StepRange)
  33.379 ns (0 allocations: 0 bytes) # this PR (returns a StepRangeLen)
```

PR #716, opened 2024-04-24
Fix codecov upload

sostock:
This uses the `codecov/codecov-action` action for uploading to codecov instead of the deprecated `julia-actions/julia-uploadcodecov`. Hopefully, this fixes the codecov upload that seems to have broken recently.


sostock:
> We can go on with this PR right now, but later this year we'll be again without codecov working.

I think merging this PR won’t do anything, the upload doesn’t work with this PR either. But I can change it to `codecov-action@v4` (I thought v3 would work without token, that’s why I chose it). Then we only need an admin to add the token.

Issue #623, opened 2023-03-04
List all units and constants in the docs


rafaqz:
Its not easy to find e.g. the Stefan Bolzmann constant. Searching the docs for `Stefan` returns nothing.

It would be nice if there was a page listing all the docs for all of these.

ggggggggg:
Agreed, I just resorted to searching github to find the stephan boltzman constant, the boltzman constant and the charge of the electron. Maybe the constants could be re-exported into `Unitful.Constants` for easier tab completion as well?

giordano:
Are you aware of [`PhysicalConstants.jl`](https://github.com/JuliaPhysics/PhysicalConstants.jl)?

ggggggggg:
Yes, but there are already lots of useful constants in Unitful, I prefer to use those and have them easily accessible when it has what I need already.

giordano:
Because I don't think constants should live here.

PR #729, opened 2024-06-01
Add docs chapter on default units and physical constants

Eben60:
Adding (only) documentation page on default units and physical constants, as requested in #623 - a *temporary* alternative to generation of this documentation as suggested in #704.

The PR #704 is somewhat stalled: Having the script to do exactly what we want might take time. 

I have now manually edited the (initially generated) doc page according to @sostock remarks (https://github.com/PainterQubits/Unitful.jl/pull/704#discussion_r1573365359, https://github.com/PainterQubits/Unitful.jl/pull/704#discussion_r1573448624, https://github.com/PainterQubits/Unitful.jl/pull/704#discussion_r1573449229).

Could we probably merge this docs page as a provisional solution?

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -21,6 +21,7 @@ makedocs(
         "Interoperability with `Dates`" => "dates.md"
         "Extending Unitful" => "extending.md"
         "Troubleshooting" => "trouble.md"
+        "Pre-defined units and constants" => "defaultunits.md"
         "License" => "LICENSE.md"
     ]
 )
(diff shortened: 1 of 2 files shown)


Eben60:
Any chance?

I created this because of the frustration due to lack of units documentation: the only way to find out being studying the source code. As Unitful belongs to the most used packages in the Julia ecosystem, one can safely assume I was not alone with my frustration.

I understand automatic generation of the docs as in my PR https://github.com/PainterQubits/Unitful.jl/pull/704  could in theory have some unexpected implications, must be thoroughly checked etc. But what could speak against merging just the docs page according to the current PR?

:frowning_face:

PR #730, opened 2024-06-02
Readme: Add reference to UnitfulParsableString.jl

Eben60:


README.md
```diff
@@ -31,6 +31,7 @@ mathematical operations and collections that are found in Julia base.
 
 - [UnitfulEquivalences.jl](https://github.com/sostock/UnitfulEquivalences.jl): Enables conversion between equivalent quantities of different dimensions, e.g. between energy and wavelength of a photon.
 - [UnitfulLatexify.jl](https://github.com/gustaphe/UnitfulLatexify.jl): Pretty print units and quantities in LaTeX format.
+- [UnitfulParsableString.jl](https://github.com/michikawa07/UnitfulParsableString.jl): add method `Unitful.string` to convert `Quantity` (or some type) to parsable `String`.
```
sostock, line 34: I think it is unfortunate that the package extends `Unitful.string`. Not just because it is type piracy, but also because it is confusing  (Unitful doesn’t define a `string` function, so `Unitful.string` is just `Base.string`. Calling it `Unitful.string` obfuscates what it is actually doing).

I have opened an issue about this: https://github.com/michikawa07/UnitfulParsableString.jl/issues/12

sostock: Now that UnitfulParsableString.jl has changed from `Unitful.string` to `Base.string`, I would change the reference like this:

```suggestion
- [UnitfulParsableString.jl](https://github.com/michikawa07/UnitfulParsableString.jl): Add a `Base.string` method that converts quantities and units to parsable strings.
```

sostock (changes requested):

PR #731, opened 2024-06-11
support date/time + uniful arithmetic

aplavin:


src/dates.jl
```diff
@@ -188,3 +188,14 @@ isapprox(x::AbstractArray{Dates.CompoundPeriod}, y::AbstractArray{<:AbstractQuan
          kwargs...) = isapprox(y, x; kwargs...)
 
 sleep(x::AbstractQuantity) = sleep(ustrip(s, x))
+
+# Dates, Times, DateTimes
+
+for f in (:+, :-)
+    @eval Base.$f(x::Dates.DateTime, y::Quantity) = $f(x, Dates.Nanosecond(y))
```
sostock, line 195: We could think of replacing `Dates.Nanosecond(y)` by `trunc(Dates.Millisecond, y)` here (`DateTime` only has `Millisecond` precision).

This would enable something like `Dates.now() + 1234567890u"ps"` (which would fail right now because `Nanosecond(1234567890u"ps")` throws).

Using `trunc` instead of `round` matches the behavior for `Microsecond` and `Nanosecond`:
```
julia> Dates.DateTime("2024-06-22T10:30") + Microsecond(1999)
2024-06-22T10:30:00.001

julia> Dates.DateTime("2024-06-22T10:30") + Microsecond(2000)
2024-06-22T10:30:00.002
```

sostock (commented):
Some tests for this should be added to the test suite. Other than that, I only have one comment.

src/dates.jl
```diff
@@ -188,3 +188,14 @@ isapprox(x::AbstractArray{Dates.CompoundPeriod}, y::AbstractArray{<:AbstractQuan
          kwargs...) = isapprox(y, x; kwargs...)
 
 sleep(x::AbstractQuantity) = sleep(ustrip(s, x))
+
+# Dates, Times, DateTimes
+
+for f in (:+, :-)
+    @eval Base.$f(x::Dates.DateTime, y::Quantity) = $f(x, trunc(Dates.Millisecond, y))
+    @eval Base.$f(x::Dates.Time, y::Quantity) = $f(x, trunc(Dates.Nanosecond, y))
+    @eval Base.$f(x::Dates.Date, y::Quantity) = $f(x, trunc(Dates.Day, y))
```
sostock, line 197: I wouldn’t use `trunc` for the `Dates.Date` method (only for the other two methods), since `Dates` itself does also not allow smaller time units to be added to `Day`:
```julia
julia> Dates.today() + Dates.Hour(1)
ERROR: MethodError: no method matching +(::Date, ::Hour)
```
So I would revert the this method to the old version while leaving the other two like this.

sostock: What do you think?

PR #740, opened 2024-09-18
Re-define Unitful's yr (year) unit to be 365.2425 days

jakewilliami:
Specify Unitful's year unit to be 365.2425 days.

This is the true average number of days in the Gregorian year, every year that is exactly divisible by four is a leap year, except for years that are exactly divisible by 100, but these centurial years are leap years, if they are exactly divisible by 400.

See [`Dates.isleapyear`](https://github.com/JuliaLang/julia/blob/00739173/stdlib/Dates/src/types.jl#L206) and [`Dates.days(::Year)`](https://github.com/JuliaLang/julia/blob/00739173/stdlib/Dates/src/periods.jl#L465).

This is a suggested change in the interest of correctness, but it likely doesn't make much of a difference to most people!

See also #288.

### DIFF
--- a/docs/src/dates.md
+++ b/docs/src/dates.md
@@ -12,7 +12,7 @@ end
 The `Dates.FixedPeriod` union type includes all `Dates.Period`s that represent a fixed period of time, i.e., `Dates.Week`, `Dates.Day`, `Dates.Hour`, `Dates.Minute`, `Dates.Second`, `Dates.Millisecond`, `Dates.Microsecond`, and `Dates.Nanosecond`. These types can be converted to `Quantity`s or used in place of them.
 
 !!! note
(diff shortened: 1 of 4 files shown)


jakewilliami:
Apologies, I made a syntax error that I didn't catch.  I've rectified it.

sostock:
To me, 365.2425 days seems like the correct value. I don’t know why 365.25 was chosen. Maybe there is a scientific field where 365.25 days is commonly used? Anyway, I think this is a good change, but I consider it breaking, so it can only be changed in Unitful v2.0.

jakewilliami:
@sostock whoops, I deleted the fork, didn't realise that would close the PR.  Can this be reopened?

giordano:
The PR can't be reopened without you restoring the fork with exactly the same content.  But even with this PR not reopened the content is accessible.

PR #744, opened 2024-10-30
Fix docstring tests for Julia v1.12 and test on v1.10 (LTS)

sostock:


sostock:
Test failure for Julia 1.10 on Windows looks like https://github.com/JuliaLang/julia/issues/34700.

Issue #582, opened 2022-11-10
`unit` and `dimension` are inconsistent for `Level`s


sostock:
For `Level`s (and mixed quantities based on `Level`s), `unit` discards the logarithmic part, `dimension` does not:
```julia
julia> unit(5u"dBV")      # NoUnits

julia> dimension(5u"dBV") # voltage
𝐋^2 𝐌 𝐈^-1 𝐓^-3

julia> unit(5u"dBV*m")      # meters (a length)
m

julia> dimension(5u"dBV*m") # voltage * length
𝐋^3 𝐌 𝐈^-1 𝐓^-3
```
I think the behavior of `dimension` makes sense. To fix `unit`, there are two possibilities:
1. Use the `unit` of the `reflevel`, i.e., `unit(5u"dBV") == V`
2. Make `unit(::Level)` throw an error

icweaver:
Sorry if this would be better as a new issue. I think I'm hitting this too over in https://github.com/JuliaAstro/DustExtinction.jl/pull/58, specifically for `mag` support:

```julia-repl
julia> using Unitful, UnitfulAstro

julia> x = 2u"mag"
2 mag

julia> unit(x)
```

Is there anything that can be done on our end?

Potentially related: #267, #265

icweaver:
ahhh, TIL about `logunit`. I see this is what Plots.jl uses for their recipes, so nvm. Sorry for the noise!

cgarling:
I would argue `logunit` should be documented if it's the main way to get the unit out of a logarithmic quantity. I opened #769 to add this documentation.

PR #769, opened 2025-02-17
Add documentation for `logunit`

cgarling:
`logunit` is exported but not documented. This function is useful and documenting it would be beneficial (it recently came up in #582). This PR adds a docstring to `logunit` and adds it to the API documentation in the `logarithm.md` file.

### DIFF
--- a/docs/src/logarithm.md
+++ b/docs/src/logarithm.md
@@ -432,6 +432,7 @@ displaying logarithmic quantities:
 ## API
 
 ```@docs
+    Unitful.logunit
     Unitful.linear
     Unitful.reflevel
     Unitful.uconvertp
--- a/src/logarithm.jl
+++ b/src/logarithm.jl
@@ -9,6 +9,24 @@ function Base.float(x::Level{L,S}) where {L,S}
     return Level{L,S,typeof(v)}(v)
 end
 big(x::Level{L,S}) where {L,S} = Level{L,S}(big(x.val))
+"""
+    logunit(x::LogScaled)
+    logunit(x::Union{Type{<:LogScaled}, MixedUnits})
+Returns the units associated with a logarithmic `Quantity`, a logarithmically-scaled
+`Quantity` type, or a `Quantity` type with mixed logarithmic / linear scaling.
+
+Examples:
+```jldoctest
+julia> using Unitful
+
+julia> logunit(1*u"dB")
+dB
+
+julia> logunit(u"dB")
+dB
+```
+"""
+function logunit end
 logunit(x::Level{L,S}) where {L,S} = MixedUnits{Level{L,S}}()
 logunit(x::Type{T}) where {L,S,T<:Level{L,S}} = MixedUnits{Level{L,S}}()
 


icweaver:
Thanks for getting this going, Chris! To my surprise, it seems that logarithmic quantities are not actually subtypes of `Quantity`, but are their own thing it seems (`Gain` and `Level`), which are both subtypes of the unexported [`Unitful.LogScale`](https://painterqubits.github.io/Unitful.jl/stable/logarithm/#Unitful.LogScaled)

This bit me in https://github.com/MakieOrg/Makie.jl/issues/4801, so just wanted to give a heads up in case things like

https://github.com/PainterQubits/Unitful.jl/blob/f927270bf298bca219e2a2b4765b00a85e7f603a/src/logarithm.jl#L15-L16

need to be clarified

cgarling:
Good point -- I updated the docstring in the PR, let me know if it looks better

icweaver:
ugh, I forgot about `Unitful.MixedUnits`. Thanks for including that too! I don't have a lot of experience with this type system stuff, so will defer to Sebastian 👍🏾

cgarling:
This is just an addition to the docs and should be ready to merge if someone can have a look.

src/logarithm.jl
```diff
@@ -9,6 +9,24 @@ function Base.float(x::Level{L,S}) where {L,S}
...
+
+Examples:
+```jldoctest
+julia> using Unitful
+
+julia> logunit(1*u"dB")
+dB
+
+julia> logunit(u"dB")
+dB
```
sostock, line 26: Maybe add an additional example for `MixedUnits`, since they are already mentioned in the docstring?
```
julia> logunit(u"dBm/s")
dBm
```

src/logarithm.jl
```diff
@@ -9,6 +9,24 @@ function Base.float(x::Level{L,S}) where {L,S}
     return Level{L,S,typeof(v)}(v)
 end
 big(x::Level{L,S}) where {L,S} = Level{L,S}(big(x.val))
+"""
+    logunit(x::LogScaled)
+    logunit(x::Union{Type{<:LogScaled}, MixedUnits})
+Returns the units associated with a `LogScaled` instance, a `LogScaled` type,
```
sostock, line 15:
```suggestion
Returns the logarithmic "units" associated with a `LogScaled` instance, a `LogScaled` type,
```
The [manual](https://painterqubits.github.io/Unitful.jl/stable/logarithm/) refers to these as `logarithmic "units"`. I think this might be less confusing, especially in the case of `MixedUnits`.

src/logarithm.jl
```diff
@@ -9,6 +9,24 @@ function Base.float(x::Level{L,S}) where {L,S}
...
+```jldoctest
+julia> using Unitful
+
+julia> logunit(1*u"dB")
+dB
+
+julia> logunit(u"dB")
+dB
+```
+"""
```
sostock, line 28: A reference to the `unit` docstring might be useful, e.g.
```
See also: [`unit`](@ref).
```
(A reference from `unit` to `logunit` could be added as well.)

cgarling:
Thanks @sostock, I have implemented your feedback in the above two commits.

docs/src/logarithm.md
```diff
@@ -432,6 +432,7 @@ displaying logarithmic quantities:
 ## API
 
 ```@docs
+    Unitful.logunit
```
sostock, line 435: It looks like this list was sorted alphabetically, so `logunit` should come after `linear`.

cgarling:
@sostock the docs block has been reordered alphabetically as you requested.

cgarling:
Thank you!

Issue #443, opened 2021-05-11
nautical units


anielsen001:
Are there units for nautical miles and knots? I didn't see any but wasn't sure if I should expect to find them in Unitful or in perhaps a module of their own. 

They are straightforward to create and I would be willing to contribute if there were some suggestions about the best place to add them.

anielsen001:
Knots and nautical miles are standard for maritime and aircraft use and I think should be included somewhere convenient. These are two that I need all the time. 

Others like fathoms and more obscure nautical units appear from time to time also, but probably would not be considered standard, and the actual definitions are not always consistently defined anyway.

lukebemish:
With the various revised unit conversion factors that NIST is adopting in 2022, many of these less-commonly seen units will be a bit more standardized (in this case, especially fathoms and leagues). Unitful currently doesn't include most of these (when was the last time you tried to use furlongs, after all?), but I could see either including these in Unitful or making some sort of "UnitfulMisc.jl" or "UnitfulExtras.jl" package for all of the less-commonly-seen but still fairly standardized units. (some more commonly seen examples would be, for instance, mmHg).

I would be willing to implement either solution.

PR #775, opened 2025-03-30
Add nautical mile and knots

alexpattyn:
In part addressing https://github.com/PainterQubits/Unitful.jl/issues/443


alexpattyn:
Still need to add testing. 

But nautical miles and knots are common within the US (at least in my work). Haven't seen fathoms personally. So it would be nice to include them directly within Unitful as they can directly be converted to SI units and a separate module is probability overkill for two units that could just be placed within the US section.

alexpattyn:
For reference I am using:
```
Göbel, E., and I. M. Mills. "The international system of units (SI)." (2006).
```

https://web.archive.org/web/20170814094625/http://www.bipm.org/utils/common/pdf/si_brochure_8_en.pdf

Issue #767, opened 2025-02-02
Float32 degrees and Int radians is promoted to Float64


ettersi:
`promote(1f0°, 2rad)` returns a `Float64` pair, while many other combinations of `Float32` and `Int` arguments are promoted to a (imho correct) `Float32` pair. 

```julia
julia> promote(1f0°, 2rad)
(0.01745329238474369, 2.0)

julia> promote(1f0, 2)
(1.0f0, 2.0f0)

julia> promote(1f0°, 2)
(0.017453292f0, 2.0f0)
```

PR #782, opened 2025-05-15
New attempt at preserving the floating-point precision in `uconvert`

### LINKED ISSUE #753: `uconvert` does not preserve the floating-point precision of quantities in some cases
MWE:
```julia
julia> using Unitful

julia> x = 100f0u"cm"
100.0f0 cm

julia> uconvert(u"m", x)
1.0f0 m

julia> x = 45f0u"°"
45.0f0°

julia> uconvert(u"rad", x)
0.7853981633974483 rad
```
@sostock, is this expected?

sostock:
Closes #753.

This is a follow-up to #754. It should now work without breaking something.

@eliascarv I took the first commits from #754 and added the `floattype` function.

We should also look at our `promote_rule` methods to see whether we can use this to address #767.

### DIFF
--- a/src/conversion.jl
+++ b/src/conversion.jl
@@ -47,8 +47,8 @@ Returns 1. (Avoids effort when unnecessary.)
 convfact(s::Units{S}, t::Units{S}) where {S} = 1
 
 """
-    convfact(T::Type, s::Units{S}, t::Units{S})
-Returns a appropriate conversion factor from unit `t` to unit `s` for number type `T`.
+    convfact(T::Type, s::Units, t::Units)
+Returns the appropriate conversion factor from unit `t` to unit `s` for the number type `T`.
 """
 function convfact(::Type{T}, s::Units, t::Units) where {T<:Number}
     cf = convfact(s, t)
--- a/src/conversion.jl
+++ b/src/conversion.jl
@@ -50,12 +50,26 @@ convfact(s::Units{S}, t::Units{S}) where {S} = 1
     convfact(T::Type, s::Units, t::Units)
 Returns the appropriate conversion factor from unit `t` to unit `s` for the number type `T`.
 """
-function convfact(::Type{T}, s::Units, t::Units) where {T<:Number}
-    cf = convfact(s, t)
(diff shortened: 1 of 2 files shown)


src/conversion.jl
```diff
@@ -46,6 +46,33 @@ Returns 1. (Avoids effort when unnecessary.)
...
+    convfact(T::Type, s::Units, t::Units)
+Returns the appropriate conversion factor from unit `t` to unit `s` for the number type `T`.
+"""
+@generated function convfact(::Type{T}, s::Units, t::Units) where T
+    cf = convfact(s(), t())
+    if cf isa AbstractFloat
+        F = floattype(T)
+        # Since conversion factors only have Float64 precision,
+        # there is no point in converting to BigFloat
+        convert(F == BigFloat ? Float64 : F, cf)
```
eliascarv, line 59: Perhaps using `<:` instead of `==` is safer to prevent possible `BligFloat` type changes.
```suggestion
        convert(F <: BigFloat ? Float64 : F, cf)
```

sostock: I think it is unlikely that `BigFloat` will be changed to a non-concrete type, but we can use `<:` if you prefer it.

eliascarv: I think you're right. Perhaps just change`==` to `===` which is more common to compare types.

sostock: I don’t really like using `===` on types, because the same type can have different representations (although it shouldn’t be a problem in this case because `BigFloat` is a concrete type):
```
julia> (Rational{T} where T) == (Rational{S} where S)
true

julia> (Rational{T} where T) === (Rational{S} where S)
false
```
I also didn’t find an instance of using `===` on types in this package, but there is one for `==`: https://github.com/PainterQubits/Unitful.jl/blob/f375caf8f7c60f0dc2e284a97fc902365f55ab76/src/conversion.jl#L67

eliascarv: I see, that makes perfect sense.

eliascarv (approved):

PR #795, opened 2025-07-04
Move UnitfulLatexify into Unitful extension

Ickaser:
This should supercede #668, including with the discussion that took place there.

A couple notes for posterity:
- In #668, there is a dispatch for ranges of quantities that would latexify with `\qtyrange`. I dropped that dispatch here, since https://github.com/korsbo/Latexify.jl/pull/333 turns ranges into vectors now.
- UnitfulLatexify had a format check in the tests with JuliaFormatter, but I am letting Unitful's own CI setup take its place.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,12 +13,15 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
+Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
+LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 
 [extensions]
 ConstructionBaseUnitfulExt = "ConstructionBase"
 ForwardDiffExt = "ForwardDiff"
 InverseFunctionsUnitfulExt = "InverseFunctions"
+LatexifyExt = ["Latexify", "LaTeXStrings"]
 PrintfExt = "Printf"
 
 [compat]
@@ -27,6 +30,8 @@ ConstructionBase = "1"
 Dates = "<0.0.1, 1"
 ForwardDiff = "0.10, 1"
 InverseFunctions = "0.1"
+LaTeXStrings = "1.2.0"
+Latexify = "0.16.8"
 LinearAlgebra = "<0.0.1, 1"
 Printf = "<0.0.1, 1"
 REPL = "<0.0.1, 1"
@@ -39,6 +44,8 @@ Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
+Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
+LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
@@ -46,4 +53,4 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Aqua", "ConstructionBase", "ForwardDiff", "InverseFunctions", "LinearAlgebra", "Test", "Random", "REPL", "Printf"]
+test = ["Aqua", "ConstructionBase", "ForwardDiff", "InverseFunctions", "Latexify", "LaTeXStrings", "LinearAlgebra", "Test", "Random", "REPL", "Printf"]
--- a/ext/LatexifyExt.jl
+++ b/ext/LatexifyExt.jl
@@ -1,0 +1,404 @@
+module LatexifyExt
+using Unitful:
+    Unitful,
+    Unit,
+    Units,
+    AbstractQuantity,
+    AffineUnits,
+    AffineQuantity,
+    power,
+    abbr,
+    name,
+    tens,
+    sortexp,
+    unit,
+    NoDims,
+    ustrip,
+    @u_str,
+    genericunit,
+    has_unit_spacing
+using Latexify:
+    Latexify,
+    @latexrecipe,
+    latexify,
+    _latexarray,
+    latexraw,
+    FancyNumberFormatter,
+    PlainNumberFormatter,
+    StyledNumberFormatter,
+    SiunitxNumberFormatter,
+    AbstractNumberFormatter
+using LaTeXStrings: LaTeXString
+
+import Latexify: latexify
+
+import Base.*
+
+# utility functions ------------------
+
+function get_formatter(kwargs)
+    fmt = get(kwargs, :fmt, FancyNumberFormatter())
+    if fmt isa String
+        fmt = StyledNumberFormatter(fmt)
+    end
+    return fmt
+end
+get_format_env(fmt::SiunitxNumberFormatter) = :raw
+get_format_env(fmt) = :inline
+
+
+function getunitname(p::T, unitformat) where {T<:Unit}
+    unitname = get(unitnames, (unitformat, name(p)), nothing)
+    isnothing(unitname) || return unitname
+    if unitformat === :siunitx
+        return "\\$(lowercase(String(name(p))))"
+    end
+    return abbr(p)
+end
+
+function listunits(::T) where {T<:Units}
+    return sortexp(T.parameters[1])
+end
+
+"""
+```julia
+intersperse(t, delim)
+```
+Create a vector whose elements alternate between the elements of `t` and `delim`, analogous
+to `join` for strings.
+
+# Example
+```julia
+julia> intersperse((1, 2, 3, 4), :a)
+[1, :a, 2, :a, 3, :a, 4]
+```
+"""
+function intersperse(t::T, delim::U) where {T,U}
+    iszero(length(t)) && return ()
+    L = length(t) * 2 - 1
+    out = Vector{Union{typeof.(t)...,U}}(undef, L)
+    out[1:2:L] .= t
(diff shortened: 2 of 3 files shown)


Ickaser:
CI failure on Julia 1.0 is because Latexify v.0.15.16 (current version: 0.16.8, needed for extension) upped compat bound to Julia 1.6, and for testing, Latexify needs to be loaded. I suspect that pre-1.6 users who don't need Latexify might still be able to install and use Unitful, although I don't know for sure.

I'm not sure what the value is of testing as far back as Julia 1.0, at this point--even 1.6 is no longer considered LTS.

Project.toml
```diff
@@ -13,12 +13,15 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
...
 InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
+Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
+LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 
 [extensions]
 ConstructionBaseUnitfulExt = "ConstructionBase"
 ForwardDiffExt = "ForwardDiff"
 InverseFunctionsUnitfulExt = "InverseFunctions"
+LatexifyExt = ["Latexify", "LaTeXStrings"]
```
sostock, line 24: Does this mean that it is also an extension for `LaTexStrings`? So when someone includes `LaTexStrings` and `Unitful`, but not `Latexify`, this extension will also get loaded?

Ickaser:
(I guess @gustaphe may not see this unless I ping him directly.)

Ickaser:
@sostock any objections to this PR? Or things you would like clarified?

docs/src/assets/latex-allunits.png
giordano, line 1: If these images are generated by `docs/generate_latex_images.jl`, do we need to check them in the repo, as opposed to automatically generating them when building the documentation?

gustaphe: I haven't used it much, but `using tectonic_jll` makes `Latexify.render` use a lightweight LaTeX compiler.

Ickaser: I started trying to redo this image generation with `latextabular` and `Latexify.render` instead of writing to a `.tex` file and then compiling, and ran into a combination of LaTeX problems and Windows problems. Probably at least partly my misunderstanding but there were some quirks I don't have more time to figure out today.

Ickaser: After another afternoon trying to wrangle this into a working form, here are some concrete problems to work around for generating these images:
- With `tectonic_jll`, there is no way to pass a `continue-on-error` flag through `Latexify.render`. So any units without `siunitx` definitions halt the LaTeX compilation rather than not printing, for example_units.png
- To render a PNG, `Latexify.render` implicitly depends on a Ghostscript installation, and I can't get that working in WSL; it might be easier on the GitHub runner but I don't want to count on it.
- `tectonic_jll` doesn't seem to respect setting a white page background.

So: I am thinking that, although `Latexify.render` seems nice, what I will actually do is use Latexify to generate the tables and .tex files, then use separate GitHub actions to compile the LaTeX and convert to png.

Ickaser: Finally got this working, I think! (Although I won't be 100% until documentation gets deployed for real.)

Ickaser:
@giordano or @sostock , any other questions on this?

Ickaser:
Is there anyone who feels qualified to comment on this? This is holding up https://github.com/JuliaPlots/Plots.jl/pull/5143, which will make it so that the Plots.jl extension for Unitful can fully drop UnitfulLatexify as a dependency and make `using Plots, Unitful` load correctly in the gradually-approaching Plots.jl 2.0 (https://github.com/JuliaPlots/Plots.jl/issues/5093 ).

sostock:
When I build the docs locally, some things don’t look right. But that may just be a problem with my setup?

* In the “Arrays” section, the latex code is not rendered:
  Screenshot 2025-08-28 at 09-02-45 Latexifying units · Unitful jl

* The long table has no header:
  Screenshot 2025-08-28 at 09-04-58 Latexifying units · Unitful jl

sostock (changes requested):

docs/src/latexify.md
```diff
@@ -0,0 +1,177 @@
...
+using Unitful, Plots, Latexify 
+gr()
+default(fontfamily="Computer Modern")
+
+m = randn(10)u"kg"
+v = randn(10)u"m/s"
+plot(m, v; xguide="\\mathrm{mass}", yguide="v_x", unitformat=latexify)
+```
+
+This format, ``v_x\;\left/\mathrm{m}\,\mathrm{s}^{-1}\right.``, is subject to personal
```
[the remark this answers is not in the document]
Ickaser: image

This is what I see right now, which seems close to even to me. Perhaps worth noting though is that in the actual plot, what happens is GR consumes LaTeX syntax without actually using a TeX engine so it may not behave exactly as in LaTeX.

Changing the docs to correctly match the code output here, which uses `\;` before and after the slash.

ext/LatexifyExt.jl
```diff
@@ -0,0 +1,409 @@
...
+
+function _transform(q::AbstractQuantity, fmt::SiunitxNumberFormatter)
+    opening = fmt.version < 3 ? "\\SI{" : "\\qty{"
+    return Expr(:latexifymerge, opening, NakedNumber(q.val), "}{", NakedUnits(unit(q)), "}")
+end
+function _transform(q::AbstractQuantity, ::AbstractNumberFormatter)
+    Expr(
+        :latexifymerge,
+        NakedNumber(q.val),
+        has_unit_spacing(unit(q)) ? "\\;" : nothing,
```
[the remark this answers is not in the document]
Ickaser: @gustaphe siunitx does default to a thin space `\,` between number and unit (option `quantity-product` on page 46 of https://texdoc.org/serve/siunitx/0), where we are defaulting to `\;`. I will freely admit that I have never used this outside of Plots, so I haven't needed this option. Following siunitx seems logical to me, and this is as good a time as any to get in a breaking change like that. Do you have an opinion either way?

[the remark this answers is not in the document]
sostock: Do you want to switch to `\,` before merging this PR or should we merge this now (and make a separate PR for `\,`)?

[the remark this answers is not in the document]
Ickaser: I'm going to switch to `\,` wit this PR I think. As anxious as I am to get this in, this would be (formally) a breaking change so better get it in now. Finding/replacing `\;` should be pretty quick and easy, I'll take care of that Monday or Tuesday then this can merge.)

docs/Project.toml
```diff
@@ -2,9 +2,13 @@
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
+LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
```
giordano, line 8: Also, it'd be always good to have compat bounds at least for non-stdlibs non-jll packages.

giordano:
I'm a bit contrived.  It's annoying this is running with a somewhat unrelated issue (platform-specific rendering of the units).  I don't remember, is there a way to render the powers the same across all platforms, which would be used for generating the docs?

sostock:
I don’t know what happens there, but to me the error doesn’t suggest that it has anything to do with the exponent rendering.

giordano:
Ok, so this is even worse 😂

sostock:
Are you building the docs with a different Unitful version (which would happen as you noted in https://github.com/JuliaPhysics/Unitful.jl/pull/795#discussion_r2318652124)? I get that same error if I try to build the docs from this PR with the released Unitful (v1.24.0), since that doesn’t have the Latexify extension.

giordano:
Nuking the manifest and re-resolving the environment on `master`, I get a different error (which is probably progress):
```
julia> include("make.jl")
Precompiling Documenter...
  1 dependency successfully precompiled in 15 seconds. 52 already precompiled.
[ Info: Generating latex images for documentation
[ Info: Precompiling TectonicExt [35afe39f-415e-5b39-968f-951aa6823229]
┌ Info: Skipping precompilation due to precompilable error. Importing TectonicExt [35afe39f-415e-5b39-968f-951aa6823229].
└   exception = Error when precompiling module, potentially caused by a __precompile__(false) declaration in the module.
warning: accessing absolute path `/dev/null`; build may not be reproducible in other environments
ERROR: LoadError: IOError: could not spawn `gs -sDEVICE=pngalpha -dTextAlphaBits=4 -r300 -o /Users/mose/.julia/dev/Unitful/docs/src/assets/latex-examples.png /var/folders/v2/hmy3kzgj4tb3xsy8qkltxd0r0000gn/T/jl_7ehMJT.pdf`: no such file or directory (ENOENT)
```
This should really be using `Ghostscript_jll`....

Ickaser:
Shoot, sorry @giordano . I will freely admit that I did not anticipate people needing to build docs on another OS--in the CI I separately installed ghostscript (and also a small Debian font package) in order to get the docs working. 

> Nuking the manifest and re-resolving the environment on `master`, I get a different error (which is probably progress):
> 
> ```
> julia> include("make.jl")
> Precompiling Documenter...
>   1 dependency successfully precompiled in 15 seconds. 52 already precompiled.
> [ Info: Generating latex images for documentation
> [ Info: Precompiling TectonicExt [35afe39f-415e-5b39-968f-951aa6823229]
> ┌ Info: Skipping precompilation due to precompilable error. Importing TectonicExt [35afe39f-415e-5b39-968f-951aa6823229].
> └   exception = Error when precompiling module, potentially caused by a __precompile__(false) declaration in the module.
> warning: accessing absolute path `/dev/null`; build may not be reproducible in other environments
> ERROR: LoadError: IOError: could not spawn `gs -sDEVICE=pngalpha -dTextAlphaBits=4 -r300 -o /Users/mose/.julia/dev/Unitful/docs/src/assets/latex-examples.png /var/folders/v2/hmy3kzgj4tb3xsy8qkltxd0r0000gn/T/jl_7ehMJT.pdf`: no such file or directory (ENOENT)
> ```
> 
> This should really be using `Ghostscript_jll`....

This error looks like what you get if ghostscript isn't installed, which is why in the CI I had to manually install Ghostscript. I had no idea there was such a thing as `Ghostscript_jll`, that should probably have an extension for Latexify (where right now ghostscript is documented as a dependency of `Latexify.render` for .png output, but not installed). I'll try to address that with Latexify, and make another PR to Unitful with the `sources` in docs/Project.toml fix.

Ickaser:
#808 should make the docs build easier to work with once https://github.com/korsbo/Latexify.jl/pull/344 merges. I apologize for the confusion about that, and thanks @giordano, @sostock , @gustaphe for following me along here!

Issue #796, opened 2025-07-07
Testing back to Julia 1.0: is there still value in doing so?


Ickaser:
The CI for Unitful currently tests the following versions of Julia:
- 1.0
- 1.6
- 1.10
- 1 (currently 1.11)
- nightly

To me, testing even as far back as 1.6 seems excessive--from what I have seen of other packages, a more common test setup would be
- lts (currently 1.10)
- 1 (currently 1.11)
- pre (currently 1.12)

To me, it's an interesting curiosity that Unitful still works as far back as 1.0. But at this point, my feeling is that any new features don't need to deliberately support that far back--even 1.6 was released 4 years ago, and the ecosystem has come a long way in that time; users still working with pre-Julia 1.6 probably can't use current versions of a lot of other packages, and old versions of Unitful would still be available to them.

(Specifically, for #795 , the package I want to provide an extension for has compat limited to Julia 1.6 and up.)

Can @sostock or @giordano comment on this? Are there any other maintainers still involved who have strong preferences? (Is there anyone else who might be added as a maintainer?)

PR #798, opened 2025-07-18
Up minimum Julia to 1.6, change CI to use min, lts, 1, pre

Ickaser:
See discussion in #796, which technically this would close.

This would also make tests pass in #795, I believe.


Ickaser:
So... is there intent to merge this, now it has been approved by the two available committers? After today's 1.24 release I have this bumping the version to 1.25.0.

PR #803, opened 2025-08-01
Replace `PainterQubits` by `JuliaPhysics` in all relevant references

fhagemann:
Only place where `PainterQubits` was not changed to `JuliaPhysics` was in references to `PainterQubits/UnitfulUS.jl`:
https://github.com/PainterQubits/UnitfulUS.jl


fhagemann:
And what about this?
https://github.com/JuliaRegistries/General/blob/master/U/Unitful/Package.toml

giordano:
That should be updated too (otherwise we can't register new versions)

PR #804, opened 2025-08-03
Remove code for Julia <1.6

sostock:
Since Julia 1.6 is now the minimum supported version, a few `if isdefined(Base, …)` and `if VERSION ≥ …` can be removed.


giordano (approved):
Deleting code 🥳

PR #807, opened 2025-08-29
Define unit(::Any)m unit(::Funciton) and add a type restriction to unit(::Type{Union{Missing,T}})

### LINKED ISSUE #806: The method of unit(x::Type{Union{Missing,T}}) is causing stack overflow
I've been working on adding Unitful support to FiniteDifferences.jl in https://github.com/JuliaDiff/FiniteDifferences.jl/pull/244, and ran into this issue when running the test suite of the package. The method https://github.com/JuliaPhysics/Unitful.jl/blob/c1a9b4ea48ba5298589dbc95e4169ca9ebb0d582/src/utils.jl#L143-L145 on line 144 of `utils.jl` is causing stack overflow, probably because of the methods immediately surrounding it. Why does this method exist, if we already have separate methods for `Missing` and `T <: Any`? To prefer `T != Missing`, if it exists?

Anyways, here is the total output of the FiniteDifferences.jl test suite for reference: [test.out.txt](https://github.com/user-attachments/files/22041120/test.out.txt).

SeSodesa:
This is to resolve errors with uncluding units into FiniteDifferences.jl. See https://github.com/JuliaDiff/FiniteDifferences.jl/pull/244 for details.

Tests for Units.jl need to be updated, if this is to be accepted. The totalness of unit introduced via `unit(::Any)` might be a bit controversial, but I don't think that should be a problem. In general, things don't have units. Only Numbers do.

Closes #806.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -139,9 +139,11 @@ julia> unit(1.0) == NoUnits
 true
 ```
 """
+@inline unit(x::Any) = NoUnits # In general, things do not have units.
 @inline unit(x::Number) = NoUnits
 @inline unit(x::Type{T}) where {T <: Number} = NoUnits
-@inline unit(x::Type{Union{Missing, T}}) where T = unit(T)
+@inline unit(x::Function) = NoUnits # Functions do not have units. Function values might.
+@inline unit(x::Type{Union{Missing, T}}) where { T <: Number } = unit(T) # Type restriction required here for T, or might result in infinite recursion.
 @inline unit(x::Type{Missing}) = missing
 @inline unit(x::Missing) = missing
 


SeSodesa:
Ah, right, the methods were intentionally missing for date-type units where a duration cannot be fixed sensibly. I'll need to modify this a bit to keep those semantics intact.

sostock (changes requested):
Restricting `T` in the method `unit(::Type{Union{Missing, T}})` is the correct thing to do to fix the stack overflow. Adding `unit(::Any)` and `unit(::Function)` is not.

SeSodesa:
The different kind of error might be a good thing, though. It could function as an explicit way of telling a user that they misimplemented something, which is what I did and spent some time tracking down what and why, especially since the purpose of each method is not documented.

The method `unit(::Any)` could throw a specific error with a more informative error message, such as
```julia
function unit(x::Any)

    NotImplementedError("You tried calling unit on an object $x, that has none defined. Not everything has (or should have) units.")

end
```
User experience, and all that. StackOverflowErrors are never very nice.

sostock (changes requested):

src/utils.jl
```diff
@@ -145,6 +145,19 @@ true
 @inline unit(x::Type{Missing}) = missing
 @inline unit(x::Missing) = missing
 
+"""
+    unit(::Type{Any})
+
+This method is here to prevent infinite recursion and to provide a helpful error message in case it occurs.
+"""
```
sostock, line 152: I wouldn’t add a docstring to this method. This is not something the user need to know when they look up the documentation for the unit function. I would just add a comment.

```suggestion
# prevent infinite recursion when calling unit(Any)
```

sostock (changes requested):

src/utils.jl
```diff
@@ -145,6 +145,12 @@ true
 @inline unit(x::Type{Missing}) = missing
 @inline unit(x::Missing) = missing
 
+# Prevent infinite recursion, in case unit(Any) is called.
+
+@inline unit(x::Type{Any}) = throw( ArgumentError(
+    "unit(Any) was called, which has no meaningful result. This may be due to calling unit(eltype(…)), since eltype has a generic fallback method that returns Any."
+) )
```
sostock, line 152:
```suggestion
))
```
remove space between parentheses

PR #808, opened 2025-09-03
Make docs build less platform-dependent

Ickaser:
~This should make it easier to build docs locally cross-platform, making explicit a Ghostscript_jll dependency that wasn't obvious in #795 . Depends on https://github.com/korsbo/Latexify.jl/pull/344 getting merged and released (hi @gustaphe), and Latexify compat should get bumped accordingly in `docs/project.toml`.~

Existing caveats:
- ~Still depends on system having the `FreeSerif` and `FreeMono` fonts installed. The default font in tectonic for `Latexify.render` was not sufficiently clear for my tastes. Fortunately, these fonts might be installed by default by some unix packages, and if not can be installed for free with `sudo apt install fonts-freefont-ttf`, or from places online it seems.~
- Until https://github.com/JuliaPlots/Plots.jl/pull/5174 merges, the docs can only be built once per Julia session. (The docs build loads Plots, which loads UnitfulLatexify, which seems to overwrite some of the Latexify recipes and generate incorrect LaTeX on the next build.)

That second can possibly be assuaged (for the meantime) by adjusting `docs/make.jl` to check if the images already exist:
```julia
if !isfile((@__DIR__) * "/src/assets/latex-examples.png") || !isfile((@__DIR__) * "/src/assets/latex-allunits.png")
    @info "Generating latex images for documentation"
    include("generate_latex_images.jl")
end
```
where currently the latex images are generated either way.

### DIFF
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -28,7 +28,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
-      - run: sudo apt install -y fonts-freefont-ttf
       - uses: julia-actions/setup-julia@v2
         with:
           version: 1
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
-      - run: sudo apt install -y ghostscript fonts-freefont-ttf
+      - run: sudo apt install -y fonts-freefont-ttf
       - uses: julia-actions/setup-julia@v2
         with:
           version: 1
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,7 +1,6 @@
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
-Ghostscript_jll = "61579ee1-b43e-5ca0-a5da-69d92c66a64b"
 InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,6 +1,7 @@
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Ghostscript_jll = "61579ee1-b43e-5ca0-a5da-69d92c66a64b"
 InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,6 +1,7 @@
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Fontconfig = "186bb1d3-e1f7-5a2c-a377-96d770f13627"
 InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
@@ -16,3 +17,4 @@ Unitful = {path = ".."}
 Documenter = "1"
 Latexify = "0.16.9"
 Plots = "1"
+Fontconfig = "0.4"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -15,6 +15,6 @@ Unitful = {path = ".."}
 
 [compat]
 Documenter = "1"
-Latexify = "0.16.9"
+Latexify = "0.16.10"
 Plots = "1"
 Fontconfig = "0.4"
--- a/docs/generate_latex_images.jl
+++ b/docs/generate_latex_images.jl
@@ -1,4 +1,5 @@
-using LaTeXStrings, Unitful, Latexify, tectonic_jll
+using LaTeXStrings, Unitful, Latexify
+import tectonic_jll, Ghostscript_jll # needed for Latexify extensions that help render
 
 commands = [
     :(latexify(612.2u"nm")),
--- a/docs/generate_latex_images.jl
+++ b/docs/generate_latex_images.jl
@@ -32,7 +32,7 @@ ltab1 = LaTeXString("""
     \\color{black}
 """ * ltab1)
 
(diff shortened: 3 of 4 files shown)


giordano:
This has to be fixed upstream: https://github.com/korsbo/Latexify.jl/issues/343.  Wont do much here.

sostock:
> Still depends on system having the `FreeSerif` and `FreeMono` fonts installed. The default font in tectonic for `Latexify.render` was not sufficiently clear for my tastes. Fortunately, these fonts might be installed by default by some unix packages, and if not can be installed for free with `sudo apt install fonts-freefont-ttf`, or from places online it seems.

These fonts are only used in the second table. The first table uses the standard fonts. Is there a specific glyph that is not “sufficiently clear” in the standard font and that only appears in the second table? IMO, having to install a font to build the docs is not ideal, especially since it does not error when the font is missing, but silently generates the table with the header and the first column just empty.

Ickaser:
> > Still depends on system having the `FreeSerif` and `FreeMono` fonts installed. The default font in tectonic for `Latexify.render` was not sufficiently clear for my tastes. Fortunately, these fonts might be installed by default by some unix packages, and if not can be installed for free with `sudo apt install fonts-freefont-ttf`, or from places online it seems.
> 
> These fonts are only used in the second table. The first table uses the standard fonts. Is there a specific glyph that is not “sufficiently clear” in the standard font and that only appears in the second table? IMO, having to install a font to build the docs is not ideal, especially since it does not error when the font is missing, but silently generates the table with the header and the first column just empty.

I couldn't remember specifically what it was, but the default monospaced font has no glyph at all for pi:
image

We could use Courier New instead (which is _probably_ available by default [on Windows and Mac](https://www.ampsoft.net/webdesign-l/WindowsMacFonts.html)), but I had a hard time ascertaining whether that would be available on the Ubuntu used by a GitHub runner (or how to conveniently install it).

sostock:
We could use fontconfig to get the name of an installed monospace font (but I have not checked whether this font name can then be used to render the tables):
   ```
   julia> using Fontconfig: format, Pattern
   
   julia> format(match(Pattern("monospace")), "%{family}")
   "Noto Sans Mono"
   ```
   It should even be possible to check whether the font has a `π` glyph (U+03C0), by checking whether 3c0 is in one of the   ranges in this list:
   ```
   julia> format(match(Pattern("monospace")), "%{charset}")
   "20-7e a0-ac ae-377 37a-37f 384-38a 38c 38e-3a1 3a3-3e1 3f0-52f 10fb 1ab0-1ac0 1ac5 1ac7-1ace 1c80-1c88 1d00-1df9 1dfb-1f15 1f18-1f1d 1f20-1f45 1f48-1f4d 1f50-1f57 1f59 1f5b 1f5d 1f5f-1f7d 1f80-1fb4 1fb6-1fc4 1fc6-1fd3 1fd6-1fdb 1fdd-1fef 1ff2-1ff4 1ff6-1ffe 2000-2064 2066-2" ⋯ 504 bytes ⋯ "2736 2758-275a 27d5-27d7 27dc 27e6-27eb 27f5-27f6 2987-2988 29a3 29b8 2a00 2a05-2a06 2c60-2c7f 2de0-2e5d a640-a69f a700-a7ca a7d0-a7d1 a7d3 a7d5-a7d9 a7f2-a7ff a92e ab30-ab6b fe00 fe20-fe2f feff ff5b ff5d fffc-fffd 10780-10785 10787-107b0 107b2-107ba 1df00-1df1e 1f67c-1f67f"
   ```

Another option would be to render the individual cells in LaTeX and put them into a markdown table in the docs (the first column and the header would not be rendered by LaTeX, thus no font problems). But mixing the non-LaTeX text and the rendered PNGs would lead to inconsistent font sizes (depending on screen DPI).

Ickaser:
Using Fontconfig turns out to be simple and easy and handle this perfectly--`match(Pattern(spacing=100, charset=0x3c0))` finds a monospaced font with the π glyph. And the upstream Latexify change is merged and released, so I think this is ready to go if it looks alright.

sostock (approved):

Ickaser:
> Thanks for taking the time to improve this! Some comments:
> 
>     * overall this seems to be a major improvement compared to previous setup
> 
>     * while we're making platform-independent improvements, wrapping https://github.com/JuliaPhysics/Unitful.jl/blob/243efcb0f1da3496218250997e68da5b167c87e0/docs/make.jl#L3-L33
>        with
>       ```julia
>       withenv("UNITFUL_FANCY_EXPONENTS" => "false") do
>           # ...
>       end
>       ```
>         
>       would make doctests pass on macOS

I like that, would be happy to make that change.

>     * with the `withenv` above, I can successfully build the docs once in a new session, but an attempt at rebuilding the docs in the same session fails
>             

>           
>         
>       Any idea of what's going on?

That error on second build is a consequence of Plots still loading UnitfulLatexify as a dependency, which I can fix in https://github.com/JuliaPlots/Plots.jl/pull/5174 once Unitful has a new release. (Interlinking dependencies have been a bit of a bugbear here.)

giordano:
I can confirm I can successfully build the docs multiple times if I `]add https://github.com/Ickaser/Plots.jl#drop-unitful-latexify-dep` in the environment!  One minor annoyance is that during the build I get windows displaying `latex-allunits.png` and `latex-examples.png`, I'm not sure what's causing those windows to appear.  It's not blocking, but it's a small nuisance during the build. 

If you add `UNITFUL_FANCY_EXPONENTS=false` as suggested above, I'm happy with this PR.

sostock:
> That error on second build is a consequence of Plots still loading UnitfulLatexify as a dependency, which I can fix in https://github.com/JuliaPlots/Plots.jl/pull/5174 once Unitful has a new release. (Interlinking dependencies have been a bit of a bugbear here.)

Should we add Plots to our `[compat]` section (does that work when Plots is not a dependency?), so that users cannot install the new Unitful with an old Plots version?

sostock:
Then I think we should retroactively mark UnitfulLatexify (all versions) as incompatible with Unitful ≥ 1.25 in the general registry.

Ickaser:
> Then I think we should retroactively mark UnitfulLatexify (all versions) as incompatible with Unitful ≥ 1.25 in the general registry.

I intend to have new version of Plots require Unitful ≥ 1.25 and drop UnitfulLatexify, and UnitfulLatexify is now marked as having an upper bound of Unitful 1.23; would that be sufficient?

sostock:
No, because it would not prevent users from installing an old Plots, old UnitfulLatexify und new Unitful version. If they have those three packages in their Project.toml and run `Pkg.update`, that could happen since the resolver will make sure that compatible versions of all packages are installed. (It also could resolve to old Plots, old Unitful, new UnitfulLatexify, I don't know.)

I don't know whether that is a real problem, since some users may just have Unitful and Plots in their Project.toml (with UnitfulLatexify being installed as weakdep of Plots). But if they need Latexify functionality independent from plotting, they might have UnitfulLatexify in their Project.toml.

Another option might be to release a new UnitfulLatexify version that does nothing (or only prints a warning that it is no longer required) and is only compatible with upcoming Unitful versions. Then users who update an existing project would get the newest versions of everything. But it would not stop anyone from manually installing new Unitful and old UnitfulLatexify. If we want to prevent that, we need an upper bound on Unitful in all UnitfulLatexify versions.

sostock:
I see – then I think it should definitely be UnitfulLatexify 2.0.

Ickaser:
I think this PR should be good to go, then. We can continue conversation about UnitfulLatexify compat at https://github.com/gustaphe/UnitfulLatexify.jl/pull/22 if necessary, but once this PR and that PR are merged & released I can double check https://github.com/JuliaRegistries/General/pull/138160 (pending advice from someone who knows more than I do) to make sure that all old Unitful is treated as incompatible with new UnitfulLatexify and vice versa.

PR #809, opened 2025-09-04
Small fixes for new Latexify docs page

sostock:
Keeping the `\begin{equation}…\end{equation}` had the effect that the equations were numbered. Also a sentence was missing a verb.

@Ickaser fyi


Ickaser:
Thanks--I should maybe be reviewing my contributions a little more carefully 😶‍🌫️

PR #810, opened 2025-09-05
Swap `\;` for `\,` in LatexifyExt and matching tests

Ickaser:
As briefly discussed in https://github.com/JuliaPhysics/Unitful.jl/pull/795#discussion_r2311799069, it's marginally more appealing to use a thin space (`\,`) than the wider space (`\;`)  between the number and unit of a quantity. This is also the default behavior in LaTeX's siunitx. 

This PR makes that change (fortunately a two-character change), as well as updating the corresponding tests.
Current master: 
image

This PR:
image

The aspect which could depend on taste is that I also removed the spacing between an array and its unit, which seems to visually match the regular quantity spacing. 

For arrays, options are:
With `\,` (~third option~ this PR): (`\begin{equation}
\left[
\begin{array}{c}
1 \\
2 \\
3 \\
\end{array}
\right]\,\mathrm{m}
\end{equation}`)
image
With `\;` (current master):
image
With no extra spacing at all (~this PR~ former version of this PR): (`\begin{equation}
\left[
\begin{array}{c}
1 \\
2 \\
3 \\
\end{array}
\right]\mathrm{m}
\end{equation}`)
image

The one caveat I see to removing the space between array and unit is that, formally, the space should be left out when the unit is degrees, and so this puts other units at the same separation as degrees (instead of having an extra space).

### DIFF
--- a/ext/LatexifyExt.jl
+++ b/ext/LatexifyExt.jl
@@ -193,7 +193,7 @@ function _transform(q::AbstractQuantity, ::AbstractNumberFormatter)
     Expr(
         :latexifymerge,
         NakedNumber(q.val),
-        has_unit_spacing(unit(q)) ? "\\;" : nothing,
+        has_unit_spacing(unit(q)) ? "\\," : nothing,
         NakedUnits(unit(q)),
     )
 end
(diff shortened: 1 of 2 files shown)


giordano:
Did you push https://github.com/JuliaPhysics/Unitful.jl/commit/cf9ad0612e7c2632e32907089525b32f71555302 to the wrong branch?

Ickaser:
> Did you push [cf9ad06](https://github.com/JuliaPhysics/Unitful.jl/commit/cf9ad0612e7c2632e32907089525b32f71555302) to the wrong branch?

Yep, reverting that here

sostock:
For consistency, I would go with the third option, but it’s not a strong preference.

gustaphe:
No opinion, array times unit looks awkward either way, and when they are not side by side I doubt anyone will register the difference :D

I like the PR in general.

PR #811, opened 2025-09-13
Declare public symbols

Eben60:
Declaration of public symbols as requested in https://github.com/JuliaPhysics/Unitful.jl/issues/749

Notes:

- I had no idea which symbols maintainers would like to declare as public, thus tried my best guess.
- Variable names starting with underscore considered private.
- The file `public.julia` lists non-public symbols, too. That is for maintainers to have an overview, can be removed later, of course.
- The file `public.julia` has non-standard suffix because I remember having some issues with `.jl` suffix in such a case. It could be CodeCov, not sure. Can be changed to `.jl` if desired.
- See https://github.com/Eben60/Unitful.jl/blob/list_pub_symbols/docs/public_listing.jl to see how the file `public.julia` was generated.
- I've created a `testset` to check if all variables are declared in one way or another. This `testset` is put it to the top of `runtest.jl` because I got errors elswhere (see below).
- Declaring `VelocityFreeUnits` as `public` (or exported, for that case) leads to failure of the "Display" testset. No idea why, but reproduced on Mac and Ubuntu. For debugging, see https://github.com/Eben60/Unitful.jl/tree/mark_public_units_debug


src/public.julia
giordano, line 1: 1. Why `.julia` extension instead of `.jl`?
2. This file is simply unmaintainable.  Adding an option to macros like `@unit`/`@dimension`/`@derived_dimension` to make the symbol public looks a much more sustainable solution going forward

giordano (changes requested):

src/public.julia
Eben60: > Adding an option to macros

I agree, and would look into a macro solution. 

Independent on that: Before I continue, I'd prefer to know whether the the variables sorting into private and public, as currently implemented in `public.julia` is OK, or needs any changes. 

May I ask the maintainers to have a look?

Eben60: @sostock On log units: They are intended for the package user, not for some internal purposes. Therefore IMO they should be declared public, too. I'd suggest adding a warning to the docstrings about their experimental nature.

sostock (changes requested):
I haven’t reviewed the code yet, these comments are just about the default value of the new argument and the docstrings.

src/user.jl
```diff
@@ -351,31 +372,37 @@ macro prefixed_unit_symbols(symb,name,user_dimension,basefactor,autodocs=false)
...
 Not called directly by the user. Given a unit symbol and a unit's name,
 will define units without SI power-of-ten prefixes.
+If `makepublic=true`, makes the unit name public.
 
 !!! compat "Unitful 1.10"
     Documenting the resulting unit by adding a docstring before the `@unit_symbols` call
     requires Unitful 1.10 or later.
 
-Example: `@unit_symbols ft Foot 𝐋` results in `ft` getting defined but not `kft`.
+Example: `@unit_symbols ft Foot 𝐋` results in `ft` getting defined but not `kft`.
```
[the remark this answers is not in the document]
sostock: For `@affineunit`, the new argument still defaults to `true`.

test/runtests.jl
```diff
@@ -2547,6 +2556,109 @@ using REPL # This is necessary to make `@doc` work correctly
         """
 end
 
+VERSION >= v"1.11.0-DEV.469" && @testset "Declare Public" begin
```
sostock: The display tests above use `if isdefined(Base, :ispublic)` instead of `VERSION >= v"1.11.0-DEV.469"` to test whether `public` is available. They should all use the same test.

test/runtests.jl
```diff
@@ -2547,6 +2556,109 @@ using REPL # This is necessary to make `@doc` work correctly
...
+    all_public_vars = Symbol[
```
sostock: In my opinion, we don’t need to test all of these. Testing every public symbol means that someone needs to maintain this list (new units get added sometimes). We just need to test that the macros work, so I would test the following:
 * all symbols for one `@dimension`
 * all symbols for one `@derived_dimension`
 * one `@refunit` (the non-prefixed unit and one SI-prefixed unit)
 * one `@unit` (the non-prefixed unit and one SI-prefixed unit)
 * one `@affineunit`

sostock:
> I'd also suggest to declare `DefaultSymbols` as public and attach docstring to it. @sostock should I?

Yes, I would also declare `DefaultSymbols` public.

test/runtests.jl
Eben60: changed to `VERSION >= v"1.11.0-DEV.469"`

src/user.jl
```diff
@@ -351,31 +371,37 @@ macro prefixed_unit_symbols(symb,name,user_dimension,basefactor,autodocs=false)
...
             end
         end
+        VERSION >= v"1.11.0-DEV.469" && makepublic && push!(expr.args, Expr(:public, s))
         push!(expr.args, ea)
     end
 
     esc(expr)
 end
 
+# TODO update example in the docstring
```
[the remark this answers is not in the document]
Eben60: done

Eben60:
> > I'd also suggest to declare `DefaultSymbols` as public and attach docstring to it. @sostock should I?
> 
> Yes, I would also declare `DefaultSymbols` public.

done

test/runtests.jl
Eben60: Coming back to it, I see three different possibilities:

1. Keep complete lists of public AND private identifiers in the test, 
2. Define limited tests for public symbols as you suggested above, but keep full list of private symbols. This guarantees that for every added identifier, the developer must make a conscious choice: either declare it as public/exported, or add to the list of public symbols in the test suite or prepend name by underscore.
3. Just limited tests for public symbols, and no tests for private ones.

I've updated the test suite to print informative messages like 

> Error: These symbols in `public_vars_list` are already exported (remove them from the list): [:foo, :bar]"

With that, I think maintaining the lists in case identifiers are added or change public status would be easy and require no mental effort.

@sostock, I believe this is the last issue to act or decide on in this PR. Please let me know which way should we go, and if your preference is 2. or 3., I'd implement it according to your suggestion.

PR #845, opened 2026-03-30
Fix unit promotion for same units, different context

gpeairs:
Unit promotion for "same units but different promotion context" uses a unit singleton instance as a type parameter, but it should use the type.

Example:

```julia
import Unitful: μm, mm, ContextUnits
μm2μm = ContextUnits(μm,μm)
μm2mm = ContextUnits(μm,mm)
x, y = promote(1.0μm2μm, 1.0μm2mm)
upreferred(x)
```

Before the fix:

```
ERROR: MethodError: no method matching (::Unitful.FreeUnits{(m,), 𝐋, nothing})()
```

After the fix:
```
1.0e-6 m
```

### DIFF
--- a/src/promotion.jl
+++ b/src/promotion.jl
@@ -28,7 +28,7 @@ promote_unit(x::Units, y::Units, z::Units, t::Units...) =
 @inline _promote_unit(x::ContextUnits{N,D,P,A}, y::ContextUnits{N,D,P,A}) where {N,D,P,A} = x  #ambiguity reasons
 # same units, but promotion context disagrees
 @inline _promote_unit(x::ContextUnits{N,D,P1,A}, y::ContextUnits{N,D,P2,A}) where {N,D,P1,P2,A} =
-    ContextUnits{N,D,promote_unit(P1(), P2()),A}()
+    ContextUnits{N,D,typeof(promote_unit(P1(), P2())),A}()
 # different units, but promotion context agrees
 @inline _promote_unit(x::ContextUnits{N1,D,P}, y::ContextUnits{N2,D,P}) where {N1,N2,D,P} =
     ContextUnits(P(), P())
--- a/src/promotion.jl
+++ b/src/promotion.jl
@@ -26,13 +26,10 @@ promote_unit(x::Units, y::Units, z::Units, t::Units...) =
     upreferred(dimension(x))
 
 @inline _promote_unit(x::ContextUnits{N,D,P,A}, y::ContextUnits{N,D,P,A}) where {N,D,P,A} = x  #ambiguity reasons
-# same units, but promotion context disagrees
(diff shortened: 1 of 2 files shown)


Ickaser:
FWIW, this seems reasonable to me, although I'm not especially familiar with the machinery here.

gpeairs:
Gentle bump

Ickaser:
@sostock any thoughts on this?

src/promotion.jl
```diff
@@ -28,7 +28,7 @@ promote_unit(x::Units, y::Units, z::Units, t::Units...) =
 @inline _promote_unit(x::ContextUnits{N,D,P,A}, y::ContextUnits{N,D,P,A}) where {N,D,P,A} = x  #ambiguity reasons
 # same units, but promotion context disagrees
 @inline _promote_unit(x::ContextUnits{N,D,P1,A}, y::ContextUnits{N,D,P2,A}) where {N,D,P1,P2,A} =
-    ContextUnits{N,D,promote_unit(P1(), P2()),A}()
+    ContextUnits{N,D,typeof(promote_unit(P1(), P2())),A}()
```
sostock: I actually don’t see a reason to return `ContextUnits` here.

When the units and contexts both disagree, we return `FreeUnits` (line 36 in this file). But when only the contexts disagree, we create new context by promoting the two contexts. This seems inconsistent to me. I think we should either return `FreeUnits` in both cases or `ContextUnits` in both cases. I would prefer `FreeUnits` since making up a new context (that might be different from either of the contexts) seems wrong to me.

The current behavior has been around since `ContextUnits` were introduced (9545246). But since this particular method errors due to the wrong type parameter, we can change it to whatever we want without breaking something.

Therefore, my preference would be to return `FreeUnits` here. This can be achieved by just deleting this method (the one at line 36 will be used in its place).

sostock: Actually, on second thought, promoting the contexts might be the right thing to do (since the object should promote as if it had that unit). But then I would argue that the method at line 36 is wrong. However, changing the one at line 36 (to use `promote_unit(P1(), P2())` as well) would be breaking.

gpeairs: For me, this comes up when there's a stray quantity with the "wrong" context. I'd prefer the result become `FreeUnits` and use the next promotion context it sees rather than gradually contaminate everything with the base unit promotion context. I can see how promoting the contexts could be considered "correct", but I'd be happy with `FreeUnits` and consistency with the line 36 method.

For the sake of clarity, the `_promote_unit` method itself doesn't error, but there's an error whenever one tries to use the promotion context, so I think changing this to `FreeUnits` still doesn't break anything that worked before.

Let me know what you decide, and I can update the PR.

gpeairs:
I switched to the FreeUnits suggestion above by deleting the buggy method, but can revert if you have a different preference.

gpeairs:
Gentle bump

gpeairs:
Another bump

gpeairs:
Gentle bump again, is there anything I can do to help get this fix merged?

PR #849, opened 2026-05-22
Make `numtype` public

PatrickHaecker:
Without `numtype`, there is no function to just get the numeric type. So make it `public` and support common `Base` types.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,8 +1,41 @@
 @inline isunitless(::Units) = false
 @inline isunitless(::Units{()}) = true
 
+"""
+    numtype(x::Number)
+    numtype(::Type{T}) where {T<:Number}
+    numtype(x::Base.Enum)
+    numtype(::Type{T}) where {T<:Base.Enum}
+
+Return the underlying numeric type of a number or number type.
+
+For an [`AbstractQuantity`](@ref Unitful.AbstractQuantity), this is the type
+parameter describing the value stored alongside the units. For an `Enum`, it is
+the integer type that backs the enumeration. For a plain `Number`, there is
+nothing to strip, so `numtype` returns its type. Other number-like types can
+extend `numtype` by adding methods.
+
+```jldoctest
+julia> using Unitful
+
+julia> Unitful.numtype(1.0u"m")
+Float64
+
+julia> Unitful.numtype(typeof(1.0u"m"))
+Float64
+
+julia> Unitful.numtype(1 + 2im)
+Complex{Int64}
+```
+"""
+function numtype end
+
+@inline numtype(x::Number) = typeof(x)
+@inline numtype(::Type{T}) where {T<:Number} = T
 @inline numtype(::AbstractQuantity{T}) where {T} = T
 @inline numtype(::Type{Q}) where {T, Q<:AbstractQuantity{T}} = T
+@inline numtype(x::Base.Enum) = numtype(typeof(x))
+@inline numtype(::Type{E}) where {T, E<:Base.Enum{T}} = T
 
 @inline dimtype(u::Unit{U,D}) where {U,D} = D
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # Unitful.jl changelog
 
+## Unreleased
+
+* ![Feature:](https://img.shields.io/badge/-feature-green) `Unitful.numtype` is now part of the public API. It returns the underlying numeric type of a quantity (e.g. `Float64` for `1.0u"m"`), of any `Number` (where it just returns its type), or of a `Base.Enum` (the backing integer type). On Julia ≥ 1.11 the name is declared `public`.
+
 ## v1.28.0 (2026-01-29)
 
 * ![Feature:](https://img.shields.io/badge/-feature-green) Dimensionless quantities now support `iseven` and `isodd` ([#829](https://github.com/JuliaPhysics/Unitful.jl/pull/829)).
--- a/docs/src/manipulations.md
+++ b/docs/src/manipulations.md
@@ -34,6 +34,12 @@ Unitful.dimension
 Unitful.ustrip
 ```
 
+## Numeric type extraction
+
+```@docs
+Unitful.numtype
+```
+
 ## Unit multiplication
 
 ```@docs
--- a/src/Unitful.jl
+++ b/src/Unitful.jl
@@ -36,6 +36,11 @@ export @logscale, @logunit, @dB, @B, @cNp, @Np
 export Level, Gain
 export uparse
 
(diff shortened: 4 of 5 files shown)


PatrickHaecker:
@sostock: Anything I can do to get this decided?

cstjean (approved):
LGTM (for whatever that's worth, as an outsider)

PatrickHaecker:
CI errors seem to be unrelated.

src/utils.jl
```diff
@@ -1,8 +1,41 @@
...
+```
+"""
+function numtype end
+
+@inline numtype(x::Number) = typeof(x)
+@inline numtype(::Type{T}) where {T<:Number} = T
 @inline numtype(::AbstractQuantity{T}) where {T} = T
 @inline numtype(::Type{Q}) where {T, Q<:AbstractQuantity{T}} = T
+@inline numtype(x::Base.Enum) = numtype(typeof(x))
+@inline numtype(::Type{E}) where {T, E<:Base.Enum{T}} = T
```
sostock, line 38: What is the usecase for adding these? Are there users who use enums as numerical values with Unitful? If that’s the case, I think it should return the Enum type, not the number type backing the enum.

PatrickHaecker: That is a good point. But it turns out that `Base.Enum{T}` requires `T<:Integer`, so an enum can't have a `Quantity` as its base type, and an enum isn't `<:Number`, so an enum can't be the value inside a `Quantity` either.

The use case here is simpler: We read a configuration file into a struct, and have some values which need to parsed as a unit and some are an `Enum` value. When `numtype` is defined both for units and for enums, it just works with multiple dispatch without special casing.

You could say that it's not Unitful's business to define such methods. However, as the functions fully qualified name is `Unitful.numtype` and `Base` can't depend on `Unitful`, only `Unitful` can provide the method without committing type piracy.
Given the constraint that enums can't be used as numerical values within `Unitful`, the proposed definition in the PR seems to be the only semantically correct way: The integer type is the "numeric type" of an enum. It is also the "base type" (`Base.Enums.basetype`), but that does not mean that it would not be the `numtype`, too.

The alternative to implementing the methods in `Unitful` (if type piracy is not an option), would be that packages needed to define the method for each of their custom enum types. Not the end of the world, but less convenient and the question would probably come up why there is no generic definition.

sostock: I don’t agree that this is semantically correct. For me, the numeric type of a quantity and the backing type of an enum are semantically very different. Therefore, I don't think that `numtype` should handle enums.

An enum is not composed of a value and a unit. In many (not all!) cases, the fact that enum values are backed by numbers *at all* is an implementation detail, whereas for a quantity the numeric value (the value itself, not necessarily its type) is an integral part. The purpose of `numtype` is to extract the type of the numerical value of a quantity, and not “the number type that any value uses internally”.

> You could say that it's not Unitful's business to define such methods. However, as the functions fully qualified name is Unitful.numtype and Base can't depend on Unitful, only Unitful can provide the method without committing type piracy.

I don’t see how type piracy has anything to do with this. Any package could provide such a combined function without committing type piracy. The following is not type piracy if the package that defines these methods owns the `my_numtype` function:
```julia
my_numtype(x) = Unitful.numtype(x)
my_numtype(x::Base.Enum) = my_numtype(typeof(x))
my_numtype(::Type{E}) where {T, E<:Base.Enum{T}} = T
```

> The alternative to implementing the methods in Unitful (if type piracy is not an option), would be that packages needed to define the method for each of their custom enum types.

No, they don’t need to define it for each of their custom enums. They can generically do it for all enum types, just like the `my_numtype` function above does.

sostock:
CI failures are due to https://github.com/coverallsapp/github-action/issues/264.

src/utils.jl
PatrickHaecker: Thanks for the explanations. 

> For me, the numeric type of a quantity and the backing type of an enum are semantically very different.

We agree here.

> Therefore, I don't think that `numtype` should handle enums.

This is where I come to a different conclusion. I respect that you think that `numtype` should not handle enums. I just think the "therefore" implying a logical conclusion is unfortunate. Quantities and enums are semantically very different, yet they both have a method for `typeof` (implicitly) or `hash` (explicitly), so semantic difference does not seem to be a good criterion on the question whether two types should both have methods for the same function. I think the deciding factor should be "does the method makes sense", which means "does the function make sense for the provided arguments".

> An enum is not composed of a value and a unit. In many (not all!) cases, the fact that enum values are backed by numbers _at all_ is an implementation detail

The numbers of an enum can be set by the user, so code handling an enum needs to be capable of handling such values as user input. I agree that there are enums where the numbers are an implementation detail, but as long as code handling enums does not differentiate between both, you need to consider that it _sometimes_ is user input.

> The purpose of `numtype` is to extract the type of the numerical value of a quantity, and not “the number type that any value uses internally”.

I think this is a perfect description of the _method_ `numtype(::AbstractQuantity)`. The question we should focus here, if you agree that `numtype` should be public, is what the description of the _function_ `numtype` should be. I made a proposal (in the docstring in the PR) of what I think is useful. But you have much more experience with the function. What would be your take?

> > You could say that it's not Unitful's business to define such methods. However, as the functions fully qualified name is Unitful.numtype and Base can't depend on Unitful, only Unitful can provide the method without committing type piracy.
> 
> I don’t see how type piracy has anything to do with this. Any package could provide such a combined function [`my_numtype`] without committing type piracy.
> > The alternative to implementing the methods in Unitful (if type piracy is not an option), would be that packages needed to define the method for each of their custom enum types.
> 
> No, they don’t need to define it for each of their custom enums. They can generically do it for all enum types, just like the `my_numtype` function above does.

I agree, but I was referring to `numtype` specifically. Every package can implement any logic, but would need to reimplement/dispatch on all existing `numtype` functionality to extend it. As this is more effort for the user and more error prone, the default mechanism to extend a function is by adding additional methods to existing functions, _if_ that makes sense semantically. And this is probably where we disagree. I think it would help if you could explain which problems "get the numeric type of the enum" has. Because this is how I would read `numtype(::Enum)`

Strictly speaking, I combined two unrelated things into this PR. Whether `numtype` is public and whether there is a method for enums can be decided separately. Would it help if I'd remove the enum method definitions from the PR? Or would this PR still be too controversial?

Thanks again for your time on this PR and on working on Unitful.jl. I really appreciate it!

PatrickHaecker: @sostock: Should I remove the enums from this PR to get it over the finish line? If so, should I add it as a separate PR?

sostock: Yes, please remove the enum methods.

We can still discuss here, you don’t need to make a separate PR for the enum methods (since I don’t think I will merge it).

I generally think that having a function whose methods do semantically very different things is not useful. To make it possible to write generic code, functions should have a clearly defined meaning.

